PR #19, opened 2021-01-19
setup: bump LIEF to 0.11

williballenthin:


danielplohmann (approved):

danielplohmann:
thx a lot for the pointer! I'll also have another look and check if there are other prints in SMDA that may mess up the output.

danielplohmann:
I've just also used lief.logging.disable to suppress this output, this would potentially allow to move to more recent LIEF versions.

PR #21, opened 2021-05-19
Fix invalid usage of traceback.format_exc

BonusPlay:
This function doesn't even take exception as an argument.
https://docs.python.org/3/library/traceback.html#traceback.format_exc


danielplohmann:
good point, thx!

Issue #23, opened 2021-08-18
SMDA incorrectly maps sections/segments from ELF files


williballenthin:
The variable `min_raw_segment_offset` used https://github.com/danielplohmann/smda/blob/bb9cdab2279f255f5516750160c3a1ff424f8882/smda/utility/ElfFileLoader.py#L61 and https://github.com/danielplohmann/smda/blob/bb9cdab2279f255f5516750160c3a1ff424f8882/smda/utility/ElfFileLoader.py#L77 in the ElfFileLoader are used incorrectly, leading to sections/segments being truncated.

This variable contains a virtual address; however, it is used as a raw file offset (the raw file offset of the first section/segment). When the virtual address is something like 0x401000, then the assignment tries to copy 0x401000 from the source binary, ends up with much less data (the size of the file amount of data), and this truncates the mapped data to `len(size of file)`. Subsequent assignments to the mapped data somehow don't throw exceptions despite failing to write beyond the end of the truncated mapping. 

A reasonable fix is to operate on physical/raw file offsets rather than virtual addresses.

williballenthin:
I'll wrap this fix into my PR for #22.

williballenthin:
![image](https://user-images.githubusercontent.com/156560/129838512-cc3af6d3-2d47-45ea-b0b3-df8f20316b27.png)
pending code, needs a bit more documentation, likely to be submitted tomorrow.

williballenthin:
incidentally, should we be using both sections *and* segments, rather than just one *or* the other? maybe its good enough as is. certainly i'll open an issue if i find this breaking in the real world.

danielplohmann:
Oh, that's a good catch!
I must have overlooked that when I previously added/fixed the simulated loading of the ELF files.
At least I didn't notice such significant cases of code truncation when running SMDA against my groundtruth in the evaluation. The results were within the expected margin that I didn't even consider that the loading was bugged that much.
I don't have an understanding of ELF files deep enough to propose a solid solution as my previous attempt at loading showed. :)
The first segments - then sections sounds really good though and appears to fix the problems you encountered!

Issue #22, opened 2021-08-17
SMDA does not recognize imported APIs from ELF files


williballenthin:
As triaged in https://github.com/fireeye/capa/issues/725, SMDA doesn't parse imports from ELF files into a representation used for APIs.

I'd propose implementing an `ElfApiResolver` that uses LIEF to parse the ELF imports and responds to `getApi`.

williballenthin:
```py
for relocation in lief.ELF.Binary.relocations:
    if not relocation.has_symbol:
        continue
    if not relocation.symbol.imported:
        continue
    if not relocation.symbol.is_function:
        continue

    # we can't really say what library the symbol came from
    # however, we can treat the version (if present) as relevant metadata?
    lib = None
    if relocation.symbol.has_version and relocation.symbol.symbol_version.has_auxiliary_version:
        # like "GLIBC_2.2.5"
        lib = relocation.symbol.symbol_version.symbol_version_auxiliary.name

    name = relocation.symbol.demangled_name

    print(lib, name)
```

williballenthin:
i'm working on a PR for this. hope to have it to you tomorrow.

williballenthin:
how important is the check `not self.disassembly.isAddrWithinMemoryImage(dereferenced)` here:
https://github.com/danielplohmann/smda/blob/bb9cdab2279f255f5516750160c3a1ff424f8882/smda/intel/IntelDisassembler.py#L179

In the ELF binary I'm using for testing (7351f8a40c5450557b24622417fc478d), the "API thunks" look like `jmp qword ptr [...]` where the to_addr is in the `.got.plt` section and the dereferenced value is a valid address in the `.extern` section. Therefore, `isAddrWithinMemoryImage(dereferenced) == True` and the condition fails, leading to APIs not getting recognized.

I wonder if we can remove this restriction? Alternatively, add another condition to check for "is within the import section?" or similar.

I've been trying to imagine why it was included in the first place. My best guess is that SMDA supports analysis of memory images, and this check is useful with PE files whose import table contains fixed-up pointers to other DLLs in the memory image. Is there a different way we could structure this check to support ELF files, too?

williballenthin:
how important is the check `not self.disassembly.isAddrWithinMemoryImage(dereferenced)` here:
https://github.com/danielplohmann/smda/blob/bb9cdab2279f255f5516750160c3a1ff424f8882/smda/intel/IntelDisassembler.py#L179

[...]

PR #24, opened 2021-08-18
ELF loading and API fixes

williballenthin:
closes:
  - #22 
  - #23 

Adds an `ElfApiResolver` that parses the ELF imports. Fixes ELF loading and updates it for more consistent handling of segments and sections. However, I'm not particularly experienced with ELF files, so please also feel free to cherry-pick some/all/none of those changes.

Also:
  - better check None vs 0
  - support extracting API names without associated DLL (such as in ELF files)
  - don't require APIs to dereference outside of the the image

### DIFF
--- a/smda/common/labelprovider/ElfApiResolver.py
+++ b/smda/common/labelprovider/ElfApiResolver.py
@@ -1,0 +1,63 @@
+#!/usr/bin/python
+import lief
+lief.logging.disable()
+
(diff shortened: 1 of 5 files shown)


danielplohmann:
Thank you very much for this extensive pull request!
That is a great addition of functionality to SMDA.
I'll provide some insight into the previous choices in the issues!

PR #25, opened 2021-08-19
elf: fix alignment calculcation

williballenthin:
silly arithmetic bug. not sure how it ever worked. identified over [here](https://github.com/vivisect/vivisect/pull/436/commits/14f9c972b3bda6b8863600d35b797d04ca38c2dc#r692441396) when it broke some other tests.

### DIFF
--- a/smda/utility/ElfFileLoader.py
+++ b/smda/utility/ElfFileLoader.py
@@ -16,7 +16,7 @@ def align(v, alignment):
     if remainder == 0:
         return v
     else:
-        return v + remainder
+        return v + (alignment - remainder)
 
 
 class ElfFileLoader(object):


danielplohmann:
thx for the fix! 

I just had a look - the way it is used in SMDA will only affect the very end of the mapping anyway.
wrt  bytearray behavior - seems like if you assign to a range that spans outside the previously declared size of the bytearray, it will be expanded as needed - which will either simply have end up almost having the intended effect (in case target slice overlaps with the size) or it will be appended (in case the target slice of the assignment is beyond the array - this leads to broken layout).

```
>> a = bytearray(b"a"*10)
>>> a
bytearray(b'aaaaaaaaaa')
>>> a[5:15] = b"b" * 10
>>> a
bytearray(b'aaaaabbbbbbbbbb')
>>> a[25:30] = b"c" * 5
>>> a
bytearray(b'aaaaabbbbbbbbbbccccc')
>>> len(a)
20
```

PR #38, opened 2023-01-31
lief.parse do not accept "bytearray", so: change bytearray(xxxx) to x…

lainswork:
I'm new to python. I don't know what it means，so....be careful


danielplohmann:
thank you very much for the pull request.
I've verified that everything works as intended and your changes were thorough.
Looks like the LIEF API was recently changed, as this very same parsing from the bytearray has worked for years before. ;)

PR #39, opened 2023-02-03
Allows for finding a function or a block by searching for any addresses that it contains

cccs-ay:
Previous behaviour only allowed to find a function by the start address. Now, any address within any of the function's blocks can be used to find a function. 

To allow this, the BlockLocator is instantiated the first time SmdaReport.findBlockByContainedAddress is called.

BlockLocator creates the required datastructure to perform a binary search on the list of blocks start addresses to find the closest match. 

BlockLocator could be instantiated explicitely (in a similar fashion the CodeXrefs) if that would be a cleaner design.

### DIFF
--- a/smda/common/BlockLocator.py
+++ b/smda/common/BlockLocator.py
@@ -20,10 +20,10 @@ class BlockLocator():
         # 2 a dict of blocks by addresses
         self.blocks_dict = {b.offset:b for b in blocks}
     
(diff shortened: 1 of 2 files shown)


danielplohmann:
thanks a lot!!
I'll proceed and publish a version 1.11.0 in the next minutes, reflecting your changes into PyPI availability.

PR #55, opened 2024-04-17
Add GO magic bytes for version >=1.20

Manny684:
Go version >=1.20 uses different magic bytes. As far as I can see, pclntab did not change in a relevant way otherwise.


danielplohmann:
that's good news, thanks a lot for the fix!

PR #63, opened 2025-02-21
Config Hash Option 

alexander-hanel:
Hello, 

While profiling SMDA in Pycharm, I noticed that the hashing can take up some cycles so I added an option in the config to disable it. I didn't see any code dependencies on the hashing so feel free to reject this pull request if I missed something. 

Cheers, 

Alex


danielplohmann:
I fully agree, thanks for the extension!

Issue #65, opened 2025-02-22
Lief symbols never gets added to func_symbols due to python's typing


N0fix:
While working on MinGW compiled executables, I noticed that no symbols were found, although IDA could load every DWARF symbols properly.

Looking at commits, I noticed that the code responsible is the following function:

https://github.com/danielplohmann/smda/blob/c11aabe77ed8d98fe5df1271d3682078b6d46078/smda/common/labelprovider/PeSymbolProvider.py#L60-L82

Line 78 is wrong:

https://github.com/danielplohmann/smda/blob/c11aabe77ed8d98fe5df1271d3682078b6d46078/smda/common/labelprovider/PeSymbolProvider.py#L78

It is comparing every character in `function_name` with integer values, which can never match.

A simple fix would be:

```diff
- if function_name and all(c in range(0x20, 0x7f) for c in function_name):
+ if function_name and all(ord(c) in range(0x20, 0x7f) for c in function_name):
```

This can easily be tested in a python interpreter:
```python
In [7]: all(c in range(0x20, 0x7f) for c in 'calloc')
Out[7]: False

In [8]: all(ord(c) in range(0x20, 0x7f) for c in 'calloc')
Out[8]: True
```

danielplohmann:
Thanks!
What's weird: I think this worked before, so not sure if this was also broken by upgrading LIEF lately. 
Anyway, I'm going to cover these aspects with tests now, this should ensure that things like this no longer go unnoticed.

N0fix:
I suspected that the code was tested and working at some point, but I couldn't find any case where this line of code works.

Thanks for adding the test cases !

PR #66, opened 2025-02-22
Fix invalid function name validation on PeSymbolProvider

N0fix:
This PR should fix https://github.com/danielplohmann/smda/issues/65


danielplohmann:
Thanks, I will add a PE with DWARF symbols to also cover this with a test, will do the same for ELF.

PR #68, opened 2025-03-18
Fix ValueError in `parseOepFromArgs` due to empty string argument

r0ny123:
This prevents a `ValueError` when `args.oep` is an empty string.


danielplohmann:
Fair point, I should simply have initialized all those with None to avoid such things. :)

PR #69, opened 2025-06-10
Adding fixes for IDA 9.0 IDAPython changes.

jershmagersh:
There were a few minor fixes needed to be compatible with IDA 9.0. I created a 9.0 interface class, but if you'd like these changes bundled with the 7.4 interface please let me know.

### DIFF
--- a/smda/ida/IdaInterface.py
+++ b/smda/ida/IdaInterface.py
@@ -32,8 +32,10 @@ class IdaInterface(object):
     instance = None
     def __init__(self):
         if not IdaInterface.instance:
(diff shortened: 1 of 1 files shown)


danielplohmann (commented):
I think the version check should be >= 740 and < 900, to exclude usage of the 74 interface for 900 and above?
I'll fix that in a commit before integration, if I'm wrong, please let me know!

jershmagersh:
Yes that's correct, thanks for fixing that and adding credits, much appreciated!

PR #92, opened 2026-05-21
fix(intel): wildcard movabs absolute offsets

r0ny123:
## Summary

This PR fixes Intel instruction wildcarding for `movabs` memory operands that Capstone reports with 64-bit absolute offsets.

Previously, `escapeBinaryPtrRef()` parsed bracketed pointer operands and tried to encode the parsed offset as a 32-bit value first. That worked for ordinary 32-bit displacements and RIP-relative references, but it did not distinguish `movabs` absolute memory offsets whose displacement is encoded as 8 bytes. As a result, those instructions could be wildcarded with the wrong operand width.

## What changed

- Uses Capstone detail metadata (`disp_size`) to detect 8-byte absolute memory offsets.
- Packs 8-byte displacements as little-endian 64-bit values and generates a matching 16-character wildcard mask.
- Keeps the existing 32-bit behavior for ordinary pointer references and RIP-relative operands.
- Preserves the existing duplicate-match behavior that escapes the rightmost occurrence when the packed offset appears twice.
- Adds regression coverage for both `movabs` absolute-memory forms:
  - `movabs rax, qword ptr [abs64]`
  - `movabs qword ptr [abs64], rax`

## Impact

This only changes Intel binary wildcarding output for pointer references whose encoded displacement is actually 8 bytes. It does not change public APIs, report serialization, or disassembly control flow.

## Validation

- `.venv/bin/python -m pytest tests/testEscaper.py -q`
- `.venv/bin/python -m ruff check .`
- `.venv/bin/python -m ruff format --check .`
- `PATH="$PWD/.venv/bin:$PATH" make lint`

### DIFF
--- a/smda/intel/IntelInstructionEscaper.py
+++ b/smda/intel/IntelInstructionEscaper.py
@@ -2278,16 +2278,22 @@ class IntelInstructionEscaper:
             offset = int(addr_match.group("dword_offset"), 16)
             if "rip -" in ins.operands:
                 offset = 0x100000000 - offset
-            # TODO we need to check if this is actually a 64bit absolute offset (e.g. used by movabs)
+            pack_format = "<I"
             try:
-                packed_hex = str(codecs.encode(struct.pack("I", offset), "hex").decode("ascii"))
+                if ins.getDetailed().disp_size == 8:
+                    pack_format = "<Q"
+            except (AttributeError, AssertionError, NotImplementedError):
+                pass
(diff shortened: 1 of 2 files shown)


r0ny123:
Closing this PR because its change has been consolidated into #97, the new refactor/modernize-repo-structure PR. The new PR keeps the relevant commit reference and ports the change onto the modernized src/smda layout.

PR #93, opened 2026-05-21
refactor(intel): clarify indirect import slot resolution

r0ny123:
## Summary

This PR clarifies and regression-tests how `IndirectCallAnalyzer` handles `mov <reg>, dword ptr [<addr>]` when `<addr>` is an import-table slot.

The existing code already had the right behavior, but it was hidden behind a `HACK` comment. If the pointer address itself resolves to an imported API, the analyzer must preserve that pointer slot address as the candidate. That keeps API resolution keyed consistently with the PE/ELF import resolvers, which can identify imports by the import slot rather than by the memory value loaded from that slot.

## What changed

- Extracted the import-slot check and dword fallback into `_resolveDwordPointerValue()`.
- Replaced the `HACK` comment with an explanation of why known import slots are preserved.
- Kept the existing final `resolveApi(candidate, candidate)` behavior unchanged.
- Added unit coverage for both paths:
  - known import slot: preserve the slot address and record API metadata under that address
  - non-import pointer: fall back to reading the dword value and queue it as a function candidate when appropriate

## Why

Indirect register-call recovery backtracks register values. For instructions like:

```asm
mov eax, dword ptr [0x403000]
call eax
```

`0x403000` may be an import slot. Dereferencing it too early can lose the address shape that SMDA's import providers use to resolve and record API calls. This PR makes that domain rule explicit and protects it with tests.

## Validation

Run from a clean worktree based on `upstream/master`:

- `python -m pytest tests/testIndirectCallAnalyzer.py -q`
- `python -m ruff check smda/intel/IndirectCallAnalyzer.py tests/testIndirectCallAnalyzer.py`
- `make lint`
- `make test`

Result: all checks passed. Full test suite passed with 82 tests.

### DIFF
--- a/smda/intel/IndirectCallAnalyzer.py
+++ b/smda/intel/IndirectCallAnalyzer.py
@@ -31,6 +31,12 @@ class IndirectCallAnalyzer:
             return None
         return struct.unpack("I", self.disassembly.getBytes(addr, 4))[0]
 
+    def _resolveDwordPointerValue(self, addr):
+        dll, api = self.disassembler.resolveApi(addr, addr)
+        if dll or api:
+            return addr, dll, api
+        return self.getDword(addr), dll, api
+
     def processBlock(self, analysis_state, block, registers, register_name, processed, depth):
         if not block:
             return False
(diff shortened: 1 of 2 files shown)


r0ny123:
Closing this PR because its change has been consolidated into #97, the new refactor/modernize-repo-structure PR. The new PR keeps the relevant commit reference and ports the change onto the modernized src/smda layout.

PR #100, opened 2026-05-25
fix(security): avoid swallowing non-operational exceptions

r0ny123:
## Summary

- add a shared helper for broad exception handlers to re-raise non-operational exceptions
- keep public disassembly entry points returning error reports for operational failures
- add an AST regression guard that checks broad `except Exception` handlers under `smda/`

## Review

- repo-local `smda-pattern-scanner` found no confirmed issue in the branch diff
- pattern: broad exception handlers must not swallow developer/runtime-control failures
- instances covered: 19 broad `except Exception` handlers under `smda/`

## Validation

- `python -m pytest tests/testExceptionHandling.py -q` -> 5 passed

### DIFF
--- a/smda/Disassembler.py
+++ b/smda/Disassembler.py
@@ -5,6 +5,7 @@ import traceback
 
 from smda.cil.CilDisassembler import CilDisassembler
 from smda.common.BinaryInfo import BinaryInfo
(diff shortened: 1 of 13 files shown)


danielplohmann:
decided to just rebuilt this commit with f2753e5c0d8e191cb25d24a533b3290539f8e313, now pushed. :)

PR #101, opened 2026-05-25
refactor(dalvik): dispatch opcode decoding by format

r0ny123:
## Summary

- replace the large Dalvik opcode format conditional with a format-dispatch table
- keep decoder return fields assembled in one final `DecodedDalvikInstruction` construction
- preserve existing operand, register, literal, reference, branch, and payload decoding behavior

## Review

- repo-local `smda-pattern-scanner` found no confirmed decoder correctness regression
- residual risk noted by the scanner: most moved Dalvik format handlers rely on existing fixture-level coverage rather than one direct synthetic test per format

## Validation

- `python -m pytest tests/testDalvikDisassembler.py -q` -> 30 passed

### DIFF
--- a/smda/dalvik/DalvikOpcodeDecoder.py
+++ b/smda/dalvik/DalvikOpcodeDecoder.py
@@ -444,46 +444,18 @@ FORMAT_DECODERS = {}
 
 
 def _decode_fmt_10x(raw_bytes, byte_idx, opcode, resolve_ref):
-    registers = []
-    operands = ""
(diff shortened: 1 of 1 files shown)


danielplohmann:
merged manually, github doesnt seem to catch up O_o

r0ny123:
Yeah, thanks. @danielplohmann, could you please rebase the merge commits if possible when merging them so that the history looks clearer?

PR #102, opened 2026-05-25
refactor(loaders): split ELF binary mapping helpers

r0ny123:
## Summary

- split ELF `mapBinary` boundary calculation, segment mapping, and section mapping into dedicated helpers
- preserve the existing segment-before-section mapping order
- keep header mapping, size checks, and empty-ELF behavior in `mapBinary`

## Review

- repo-local `smda-pattern-scanner` found no confirmed ELF mapper behavior regression
- residual risk noted by the scanner: current tests cover ELF parsing at fixture level, not every synthetic padding, bogus-section, or overlap edge case

## Validation

- `python -m pytest tests/testFileFormatParsers.py -q` -> 4 passed

### DIFF
--- a/smda/utility/ElfFileLoader.py
+++ b/smda/utility/ElfFileLoader.py
@@ -61,14 +61,14 @@ class ElfFileLoader:
         return base_addr
 
     @staticmethod
-    def _calculate_boundaries(elffile):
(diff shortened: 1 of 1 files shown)


danielplohmann:
integrated and rebased manually

PR #108, opened 2026-05-26
perf(loaders): replace O(n²) mergeCodeAreas with linear merge in ELF and Macho

r0ny123:
## Branch and base
Branch: claude/perf-sweep-20260527-merge-code-areas
Commit: 8433f00
Base: master (c1227d7)

## Optimization class
OC-2 — mergeCodeAreas-O(n²)

## Root cause
`ElfFileLoader.mergeCodeAreas` and `MachoFileLoader.mergeCodeAreas` used an identical
while-loop that merged adjacent intervals by list slicing on every merge step:

```python
merged_code_areas = merged_code_areas[:index] + [[a, b]] + merged_code_areas[index + 2:]
```

Each splice copies the entire list, making k merges cost O(n·k) ≤ O(n²). The loop also
maintained a `result` list that was appended to but never returned — dead code.

`PeFileLoader.mergeCodeAreas` already used the correct O(n) pattern: seed `result` with the
first area, walk forward linearly, extend `result[-1]` in-place on adjacency, otherwise
append.

## Change
Both ELF and Macho implementations are replaced with the PE pattern. No logic changes —
same merge condition (`area_a[1] == area_b[0]`), same output for all inputs. The dead
`result` variable is removed.

## Before / after evidence
Targeted test suite (testFileFormatParsers + testIntegration, 22 tests):

```
baseline:  22 passed in 2.15s
after:     22 passed in 1.99s
```

The O(n²) pathology only manifests when k adjacent-section merges occur (k ≥ 2).
Typical binaries have 1–10 executable sections, so the practical improvement is in
allocation pressure (no temporary list copies) rather than raw iteration count.

## Behavior compatibility
Reviewer subagent (fresh context) returned **PASS**. Confirmed:
- Semantic equivalence for: empty, single, all-adjacent, no-adjacent, unsorted, overlapping inputs. ✓
- Caller contract in `getCodeAreas` unchanged. ✓
- All three loaders now use byte-for-byte identical implementations — divergence reduced. ✓
- No tests removed. ✓

## Validation commands
```
```

## Residual risk
None. Pure algorithmic replacement of a static utility method with a semantically identical
but asymptotically better implementation. Not serialized; not part of the public report schema.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 PYTHON ?= python3
 
+.PHONY: init package publish ruff-check ruff-format ruff-fix lint format test test-coverage clean
+
 init:
 	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0"
 	$(PYTHON) -m pip install -e ".[dev]"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PYTHON ?= python3
 
 init:
-	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<75.4.0" wheel
+	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0,<0.48.0"
 	$(PYTHON) -m pip install -e ".[dev]"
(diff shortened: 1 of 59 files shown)


r0ny123:

PR #111, opened 2026-05-26
perf(labels): cache parsed API database at class level in WinApiResolver

r0ny123:
## Branch and base
Branch: claude/perf-sweep-20260527-winapi-cache
Commit: a16d7a2
Base: master (c1227d7)

## Optimization class
OC-5 — WinApiResolver-per-instance

## Root cause
`_loadDbFile()` reads and parses the ApiScout JSON collection on every `WinApiResolver.__init__()` call. Since each `IntelDisassembler` creation instantiates a new `WinApiResolver`, batch pipelines that create one `Disassembler` per binary re-parse the same static file for every binary.

`API_COLLECTION_FILES` defaults to `{}` so this is a no-op for most users; the optimization only fires for users who have a custom ApiScout database configured.

## Change
Add a class-level `_db_cache: dict` mapping `db_filepath -> (api_map, has_64bit)`.

- **Cache hit**: assign the shared `api_map` reference to `self._api_map[os_name]` and OR `has_64bit` into `self._has_64bit`. No disk I/O, no JSON parse, no DLL iteration.
- **Cache miss**: parse as before, store `(api_map, has_64bit)` in cache, then update instance state.
- **File-not-found**: early return without writing to cache — a reappearing file is picked up on next call.

## Before / after evidence
The test environment has `API_COLLECTION_FILES = {}`, so `_loadDbFile` is never called and timing is unchanged (10 passed in ~1.0s both ways). Direct measurement requires a configured ApiScout database, which is not part of the test fixtures. Savings are proportional to database size × number of `Disassembler` instances created per process.

## Behavior compatibility
- `self._has_64bit` final value identical: local `has_64bit` accumulates across the DLL loop then `|=` into the instance. ✓
- Shared `api_map` reference is safe: `update()` only mutates `self._api_map["lief"]` (always a fresh `{}` per instance); `getApi()` is read-only `.get()`. ✓
- File-not-found not cached: correct — missing file may reappear. ✓
- Multiple `os_name` → same path: each caller gets the correct map under their `os_name` key. ✓
- Test isolation: `_db_cache` stays empty in tests (`API_COLLECTION_FILES = {}`). ✓

## Validation commands
```
```

## Residual risk
Low. Class-level state persists for the process lifetime. The only scenario where this differs from the old behavior is if a configured API file changes on disk between two `WinApiResolver` instantiations in the same process — not a supported use case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 PYTHON ?= python3
 
+.PHONY: init package publish ruff-check ruff-format ruff-fix lint format test test-coverage clean
+
 init:
 	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0"
 	$(PYTHON) -m pip install -e ".[dev]"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PYTHON ?= python3
 
 init:
-	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<75.4.0" wheel
+	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0,<0.48.0"
 	$(PYTHON) -m pip install -e ".[dev]"
 	$(PYTHON) -m pre_commit install
 package:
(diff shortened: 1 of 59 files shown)


r0ny123:

PR #112, opened 2026-05-26
perf(common): build reverse blockrefs index for O(1) predecessor lookup

r0ny123:
## Branch and base
Branch: claude/perf-sweep-20260527-predecessors-index
Commit: 4fa0331
Base: master (c1227d7)

## Optimization class
OC-6 — getPredecessors-O(n²)

## Root cause
`SmdaBasicBlock.getPredecessors()` found incoming edges by scanning the full forward adjacency dict (`blockrefs: {frm -> [to, ...]}`) on every call:

```python
for frm, to in self.smda_function.blockrefs.items():
    if self.offset in to:
        predecessors.append(frm)
```

Calling this for all N blocks in a function costs **O(N × E)** total (E = edge count), degrading to O(N²) for dense CFGs. By contrast, `getSuccessors()` already does an O(1) direct key lookup.

## Change
Add `_blockrefs_reverse = None` sentinel on `SmdaFunction`. On the first `getPredecessors()` call for any block in a function, build `{to: [frm, ...]}` in O(E) and cache it on the function. All subsequent calls are O(1) dict lookups. Return contract preserved: a fresh list copy is returned each call.

`blockrefs` is finalized during construction (`__init__` or `fromDict`) before any `SmdaBasicBlock` is accessible, so the lazy index is always built on the final edge set.

## Before / after evidence
Targeted tests (testIntegration + testIntelDisassembler, 16 tests):

```
baseline:  16 passed in 1.05s
after:     16 passed in 1.01s
```

The full benefit is only visible when `getPredecessors()` is called for all or many blocks in a function (e.g. dominator-tree computation, graph visualization). Each function's index is built once; all blocks in that function reuse it.

## Behavior compatibility
- `rev.get(self.offset, [])` returns exactly the same `frm` addresses as the old linear scan. ✓
- `list(...)` return preserves the fresh-list-per-call contract. ✓
- `smda_function is None` early-return preserved. ✓
- Index cannot be built before `blockrefs` is finalized (no `SmdaBasicBlock` exists during construction). ✓
- Benign race under concurrent construction (identical dicts, no corruption) — SMDA is single-threaded. ✓

## Validation commands
```
```

## Residual risk
None under single-threaded use. Benign under hypothetical concurrent construction (same dict built twice, last write wins).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 PYTHON ?= python3
 
+.PHONY: init package publish ruff-check ruff-format ruff-fix lint format test test-coverage clean
+
 init:
 	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0"
 	$(PYTHON) -m pip install -e ".[dev]"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PYTHON ?= python3
 
 init:
-	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<75.4.0" wheel
+	$(PYTHON) -m pip install --upgrade pip "setuptools>=64.0.0,<82.1.0" "wheel>=0.47.0,<0.48.0"
 	$(PYTHON) -m pip install -e ".[dev]"
 	$(PYTHON) -m pre_commit install
 package:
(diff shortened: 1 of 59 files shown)


r0ny123:

PR #131, opened 2026-06-10
feat(ida): add optional IDA Domain backend

r0ny123:
## Summary

Adds IDA Domain as an optional, preferred backend for compatible IDA 9.1+ installations while retaining the existing IDAPython adapters for legacy IDA users.

This addresses the maintainer feedback by preserving IDA 8.4/8.5 and older workflows, keeping `ida-domain` out of the core installation, and selecting Domain only when both the IDA SDK version and package availability permit it.

## Compatibility policy

| IDA SDK | Preferred backend | Fallback |
| --- | --- | --- |
| below 7.4 | existing IDAPython 7.3 adapter | n/a |
| 7.4-8.4 | existing IDAPython 7.4 adapter | n/a |
| 8.5-9.0 | existing IDAPython 8.5 adapter | n/a |
| 9.1+ | IDA Domain when installed | IDAPython 8.5 adapter |

The policy lives in one ordered backend table and one Domain threshold in `IdaInterface.py`. Callers no longer make their own version or dependency decisions. Adding or dropping a generation is therefore an explicit selector-table change plus its isolated adapter, rather than a repository-wide conditional cleanup.

## Changes

- Split the legacy implementations into `IdaIdapythonInterface.py` while preserving the public `Ida73Interface`, `Ida74Interface`, and `Ida85Interface` imports.
- Added an isolated `IdaDomainInterface` adapter and a central facade that prefers it only for SDK 9.1+ when `ida-domain` imports successfully.
- Made Domain availability probing side-effect free: importing the adapter no longer initializes IDA's licensed SWIG runtime, and GUI selection still falls back to IDAPython if lazy Domain initialization fails.
- Used IDA Domain 0.5 high-level functions, flowcharts, instructions, bytes, xrefs, names, imports, and segments APIs where they directly replace low-level traversal. The low-level `ida_loader` path call remains only for the IDB directory because `Database.path` identifies the input file, not the database file.
- Added the `ida` optional extra (`pip install "smda[ida]"`); ordinary SMDA installations remain unchanged.
- Added `ida_domain_export.py` for headless database/input export through IDA Domain/idalib.
- Restored `ida_analyze.py` as an independent SMDA recovery workflow that adds functions and names to the current IDA database. It no longer re-exports IDA's existing function set.
- Kept IDA 7.x support because the isolated adapters do not block the Domain design and removing them provides no practical benefit in this change.
- Added compatibility-boundary, fallback, lifecycle, architecture, import, flowchart, byte, xref, and script-role regression tests.
- Integrated the upstream exporter and AArch64 behavior that the Domain adapter builds on.

The implementation was checked against the official [IDA Domain documentation](https://ida-domain.docs.hex-rays.com/), its [LLM reference](https://ida-domain.docs.hex-rays.com/llms-full.txt), the [Domain API overview](https://hex-rays.com/blog/whats-new-in-the-ida-domain-api), and the [IDA 9.4 release notes](https://hex-rays.com/blog/ida-9.4-release-a-new-dyld-shared-cache-swift-analysis-new-teams-add-on-and-more-). Version 0.5 is used because its current high-level imports API and compatibility work provide concrete value without pulling decompiler/microcode APIs into this exporter.

## Validation

- `make lint` - passed
- `make test` - 471 passed, 1 skipped
- `python -m ruff format --check .` - passed
- `pre-commit run --all-files` - passed
- `python -m build` - passed
- built-wheel metadata inspection - `ida-domain>=0.5.0` appears only under `Provides-Extra: ida`
- focused IDA compatibility and script tests with warnings promoted to errors - 19 passed, no warnings
- repository-wide compatibility and script-role sweep - no caller-side version selection or stale `ida_domain_analyze.py` references remain
- live IDA Domain 0.5 / IDA Professional 9.3 headless export using `tests/pe_export_label_test_xored` - report loaded successfully, matched IDA's function set and memory-image hash, and preserved `exported_test_function` at `0x401000`

Licensed GUI and idalib execution across every supported IDA release cannot be automated in this repository. Legacy 7.x/8.x and IDA 9.4 behavior therefore remains source- and mock-verified, while the Domain path has additionally been exercised end to end on IDA 9.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,21 @@ There is also a demo script:
 
 * analyze.py -- example usage: perform disassembly on a file or memory dump and optionally store results in JSON to a given output path.
 
+### IDA Pro integration
+
+SMDA can also turn an IDA-analyzed database into a SMDA report instead of running its own disassembly. Inside the IDA GUI, SMDA supports the existing IDAPython integrations for IDA 7.x, 8.x, and 9.x. On IDA 9.1 or newer, it prefers the higher-level [IDA Domain API](https://ida-domain.docs.hex-rays.com/) when the optional package is installed and otherwise falls back to IDAPython.
+
+Inside the IDA GUI, run `export.py` to export IDA's existing analysis to a `.smda` file next to the database. Run `ida_analyze.py` to have SMDA independently recover functions from the loaded bytes and add missing function starts and names back to IDA. This augmentation workflow is useful when IDA analyzes a raw or mapped buffer conservatively. Both scripts can be launched via *File -> Script file...*.
+
+For headless export of IDA's analysis (no GUI), use `ida_domain_export.py`:
+
+```
+python -m pip install "smda[ida]"
+python ida_domain_export.py /path/to/sample.i64 -o sample.smda
+```
+
+Headless export requires IDA 9.1+ and the optional `ida-domain>=0.5.0` dependency. Make sure `IDADIR` points at the IDA installation when it cannot be discovered automatically (see the [getting started guide](https://ida-domain.docs.hex-rays.com/getting_started/)). Standard SMDA installations do not include `ida-domain`.
+
 For Dalvik, the current scope is raw single-DEX inputs. APK, multi-dex container handling, and ODEX/VDEX/CDEX runtime-artifact analysis are not yet first-class workflows in SMDA.
 
 The code should be fully compatible with Python 3.8+.
--- a/export.py
+++ b/export.py
@@ -1,35 +1,34 @@
+"""Export the current IDA database to a SMDA report.
+
+For headless IDA 9.1+ export, use ``ida_domain_export.py`` instead.
+"""
+
 import json
 
 from smda.Disassembler import Disassembler
+from smda.ida.IdaInterface import getIdaSdkVersion
+from smda.SmdaConfig import SmdaConfig
 
 
 def detectBackend():
-    backend = ""
-    version = ""
     try:
-        import idaapi
-
-        backend = "IDA"
-        version = idaapi.IDA_SDK_VERSION
+        return ("IDA", getIdaSdkVersion())
     except ImportError:
-        pass
-    return (backend, version)
+        return ("", "")
 
 
 if __name__ == "__main__":
     BACKEND, VERSION = detectBackend()
     if BACKEND == "IDA":
-        from smda.ida.IdaInterface import IdaInterface
-
-        ida_interface = IdaInterface()
+        config = SmdaConfig()
+        disassembler = Disassembler(config, backend=BACKEND)
+        ida_interface = disassembler.disassembler.ida_interface
         binary = ida_interface.getBinary()
         base_addr = ida_interface.getBaseAddr()
-        DISASSEMBLER = Disassembler(backend=BACKEND)
-        REPORT = DISASSEMBLER.disassembleBuffer(binary, base_addr, architecture="aarch64")
-        output_path = ida_interface.getIdbDir()
-        output_filepath = output_path + "ConvertedFromIdb.smda"
+        REPORT = disassembler.disassembleBuffer(binary, base_addr)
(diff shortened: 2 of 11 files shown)


danielplohmann:
For the time being, we need to continue supporting older versions of IDA Pro as there will be legacy users, such as those that didn't move to the subscription model and will forever stay on something like IDA 8.4/8.5.

I would propose to still incorporate `ida_domain` as suggested for 9.1+ because it brings significant advantages for use in automation. 
I also do not agree with making `ida-domain` a core dependency because there are still plenty stand-alone users who don't need that package and we don't want to bloat them. We should keep a defensive availability check for the package and rely on it when it is there and IDA's version indicates compatibility.

r0ny123:
@danielplohmann I've updated the PR based on your suggestion. Should we consider dropping support for IDA 7.x?

r0ny123:
Earlier compatibility follow-up:

- Retained the existing IDAPython adapters for IDA 7.x, 8.4/8.5, and newer legacy workflows.
- Added a centralized selector that prefers IDA Domain only on SDK 9.1+ when the package is available, with automatic IDAPython fallback otherwise.
- Moved `ida-domain>=0.5.0` out of core dependencies into the optional `ida` extra.
- Isolated Domain and legacy API generations so future support changes are selector-table and adapter changes rather than caller-side version checks.
- Added headless Domain export plus compatibility-boundary, fallback, lifecycle, architecture, import, flowchart, byte, and xref tests.

The current PR description contains the consolidated implementation and validation details.

danielplohmann:
Thanks for the edits!

Some thoughts:

I mean, IDA 7.x is so far away by now, we could theoretically drop it, but it seems it is also not blocking more than the other legacy interfaces. So no real benefit in cleaning up here.

I think the `ida_analyze.py` and `ida_domain_analyze.py` now changed functionality: originally, the idea was that you would load something in IDA with auto-analysis disabled, then have SMDA do the disassembly and declare functions. I originally built that when IDA was still more or less incapable of finding code in x64 mapped / shellcode buffers (somewhere around late 7.x early 8.x versions?). I think by now this has gotten better a lot but you still sometimes have situations (especially when just loading buffers, i.e. with no PE header info available) where IDA would be conservative and not declare function starts etc. in a bunch of places, thus leaving manual work. For these, I then ran a SMDA disassembly run within IDA and had it "complete" the picture by declaring those functions more aggressively. 

I'm not sure if we want to rebuild this original functionality (since is less relevant nowadays) but if not I would recommend at least renaming `ida_domain_analyze.py` to `ida_domain_export.py` as that is much closer to what it actually does.

r0ny123:
Addressed this follow-up:

- Kept IDA 7.x support. The isolated adapters do not complicate the Domain selector, so removing them would not provide a practical benefit here.
- Restored `ida_analyze.py` to its original role: it now runs SMDA's own recursive disassembly over the bytes loaded in IDA and adds recovered functions/names back to the database. It no longer selects the IDA export backend.
- Renamed the headless utility to `ida_domain_export.py` and updated its function names, messages, help text, documentation, and regression tests to consistently describe export behavior.
- Deferred the optional `ida-domain` import until a real Domain database is opened. Lightweight backend probing no longer initializes IDA's SWIG runtime during test collection, while GUI fallback remains available when Domain cannot initialize.

I did not add a new headless mutating analyzer. The existing Domain utility serializes IDA's analysis, while database mutation introduces separate save/lifecycle semantics; keeping it explicitly export-only is the clearer boundary for this PR.

Validation includes the full test suite (471 passed, 1 skipped), lint/format/pre-commit, focused IDA tests with warnings promoted to errors (19 passed with no warnings), and an end-to-end `ida-domain 0.5.0` / IDA 9.3 export using the repository's `tests/pe_export_label_test_xored` fixture. The resulting report matched IDA's function set and memory-image hash and preserved the expected `exported_test_function` at `0x401000`.

danielplohmann (approved):

Issue #120, opened 2026-05-27
Use ELF machine type to report architecture and bitness consistently


r0ny123:
## Summary
`src/smda/utility/ElfFileLoader.py` has TODOs for adding machine types in `getBitness()` and `getCodeAreas()`, while `getArchitecture()` currently returns `intel` unconditionally. This can make loader metadata misleading for non-x86 ELF files even if the backend cannot disassemble those architectures yet.

## Source TODO
- `src/smda/utility/ElfFileLoader.py`: `getArchitecture()` / machine-type handling
- `src/smda/utility/ElfFileLoader.py`: `getBitness()`
- `src/smda/utility/ElfFileLoader.py`: `getCodeAreas()`

## Suggested implementation
Derive ELF architecture and bitness from `header.machine_type` using a central mapping, returning an explicit unsupported architecture or empty value instead of reporting `intel` for every ELF.
Add loader tests for x86, x86-64, and at least one recognized non-Intel ELF type so architecture, bitness, and code-area extraction stay aligned.

## Effort
Medium-high. The loader changes are contained, but behavior must be coordinated with backend support and any callers that assume ELF implies Intel.

## Related issue
This is a loader-specific slice of the broader architecture work tracked in #2.

r0ny123:
Follow-up after checking current `upstream/master`: this is partially addressed only for bitness. `ElfFileLoader.getBitness()` already reads `header.machine_type` and returns 64 for `X86_64` and 32 for `I386`.

What remains is consistent architecture reporting and broader machine-type handling: `getArchitecture()` still returns `intel` unconditionally for any ELF file, and `getCodeAreas()` is not coordinated with recognized/supported architectures. The issue should stay open until ELF architecture, bitness, and support status are derived from the same machine-type mapping.

r0ny123:
I've opened a PR addressing the remaining work here (centralizing ELF architecture/bitness/support behind one machine-type mapping so `getArchitecture()` no longer reports `intel` for every ELF): r0ny123/smda#64

Recognized non-Intel ELFs now report their real architecture and resolve to a controlled error report instead of being mis-analyzed as x86; x86/x86-64 behavior is unchanged.

PR #136, opened 2026-06-10
fix(loaders): derive ELF architecture from machine type

r0ny123:
## Summary

`ElfFileLoader.getArchitecture()` previously returned `"intel"` for **every** ELF file, so loader metadata was misleading for non-x86 inputs. This change derives ELF architecture, bitness, and backend-support status from a single machine-type mapping.

Addresses the remaining work in upstream issue [danielplohmann/smda#120](https://github.com/danielplohmann/smda/issues/120) (architecture reporting + centralized machine-type handling; bitness was already partly handled).

## Changes

- Add `_ELF_MACHINE_TYPES`: one mapping `lief.ELF.ARCH → (architecture, bitness, has_backend)`.
- Add `_resolve_elf_machine()` helper; `getArchitecture()` and `getBitness()` now both consume it.
- Width-ambiguous machine types (MIPS, RISC-V) resolve bitness from the ELF class (`identity_class`).
- Unknown machine types report `""` / `0` (explicit "unsupported") instead of `"intel"`.
- Drop the stale `# TODO add machine types` comments.

## Behavior

- **x86 / x86-64 unchanged** → still `("intel", 32/64)`.
- Recognized non-Intel ELFs (ARM/AArch64/PPC/SPARC/MIPS/RISC-V) now report their real architecture. SMDA only ships an Intel backend, so `Disassembler.initDisassembler` leaves the backend unset for these and `_disassemble` returns a **controlled error report** — it no longer mis-analyzes them as x86.
- A failed `lief.parse` now yields empty metadata (was `"intel"`), which likewise resolves to a controlled error report.

## Tests

`tests/testFileFormatParsers.py`: new mapping tests for x86, x86-64, AArch64 (recognized-but-unsupported), and MIPS/RISC-V (class-derived bitness), an unknown-machine-type case (AVR), and a failed-parse case. No new fixtures. The existing `bashlite_xored` (benign x86-64 `/bin/cat`) and `komplex` (x86-64) fixtures stay green.

## Validation

- `ruff check .` / `ruff format --check .` clean
- `make test` → 115 passed
- Reviewed by a Sonnet 4.6 subagent (extra-high reasoning); findings addressed (assert support flag, add RISC-V coverage).

Refs: #120

### DIFF
--- a/src/smda/utility/ElfFileLoader.py
+++ b/src/smda/utility/ElfFileLoader.py
@@ -44,11 +44,15 @@ def _resolve_elf_machine(elffile):
     width-ambiguous architectures (e.g. MIPS, RISC-V) it falls back to the ELF
     class so the reported bitness stays correct.
     """
-    if elffile is None:
+    # Guard against a missing header (failed parse, or an incomplete/mock
+    # object) so we report unsupported metadata instead of raising. We avoid a
+    # blanket try/except here so genuine parse/memory errors still surface.
+    if elffile is None or not hasattr(elffile, "header"):
         return "", 0, False
-    architecture, bitness, has_backend = _ELF_MACHINE_TYPES.get(elffile.header.machine_type, ("", 0, False))
+    header = elffile.header
+    architecture, bitness, has_backend = _ELF_MACHINE_TYPES.get(header.machine_type, ("", 0, False))
     if architecture and bitness == 0:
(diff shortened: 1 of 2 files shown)


danielplohmann (approved):
we should probably support the tests by actual binaries for the outlined architectures. I have full sets of mirai that could work well in that regard.

Issue #118, opened 2026-05-27
Centralize Mach-O CPU type mapping for architecture and bitness


r0ny123:
## Summary
`src/smda/utility/MachoFileLoader.py` repeats TODOs around adding machine types in `getArchitecture()`, `getBitness()`, and `getCodeAreas()`. The loader currently recognizes Intel and partially recognizes ARM, but ARM remains unsupported for disassembly and bitness/code-area behavior is split across separate methods.

## Source TODO
- `src/smda/utility/MachoFileLoader.py`: `getArchitecture()`
- `src/smda/utility/MachoFileLoader.py`: `getBitness()`
- `src/smda/utility/MachoFileLoader.py`: `getCodeAreas()`

## Suggested implementation
Introduce one CPU-type mapping helper for Mach-O architecture, bitness, and support status, then have all three methods consume that shared mapping.
Add parser-level tests for representative Mach-O CPU types, including unsupported-but-recognized architectures, so loader metadata stays consistent even when no backend exists yet.

## Effort
Medium-high. The mapping cleanup is moderate, but good coverage needs Mach-O fixtures or mocked LIEF objects and must align with the broader architecture support roadmap.

## Related issue
This is a loader-specific slice of the broader architecture work tracked in #2.

r0ny123:
Follow-up after checking current `upstream/master`: this is partially addressed. `MachoFileLoader.getArchitecture()` already recognizes x86/x86-64 as `intel` and ARM/ARM64 as `arm`; `getBitness()` also handles x86/x86-64 and explicitly rejects ARM64 as unsupported.

What remains is consistency and centralization. The CPU-type knowledge is still spread across methods, ARM handling is only partial, and `getCodeAreas()` is independent of any supported-architecture decision. A useful next step would be a shared Mach-O CPU mapping helper plus parser-level tests for supported and recognized-but-unsupported CPU types.

r0ny123:
Opened a PR centralizing the Mach-O CPU-type knowledge behind one mapping (architecture/bitness/support) so `getArchitecture()`/`getBitness()` no longer duplicate it or raise for recognized ARM/PPC types: r0ny123/smda#65

Recognized-but-unsupported CPU types now report accurate metadata and resolve to a controlled error report; x86/x86-64 behavior is unchanged.

PR #137, opened 2026-06-10
fix(loaders): centralize Mach-O CPU type mapping

r0ny123:
## Summary

`MachoFileLoader` duplicated CPU-type knowledge across `getArchitecture()`, `getBitness()`, and `getCodeAreas()`, with only partial ARM handling and per-method `NotImplementedError` raises. This centralizes that knowledge behind a single mapping + helper.

Addresses the remaining work in upstream issue [danielplohmann/smda#118](https://github.com/danielplohmann/smda/issues/118) (centralized CPU mapping + consistent metadata for recognized-but-unsupported architectures).

## Changes

- Add `_MACHO_CPU_TYPES`: one mapping `lief.MachO.Header.CPU_TYPE → (architecture, bitness, has_backend)`, built lazily via `_build_macho_cpu_types()` so the module still imports when LIEF is unavailable (consistent with the existing `LIEF_AVAILABLE` guard).
- Add `_resolve_macho_cpu()`; `getArchitecture()` and `getBitness()` both consume it.
- Drop the stale `# TODO add machine types` comments.

## Behavior

- **x86 / x86-64 unchanged** → `("intel", 32/64)`.
- Recognized non-Intel CPU types now report real metadata instead of raising: ARM → `("arm", 32)`, ARM64 → `("arm", 64)` (previously `getBitness` *raised* for ARM64), PPC/PPC64 → `("ppc", 32/64)`.
- Unknown CPU types report `""` / `0` instead of raising `NotImplementedError`.
- SMDA only ships an Intel backend, so unsupported architectures resolve to no disassembler → **controlled error report** (same end result the old raises produced, since `NotImplementedError` was already swallowed by `_handleDisassemblyException`).

## Tests

`tests/testFileFormatParsers.py`: mapping tests for X86, X86_64, ARM, ARM64, PPC, PPC64 (arch + bitness + support), an unknown-CPU case (SPARC), and a failed-parse case. No new fixtures. Existing `komplex` (x86-64) fixture stays green.

## Validation

- `ruff check .` / `ruff format --check .` clean
- `make test` → 115 passed
- Reviewed by a Sonnet 4.6 subagent (extra-high reasoning); LOW findings addressed (PPC32 coverage, falsy-guard consistency). Confirmed the raise→return change is safe (`NotImplementedError` was already caught into an error report).

Refs: #118

### DIFF
--- a/src/smda/utility/MachoFileLoader.py
+++ b/src/smda/utility/MachoFileLoader.py
@@ -48,7 +48,10 @@ def _resolve_macho_cpu(macho_file):
     """Return (architecture, bitness, has_backend) for a parsed Mach-O file,
     derived from its CPU type. Unknown CPU types report empty/unsupported
     metadata rather than guessing."""
-    if not macho_file:
+    # A FAT Mach-O parses to a lief.MachO.FatBinary, which has no ``header``;
+    # guard against that (and any incomplete object) so we report unsupported
+    # metadata instead of raising.
+    if not macho_file or not hasattr(macho_file, "header"):
         return "", 0, False
     return _MACHO_CPU_TYPES.get(macho_file.header.cpu_type, ("", 0, False))
 
--- a/src/smda/utility/MachoFileLoader.py
+++ b/src/smda/utility/MachoFileLoader.py
@@ -20,6 +20,39 @@ except ImportError:
 _NOT_PROVIDED = object()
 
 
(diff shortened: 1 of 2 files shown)


danielplohmann (approved):

danielplohmann:
had to manually fix linting for one line, thus the manual merge.

Issue #2, opened 2018-07-01
Support more architectures


danielplohmann:
The same method for CFG recovery should be applicable to other popular architectures such as ARM and MIPS(el).
Therefore, refactor and expand the disassembler part to be mostly architecture-agnostic.

sybernomad:
This would be awesome!

PR #140, opened 2026-06-10
refactor(intel): extract architecture-agnostic RecursiveDisassembler engine

r0ny123:
## Summary

First step toward issue [#2](https://github.com/danielplohmann/smda/issues/2) (*support more architectures*): make the recursive
CFG-recovery engine **architecture-agnostic** so the same method can later drive
ARM/MIPS, proving the abstraction by re-expressing the existing x86/x64 backend
through it with **zero behavior change**.

This PR is the refactor only — no new architecture is added yet (that is the
follow-up). It is deliberately scoped as its own reviewable unit.

## Motivation

The recursive disassembler was ~70% architecture-neutral CFG logic welded to
x86-specific decisions (mnemonic classification, capstone setup, operand parsing,
prologue/idiom handling). Adding ARM/MIPS first requires teasing those apart
behind a stable interface.

## What changed

- **New `common/RecursiveDisassembler.py`** — the architecture-neutral engine
  (recursive traversal, candidate/gap/tailcall orchestration, label/symbol
  resolution, code-ref/API plumbing). It imports only `common/` and delegates
  every architecture-specific decision to an injected backend.
- **New `common/arch/ArchBackend.py`** — the backend interface: collaborator
  factories (`createCapstone` / `createTfIdf` / `createCandidateManager` /
  `createAnalysisState` / jump-table + indirect-call analyzers), `probeBitness`,
  and `analyzeInstruction(...)` for per-instruction control-flow analysis.
- **New `intel/X86Backend.py`** — the x86/x64 specifics moved **verbatim**:
  capstone setup, the call/jmp/cond-jmp/loop/ret classification, and the
  push-ret / syscall-exit / alignment-after-call idioms.
- **`intel/IntelDisassembler.py`** collapses to a thin backwards-compatible
  wrapper subclassing `RecursiveDisassembler` and injecting `X86Backend`.

`FunctionCandidateManager`, `FunctionAnalysisState`, `definitions.py` and the
analyzers are **unchanged** — the engine receives them via backend factories.

## Changed behavior

**None.** This is a pure, behavior-preserving refactor.

## Validation

- `make lint` — clean (`ruff check .` + `ruff format --check .`, 100 files).
- `make test` — **111 passed, 79 subtests passed**, identical to the pre-refactor
  baseline. This includes `tests/testIntegration.py`, which asserts the exact
  serialized-report SHA-256 on the Asprox and Cutwail samples — matching hashes
  prove byte-for-byte identical disassembly output on real input.
- Timing A/B vs `master` on the Asprox sample (105 functions / 15706
  instructions): min/mean/median all within <1% (run-to-run noise). The single
  added per-instruction method dispatch is negligible.

## Follow-up

Adds an `AArch64Backend`, ELF `EM_AARCH64` detection, backend dispatch, and tests
against a generated AArch64 ELF fixture — no further engine surgery required.

## Summary by CodeRabbit

* **New Features**
  * Recursive control-flow graph recovery engine enhances binary disassembly accuracy and coverage.
  * Improved handling of indirect calls and jump table analysis.
  * Modular architecture backend framework enables support for additional processor architectures.

* **Refactor**
  * Intel x86/x64 disassembler refactored to use shared modular architecture components.


r0ny123:
**Note on ordering with #141:** #141 fixes four pre-existing CFG-recovery bugs that this refactor currently moves verbatim into the new engine/backend files. Suggested order: merge **#141 first**, then I'll rebase this PR over it — the conflicts land exactly on the moved lines and the resolution (the fixes expressed in the new layout) is already prepared, so it's mechanical. If this PR merges first instead, #141 will be reworked onto the new file layout — both orders work, the first is just less churn.

danielplohmann (approved):

danielplohmann (approved):

PR #161, opened 2026-07-01
fix(intel): preserve prefixed call blocks

r0ny123:
## Summary
- Normalize prefixed Intel mnemonics (`bnd`/`rep`/`lock`) before block-boundary checks in `FunctionAnalysisState.getBlocks()`, so a prefixed call's fallthrough instructions stay in the same block instead of being split off.
- Skip consecutive one-byte padding runs during gap scanning via `bytes.lstrip()` instead of a byte-by-byte loop. Behavior-preserving: it does not change which candidates get selected, only how fast the scanner walks past padding.

## Root cause / motivation
Capstone prepends mandatory prefixes to the mnemonic string it returns (e.g. `"bnd call"`, `"rep movsd"`, `"lock add"`). `getBlocks()` compared the raw mnemonic string directly against `CALL_MNEMONICS`/`END_MNEMONICS`, so a prefixed call or prefixed `ret` never matched either set. That caused the block-boundary logic to end a block at the wrong point (or fail to end it), dropping the fall-through instructions of a prefixed call from the function's reconstructed blocks. Every comparison now normalizes via `mnemonic.split(" ")[-1]` first.

## Changed behavior
- `src/smda/intel/FunctionAnalysisState.py`: `getBlocks()` now derives the base mnemonic before comparing against `CALL_MNEMONICS`/`END_MNEMONICS`.
- `src/smda/intel/FunctionCandidateManager.py`: gap-scan padding skip now advances past an entire run of one-byte padding bytes in one step instead of one byte at a time.

## Validation
- `make lint`
- `make test` (full suite green)
- `python -m pytest tests/testIntelDisassembler.py -q` (includes two new regression tests added in this change: one for prefixed-call block preservation, one for the padding-run skip)
- Additionally ran this branch through a 57-sample ground-truth accuracy harness (Plohmann malware corpus) against current `master`: recall/precision/F1 for both function and instruction recovery are byte-for-byte identical to `master` (this corpus's samples don't happen to contain `bnd`/`rep`/`lock`-prefixed call instructions, so the fix is a no-op here, but it closes a real gap for binaries that do use these prefixes without introducing any regression on this corpus).

### DIFF
--- a/src/smda/common/SmdaFunction.py
+++ b/src/smda/common/SmdaFunction.py
@@ -199,20 +199,25 @@ class SmdaFunction:
         architecture = self.smda_report.architecture if self.smda_report else ""
         if architecture == "dalvik":
             return sum(1 for block in self.blocks.values() for ins in block if ins.mnemonic.startswith("invoke-"))
-        return sum(1 for block in self.blocks.values() for ins in block if ins.mnemonic == "call")
+        # capstone prepends mandatory prefixes (bnd/rep/lock/...) to the mnemonic string
+        return sum(1 for block in self.blocks.values() for ins in block if ins.mnemonic.split(" ")[-1] == "call")
 
     @property
     def num_returns(self):
         architecture = self.smda_report.architecture if self.smda_report else ""
         if architecture == "dalvik":
             return sum(1 for block in self.blocks.values() for ins in block if ins.mnemonic.startswith("return"))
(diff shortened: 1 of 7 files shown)


danielplohmann (approved):

danielplohmann:
interesting that you ran into this as well, I just encountered some `bnd jmp` while investigating why MCRIT minhash and BSIM differed on some functions and it seems that bnd and other prefixes were not properly handled in older SMDA.  thx for the fix!

PR #162, opened 2026-07-02
fix(intel): gap-scan head/tail of raw memory dumps

r0ny123:
## Summary

Gap-scan the head and tail of raw memory dumps so functions lying entirely before the first or after the last already-discovered instruction are recovered. Two related commits fixing the same "tail-gap coverage" class:

1. **Create the head/tail gaps** for raw dumps (empty `_code_areas`).
2. **Scan the tail past the last instruction** — fix an off-by-one that made the gap walk abandon the entire tail region.

## Root cause

`FunctionCandidateManager.updateFunctionGaps()` only generated the head gap (before the first instruction) and tail gap (after the last instruction) by iterating `self._code_areas`. Raw memory dumps load from a buffer without section info, so `_code_areas` is empty and those regions were never turned into gaps — a function lying entirely before the first or after the last already-discovered instruction (e.g. a trailing `jmp`/thunk table) was unreachable by the gap pass.

Adding the synthetic full-image fallback exposed a second, pre-existing off-by-one: the tail gap was built as `[max_code, image_end)`, but `max_code` is the last instruction's address — itself in `code_map`. When the gap walk reaches `gap_pointer == max_code`, `nextGapCandidate()` sees it "already inside code map" and calls `getNextGap()`, whose strict `gap[0] > gap_pointer` test finds no further gap (the tail gap starts *at* `max_code`), returns the bitmask sentinel, and the scan terminates — abandoning the entire tail unscanned. The tail gap must start at `max_code + 1`, mirroring the interior-hole branch's `prev_ins + 1`.

## Changed behavior

- Fall back to the full mapped image `[base_addr, base_addr + binary_size)` as one synthetic code area when `_code_areas` is empty.
- Start that synthetic tail gap at `max_code + 1` so the gap-pointer lands past the last instruction and the tail is scanned. **Section-derived code areas (PE/ELF/Mach-O loads) keep the legacy behavior and are byte-for-byte unchanged** — the fix is scoped to the raw-dump fallback.
- AArch64 subclasses this manager and overrides only `nextGapCandidate`, so it inherits the fix; Dalvik/CIL have no gap-scanning candidate manager.

## Validation

Measured on the 57-sample malware memory-dump corpus (all loaded via `disassembleBuffer`, i.e. empty `_code_areas`), micro-averaged:

| metric | before | after |
|---|---|---|
| function recall | 98.061% | **98.814%** |
| function precision | 89.539% | 89.321% |
| function F1 | 93.606% | **93.828%** |
| missed functions | 425 | 260 |
| instruction recall | 98.898% | **99.033%** |
| instruction F1 | 94.572% | 94.571% |

Net **+165 functions recovered** (e.g. contopee's trailing import-thunk table, mewsei +44, kins +30, x64-prikorma +6). Function F1 up, instruction F1 flat. The additional "extra" functions are dominated by coherent compiler functions the IDA-snapshot ground truth did not label (verified iceix/feodo/rockloader/zlader — real prologues, calls, branches, `ret`), not garbage decode. Newly-surfaced tiny artifacts are < 10 instructions and ignored by MCRIT, consistent with the existing gap-scan "extract when in doubt" design.

Commands:
- `make lint` — clean
- `python -m pytest tests/testIntelDisassembler.py` — passes (regression test `test_function_gaps_cover_head_and_tail_without_code_areas` covers both gaps and the `max_code + 1` tail start)
- `make test` — 359 passed, 1 skipped (updated the asprox coverage/marshalling golden values to match the tail-scan recovery; the komplex Mach-O and mirai ELF loader fixtures are unchanged, confirming the section-derived path is untouched)

### DIFF
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -147,7 +147,18 @@ class FunctionCandidateManager:
         prev_ins = 0
         min_code = min(self.disassembly.code_map) if self.disassembly.code_map else self.getBitMask()
         max_code = max(self.disassembly.code_map) if self.disassembly.code_map else 0
-        for code_area in self._code_areas:
+        # Raw memory dumps are loaded without section info, so self._code_areas is empty; fall back to
+        # the full mapped image so the head (before the first instruction) and tail (after the last
+        # instruction) still get gap-scanned. Without this, functions that lie entirely before the
+        # first or after the last already-discovered instruction (e.g. trailing jmp/thunk tables) are
+        # never reached by the gap pass.
+        code_areas = self._code_areas or [
+            [
+                self.disassembly.binary_info.base_addr,
+                self.disassembly.binary_info.base_addr + self.disassembly.binary_info.binary_size,
+            ]
+        ]
+        for code_area in code_areas:
             if code_area[0] < min_code < code_area[1] and min_code != code_area[0]:
                 gaps.append([code_area[0], min_code, min_code - code_area[0]])
             if code_area[0] < max_code < code_area[1] and max_code != code_area[1]:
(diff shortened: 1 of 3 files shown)


danielplohmann (approved):

danielplohmann:
I think originally this was a trade off because sometimes the PE header / artefacts before the first function would cause issues, as well as disassembling into the .rdata/.data to follow the code section. But I guess this only ever really matters for shellcode-style buffers to be processed.

PR #168, opened 2026-07-09
fix(intel): skip chained/forwarded function-candidate table entries

r0ny123:
## Summary

Fixes three sibling instances of the same bug class: a binary-format metadata table row (PE `.pdata`
exception table, PE export table, PE/COFF symbol table) was trusted as an independent local function
start without checking its own "not defined here" flag. Two are bugs (false-positive function starts /
wrong addresses), one is an accuracy improvement (recovers real precision loss from an existing
recall-oriented feature). The accuracy improvement below (PPV/TPR/F1 table) was measured on one
specific evaluation dataset — the **Bao/byteweight `msvc10-64` corpus** (68 PE x64 binaries) — not
across the full accuracy-eval suite; see the caveats in Bug 1 for what was and wasn't affected
elsewhere.

## Bug 1 (improvement): `.pdata` chained unwind-info fragments counted as new functions

**What:** `FunctionCandidateManager.locateExceptionHandlerCandidates()` reads the PE x64 `.pdata`
section (`RUNTIME_FUNCTION` table) and injects every entry's `BeginAddress` as a near-guaranteed
function-start candidate. Some `RUNTIME_FUNCTION` entries are secondary fragments of another
function (e.g. a split-off cold/epilogue chunk) — their `UNWIND_INFO` is marked with
`UNW_FLAG_CHAININFO` specifically so tooling does not mistake them for an independent function. This
was not being checked, so every chained fragment was counted as a brand-new (false-positive)
function.

**How fixed:** read the `UNWIND_INFO` header byte at each entry's `UnwindInfoAddress` and skip the
entry when the `CHAININFO` flag (bit `0x4` of the Flags nibble) is set.

**Measured impact — Bao/byteweight `msvc10-64` dataset only** (68 PE x64 binaries, function-level
macro PPV/TPR/F1; this specific corpus was chosen because it showed the largest PPV regression
against an older SMDA baseline, which is what led to this fix):

| | PPV | TPR | F1 |
|---|---|---|---|
| before | 95.38 | 99.72 | 97.46 |
| after | 98.70 | 99.73 | 99.21 |

TPR is effectively unchanged (99.72 → 99.73) — the fix removes false positives without giving back
any of the feature's genuine recall win. Confirmed no effect on 32-bit configs (feature is 64-bit
only) or raw-memory-dump corpora (no section table present, so `.pdata` is never found).

## Bug 2: PE export forwarders/extern entries treated as local functions

**What:** `PeSymbolProvider.parseExports` and `RustSymbolProvider._update_pe` iterated
`lief_binary.exported_functions` (LIEF's abstract, format-agnostic export view) and trusted every
entry's address. A forwarded or extern export (e.g. an API set forwarder, or a DLL re-exporting
another DLL's function) has no local address — LIEF's PE-specific `ExportEntry` marks these via
`is_extern`/`is_forwarded` and sets `.address` to `0`, but the abstract wrapper does not expose those
flags at all, so every forwarder collapsed onto `base_addr + 0` and was recorded as a function there.

**How fixed:** switched to the PE-specific `lief_binary.get_export().entries`, which exposes
`is_extern`/`is_forwarded`, and skip entries with either flag set.

## Bug 3: COFF symbols with no defining section treated as local functions, wrong section base for the rest

**What:** `PeSymbolProvider.parseSymbols` and `RustSymbolProvider._update_pe` cached the virtual
address of the *first* executable section and added every `FUNCTION`-typed COFF symbol's `.value` to
that single base — regardless of which section the symbol actually belongs to, and without checking
whether the symbol is even defined in this file. A COFF symbol with `section_idx` `0`/`-1`/`-2`
(undefined-external/absolute/debug) is not a locally defined function, and its `.value` is not a
usable in-image offset; symbols legitimately defined in any section other than the first code section
resolved to the wrong address entirely.

**How fixed:** use each symbol's own `.section` (skip when `None`, i.e. not locally defined) and
resolve its address from that section's virtual address instead of a single cached first-code-section
value.

## Systemic pattern

All three are the same root-cause pattern: an unconditionally-trusted binary metadata table entry
that has its own format-native "not an independent local function" signal, unchecked. A tree-wide
sweep covering every symbol/export/exception-table consumer (aarch64, ELF, Mach-O, PDB, Go,
Delphi-KB, CIL/Dalvik label providers) found these are the only real instances; the rest were
confirmed benign for structural reasons specific to those formats (e.g. ELF has no forwarder concept
and already filters by symbol type; Mach-O addresses are gated by a code-area check).

## Validation

```
make lint
make test   # 378 passed, 1 skipped
pytest tests/testPdataExtraction.py -q
pytest tests/testPeSymbolProvider.py tests/testRustSymbolProvider.py -q
pytest tests/testFileFormatParsers.py tests/testPeFileLoader.py -q
```

All targeted and full-suite tests pass. Regression tests added for each of the three fixes.

### DIFF
--- a/src/smda/common/labelprovider/PeSymbolProvider.py
+++ b/src/smda/common/labelprovider/PeSymbolProvider.py
@@ -53,7 +53,14 @@ class PeSymbolProvider(AbstractLabelProvider):
     def parseExports(self, lief_binary, base_addr=None):
         active_base = self._resolve_base_addr(lief_binary, base_addr)
         function_symbols = {}
-        for function in lief_binary.exported_functions:
+        export = lief_binary.get_export()
+        if export is None:
+            return function_symbols
+        for function in export.entries:
+            if function.is_extern or function.is_forwarded:
+                # forwarder/extern entries redirect to another module's export and have no
+                # local address (LIEF sets .address to 0 for them) - not a local function.
+                continue
             function_name = ""
             with contextlib.suppress(UnicodeDecodeError, AttributeError):
                 # here may occur a LIEF exception that we want to skip ->
@@ -65,25 +72,20 @@ class PeSymbolProvider(AbstractLabelProvider):
 
     def parseSymbols(self, lief_binary, base_addr=None):
         active_base = self._resolve_base_addr(lief_binary, base_addr)
-        # find VA of first code section
         function_symbols = {}
-        code_base_address = None
-        for section in lief_binary.sections:
-            if section.characteristics & 0x20000000:
-                code_base_address = active_base + section.virtual_address
-                break
-        if code_base_address is None:
-            return function_symbols
         for symbol in lief_binary.symbols:
             if hasattr(symbol.complex_type, "name") and symbol.complex_type.name == "FUNCTION":
+                if symbol.section is None:
+                    # section_idx 0/-1/-2 (undefined-external/absolute/debug): not a locally
+                    # defined function, its value is not a usable in-image offset.
+                    continue
                 function_name = ""
                 with contextlib.suppress(UnicodeDecodeError, AttributeError):
                     # here may occur a LIEF exception that we want to skip ->
                     # UnicodeDecodeError: 'utf-32-le' codec can't decode bytes in position 0-3: code point not in range(0x110000)
                     function_name = symbol.name
                 if function_name and all(ord(c) in range(0x20, 0x7F) for c in function_name):
-                    # for some reason, we need to add the section_offset of .text here
-                    function_offset = code_base_address + symbol.value
+                    function_offset = active_base + symbol.section.virtual_address + symbol.value
                     if function_offset not in function_symbols:
                         function_symbols[function_offset] = function_name
         return function_symbols
--- a/src/smda/common/labelprovider/RustSymbolProvider.py
+++ b/src/smda/common/labelprovider/RustSymbolProvider.py
@@ -143,7 +143,12 @@ class RustSymbolProvider(AbstractLabelProvider):
         """Process PE binary symbols for Rust demangling."""
         active_base = resolve_pe_base_addr(lief_binary, base_addr)
         # Parse PE exports
-        for function in lief_binary.exported_functions:
+        export = lief_binary.get_export()
+        for function in export.entries if export is not None else []:
+            if function.is_extern or function.is_forwarded:
+                # forwarder/extern entries redirect to another module's export and have no
+                # local address (LIEF sets .address to 0 for them) - not a local function.
+                continue
             try:
                 try:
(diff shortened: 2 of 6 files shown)


danielplohmann (approved):

r0ny123:
@danielplohmann Following up on the post-merge CI failure on the Malpedia benchmark (https://github.com/danielplohmann/smda/actions/runs/29093435396/attempts/1#summary-86365953031) — I investigated both flagged issues using the `malpedia.zip` samples and isolated worktrees at `e0e71ee` (pre-fix) and `8466507` (post-fix, current master).

## Correctness regression (2/153 files) — traced, not a real bug

Reproduced both locally, exact match to CI:
- `win.konni_2017-07-04_..._unpacked_x64`: 456 → 436 functions (-20)
- `win.konni_2021-12-20_..._unpacked`: 291 → 289 functions (-2)

I parsed each lost address's `RUNTIME_FUNCTION`/`UNWIND_INFO` entry (against SMDA's own internally-mapped buffer, not raw file bytes — file-alignment vs section-alignment differences will give the wrong byte if you read straight from the file) and followed its chain pointer:

- Sample 2 (both losses): both resolve cleanly to an already-counted parent function.
- Sample 1 (20 losses): 17 resolve directly to an already-counted parent; 1 more resolves via a 2-hop chain (chains to another `CHAININFO` entry, which itself resolves to a counted parent) — 18/20 confirmed. The remaining 2 are ambiguous in my quick verification (one resolves to a misaligned/odd address, which looks like an artifact of my manual chain-offset parsing rather than the shipped fix itself, since the fix doesn't do multi-hop resolution — it just skips `CHAININFO`-flagged entries unconditionally per the ABI, regardless of what they chain to).

This matches the same pattern the fix was built on (from the Bao/msvc10-64 accuracy corpus: ~99% of `.pdata`-attributable extras resolved back to an already-discovered parent function). The function-count-based correctness gate can't distinguish "lost a real function" from "correctly stopped double-counting a chained fragment as a separate function" — a count-only diff will flag this class of precision fix as a regression even when working as intended. Happy to add ground-truth-aware validation to the gate if useful, or this could just be accepted as expected behavior for this fix.

## Performance regression (-11.34% median) — didn't reproduce locally

Profiled the larger sample (456 functions) pre-fix vs post-fix with `cProfile`: 0.569s vs 0.575s — about 1% difference, not the reported -11.34%. Since base and PR are timed on separate CI runners (as the report itself notes), this looks like runner-to-runner variance rather than a real cost from the change. Could you re-run the benchmark workflow once more to confirm whether the slowdown persists? If it's consistently reproducible I'm happy to dig further, but a single run's -11% didn't show up in local profiling of the actual flagged files.

PR #174, opened 2026-07-10
refactor(core): consolidate import-stub helpers and wire has_backend

r0ny123:
## Summary

Follow-up cleanup for #169, addressing items that were flagged during that PR's review and intentionally deferred to keep #169 focused on Mach-O/import-stub parity.

**Note:** this branch stacks on #169 (`fix/macho-intel-stub-parity`). GitHub does not allow a fork-only branch as the base of a cross-repo pull request, so this PR's diff will include #169's commits until #169 merges. The two commits unique to this PR are:

- `refactor(core): hoist shared import-stub range helpers into ArchBackend`
- `fix(core): report the unsupported architecture in no-backend error reports`

## Changes

- **Hoist duplicated helpers.** `_getPltRanges`/`_getMachoStubRanges`/`_getImportStubRanges` were byte-for-byte identical between `X86Backend` and `AArch64Backend`. Moved onto the shared `ArchBackend` base class so both backends inherit one implementation; each backend keeps only its architecture-specific slot-decode logic (`_resolveImportSlot` / `_resolvePltGotSlot`).
- **Consume `has_backend`.** It was exposed by every format loader (#169) but never read anywhere, so a buffer with no disassembly backend fell through to a generic `"Disassembler backend not initialized."` error, indistinguishable from a real internal precondition failure. `Disassembler._disassemble` now checks `has_backend` and raises a message naming the specific unsupported architecture, which flows into the controlled error report.

A third item flagged during #169's review — `RustSymbolProvider`'s Mach-O support living in a "loaders/stub-parity" branch — was reviewed and found to already be adequately isolated (its own commit, with the `_R`/`__R` vs. legacy `_ZN` detection tradeoff documented inline in code). No further change was needed there.

## Tests and validation

- `make lint`
- `python -m ruff format --check .`
- `make test` — 402 passed, 1 skipped
- Extended the existing ELF and Mach-O no-backend controlled-error tests to assert the new architecture-specific message.

### DIFF
--- a/src/smda/Disassembler.py
+++ b/src/smda/Disassembler.py
@@ -250,6 +250,8 @@ class Disassembler:
         if self.disassembler:
             self.disassembly = self.disassembler.analyzeBuffer(binary_info, self._callbackAnalysisTimeout)
             return SmdaReport(self.disassembly, config=self.config)
+        if not binary_info.has_backend:
+            raise RuntimeError(f"No disassembly backend available for architecture '{binary_info.architecture}'.")
         raise RuntimeError("Disassembler backend not initialized.")
 
     def _createErrorReport(self, start, exception):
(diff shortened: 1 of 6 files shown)


danielplohmann:
#169 merged, I guess you can FF and then I'll happily merge this refactoring PR. :)

r0ny123:
Sure.

danielplohmann (approved):

PR #177, opened 2026-07-11
feat(intel): extended AMD64 prologues and syscall-exit detection

### LINKED ISSUE #170: feat(intel): recognize extended GCC/Clang AMD64 prologue family and endbr64 prefix
### Summary
Modern GCC/Clang x86-64 functions open with one of a small set of prologues that SMDA's
intel backend does not currently treat as function starts:
- `endbr64` (`F3 0F 1E FA`) — CET landing pad, emitted first by modern toolchains
- `48 89 5C 24 ??` — `mov [rsp+imm], rbx` (callee-saved save w/ SP-relative disp)
- `48 83 EC ??` — `sub rsp, imm` (stack allocation)
- `41 57 41 56` — `push r15; push r14` (callee-saved push pair)

These should seed/score function candidates the way the existing classic prologues do.

### Current state
- `COMMON_PROLOGUES`/`DEFAULT_PROLOGUES` in `src/smda/intel/definitions.py:50-129` hold only
  *exact* byte sequences (the push-ebp family). The new patterns have a variable immediate
  (`??`), which the exact-match table cannot express.
- `FunctionCandidate` reads a fixed 5-byte window (`src/smda/intel/FunctionCandidate.py:15`)
  and matches via `hasCommonFunctionStart()`/`getFunctionStartScore()` (lines 69-84) against
  `COMMON_PROLOGUES`.
- `FunctionCandidateManager.locatePrologueCandidates()`
  (`src/smda/intel/FunctionCandidateManager.py:609-624`) scans `DEFAULT_PROLOGUES` with exact
  `re.finditer`.
- `endbr64` is currently only a *NOP* for the escaper
  (`src/smda/intel/IntelInstructionEscaper.py:209-210`); `0xf3` appears in the 1-byte frequency
  table (`definitions.py:116`) so endbr64-prefixed functions get a weak score but are not
  recognized as prologue-led.

### Desired behavior
Function starts matching these patterns (including when prefixed by `endbr64`) are recognized
as common function starts and scored accordingly.

### Implementation hints
- Introduce a *masked* prologue table for 64-bit (e.g. `(mask, value, score)` tuples of length
  3/5) alongside `COMMON_PROLOGUES`, and update `FunctionCandidate.hasCommonFunctionStart`/
  `getFunctionStartScore` and `locatePrologueCandidates` to match with `(word & mask) == value`.
- For `endbr64` coverage, strip a leading `F3 0F 1E FA` (4 bytes) from the candidate window
  *before* prologue matching, so `endbr64; <prologue>` is recognized (this also covers the
  upstream's `endbr64` bullet).
- Keep scores consistent with the existing 64-bit table (prologue-led starts should score high,
  ~33-50 like `push rbp`).
- Beware the candidate cap (`MAX_FUNCTION_CANDIDATES`) and `locatePrologueCandidates` ordering
  in `locateCandidates()` (`FunctionCandidateManager.py:522-541`).

### Acceptance
- A function beginning `endbr64; push rbp; mov rbp, rsp` (or `endbr64; sub rsp, 0x10`) is scored
  as a common function start.
- `push r15; push r14` and `mov [rsp+8], rbx` / `sub rsp, imm` openings raise a candidate's
  `hasCommonFunctionStart()`/`getFunctionStartScore()` to non-zero.
- No regression on the frozen golden x86/x64 fixtures (run `make test`).

### LINKED ISSUE #171: feat(intel): end functions on exit_group and int 0x80 syscalls
### Summary
`X86Backend` already ends a function when it reaches `mov eax, 60; syscall` (Linux `exit`), but
not the other process-terminating syscalls:
- `mov eax, 231; syscall` — `exit_group`
- `mov eax, 1; int 0x80` — 32-bit ABI `exit` via `int 0x80`
- `mov eax, 252; int 0x80` — 32-bit ABI `exit_group` via `int 0x80`

### Current state
- `src/smda/intel/X86Backend.py:341-348`: syscall branch calls `_resolveSyscallNumber(...)` and
  ends the function only on `syscall_number == 60`.
- `_resolveSyscallNumber()` (lines 237-286) backtracks the block for `mov`/`movabs` into
  `rax`/`eax`/`ax`/`al`/`ah` and is 32/64-bit aware (`bitness` param).
- `int`/`int3`/`hlt` are in `SYSCALL_INS` (`X86Backend.py:34`) but `int 0x80` is not given
  syscall-number semantics — `int3`/`hlt` just end the block (`X86Backend.py:335-336`).

### Desired behavior
The four terminating sequences above each cleanly terminate the containing function (mark
block-ending + sanely-ending), matching the existing `exit` (`60`) handling.

### Implementation hints
- Extend the syscall-number check at `X86Backend.py:343` to a small set `{60, 231}` for the
  `syscall` mnemonic (64-bit `rax`/`eax`; 32-bit uses `eax` only — `_resolveSyscallNumber`
  already handles this).
- Add an `int 0x80` branch: when `i_mnemonic == "int"` with operand `0x80`, resolve the eax
  value the same way (note: `int 0x80` is the 32-bit ABI even on 64-bit processes, so the
  syscall number register is `eax` and numbers are `1`/`252`). Reuse `_resolveSyscallNumber`
  (it already treats `eax` as a target reg) but force the 32-bit register set, or add a sibling
  `_resolveInt80Number`.
- End via the existing `_analyzeEndInstruction(state)` pattern (lines 232-235).
- Do **not** end on other `int` vectors or non-terminating syscalls.

### Acceptance
- Functions ending in `mov eax, 60; syscall`, `mov eax, 231; syscall`, `mov eax, 1; int 0x80`,
  `mov eax, 252; int 0x80` terminate at that instruction.
- Other `int`/`syscall` numbers and `int3`/`hlt` behavior unchanged.
- `make test` passes (golden fixtures unaffected — these are tail-of-function terminators).

r0ny123:
## Summary

Two related intel-backend FEP/CFG-accuracy improvements, implemented together:

- Closes #170: recognize an extended GCC/Clang AMD64 prologue family as common function starts.
- Closes #171: end functions on `exit_group` and `int 0x80` process-terminating syscalls.

## #170 — extended AMD64 prologue family

Modern GCC/Clang x86-64 functions often open with an `endbr64` CET landing pad, a
`push r15; push r14` callee-saved pair, or an immediate-wildcarded `mov [rsp+imm8], rbx` /
`sub rsp, imm8` stack-frame opener. None of these were recognized as common function starts,
so the existing FEP heuristics scored them the same as arbitrary bytes.

- `endbr64` is stripped from the candidate window before prologue matching, so
  `endbr64; <prologue>` is recognized without inventing a fake score for a bare, unmatched
  `endbr64`.
- A new masked (mask/value) matching path sits alongside the existing exact-byte
  `COMMON_PROLOGUES` table for the two immediate-wildcarded patterns.
- `sub rsp, imm8` is excluded from whole-binary FEP seeding (though it still contributes to
  *scoring* an already-discovered candidate): seeding it unconditionally manufactured false
  function starts inside real functions in the `mirai_x64_xored` golden fixture (151 -> 158
  functions), because it's a common mid-function idiom (e.g. `sub rsp, 8; call X; add rsp, 8;
  ret`), not just a prologue. Verified byte-for-byte identical recovery against the fixture
  baseline after the fix.

## #171 — exit_group / int 0x80 exit detection

`X86Backend` already ended a function at `mov eax, 60; syscall` (Linux `exit`) but not
`exit_group` (231), nor either exit reached via the 32-bit-ABI `int 0x80` gate (`eax=1` exit,
`eax=252` exit_group), which a 64-bit process can still use.

The new `int 0x80` branch resolves `eax` via the existing backtracking helper, but with
`bitness=64` regardless of the `int 0x80` op_str's 32-bit ABI semantics: that register set
(`rax`/`eax`/`ax`/`al`/`ah`) is a strict superset of the 32-bit one, so it also catches a
full-width `mov rax, N` write that a 64-bit binary may still use before dropping into
`int 0x80` — forcing 32-bit semantics there would drop `rax` from the clobber set and silently
walk past that write, causing a missed or misresolved exit (confirmed empirically). The
resolved value is masked to 32 bits before the exit-number check, since `int 0x80` only reads
the low 32 bits of `rax`.

## Validation

```
ruff check .
ruff format --check .
make test
```

All green (389 passed, 1 skipped), including the frozen `tests/*_xored` golden fixtures.

New/updated tests in `tests/testIntelDisassembler.py` cover: scoring of all four new prologue
patterns (including the `endbr64` + prologue combinations and the bare-`endbr64`-scores-zero
case), 32-bit exclusion of the masked patterns, FEP seeding of the seedable patterns and
non-seeding of `sub rsp, imm8`, `exit_group` via `syscall`, `exit`/`exit_group` via `int 0x80`
(including the full-width `rax` write and the wider-than-32-bit-resolution truncation case),
and negative guards for unrelated syscall numbers / `int` vectors.

### DIFF
--- a/src/smda/intel/FunctionCandidate.py
+++ b/src/smda/intel/FunctionCandidate.py
@@ -1,10 +1,34 @@
 from binascii import hexlify
 
-from .definitions import COMMON_PROLOGUES
+from .definitions import COMMON_PROLOGUES, ENDBR64_BYTES, MASKED_PROLOGUES_64
 
 # Hoisted: prologue lengths are checked longest-first on every candidate scoring call.
 # Pre-sort once at import time instead of re-sorting per call (hot path during CFG recovery).
 _COMMON_PROLOGUE_LENGTHS = sorted((int(k) for k in COMMON_PROLOGUES), reverse=True)
+
+
+def _buildMaskedPrologueTable(templates):
+    # scoring applies regardless of is_seedable: a candidate already found by other means
+    # (call ref, gap, symbol) still deserves credit for looking like a prologue.
+    table = []
+    for template, score, _is_seedable in templates:
+        mask = int.from_bytes(bytes(0xFF if b is not None else 0x00 for b in template), "big")
+        value = int.from_bytes(bytes(b if b is not None else 0x00 for b in template), "big")
+        table.append((len(template), mask, value, score))
+    return table
+
+
+# (length, mask, value, score) tuples, matched via (window_as_int & mask) == value.
+_MASKED_PROLOGUES_64 = _buildMaskedPrologueTable(MASKED_PROLOGUES_64)
+_ENDBR64_LEN = len(ENDBR64_BYTES)
+# lengths to probe longest-first, so a specific multi-byte match always wins over a weak
+# single-byte fallback, regardless of whether it comes from COMMON_PROLOGUES or a masked entry.
+_ALL_PROLOGUE_LENGTHS = sorted(
+    {*_COMMON_PROLOGUE_LENGTHS, *(length for length, *_ in _MASKED_PROLOGUES_64)}, reverse=True
+)
+# longest byte window a candidate ever needs to see: the widest prologue match, plus a
+# potential leading endbr64 to strip before re-matching the remainder.
+_PROLOGUE_WINDOW_SIZE = _ALL_PROLOGUE_LENGTHS[0] + _ENDBR64_LEN
 
 
 class FunctionCandidate:
@@ -12,7 +36,7 @@ class FunctionCandidate:
         self.bitness = binary_info.bitness
         self.addr = addr
         rel_start_addr = addr - binary_info.base_addr
-        self.bytes = binary_info.binary[rel_start_addr : rel_start_addr + 5]
+        self.bytes = binary_info.binary[rel_start_addr : rel_start_addr + _PROLOGUE_WINDOW_SIZE]
         self.lang_spec = None
         # set, not list: addCallRef / removeCallRefs do membership tests in the inner
         # CFG-recovery loop. Order is never read externally (only len + truthiness).
@@ -66,21 +90,34 @@ class FunctionCandidate:
                 self._confidence = round(weighted_confidence, 3)
         return self._confidence
 
+    def _lookupPrologueScore(self, byte_window):
+        # longest-first, and masked entries checked alongside exact ones at the same length,
+        # so a specific multi-byte match always wins over a weak single-byte fallback.
+        for length in _ALL_PROLOGUE_LENGTHS:
+            window_slice = byte_window[:length]
+            if len(window_slice) < length:
+                continue
+            prologue_score = COMMON_PROLOGUES.get(f"{length}", {}).get(self.bitness, {}).get(window_slice)
+            if prologue_score is not None:
+                return prologue_score
+            if self.bitness == 64:
+                window_int = int.from_bytes(window_slice, "big")
+                for masked_length, mask, value, score in _MASKED_PROLOGUES_64:
+                    if masked_length == length and (window_int & mask) == value:
+                        return score
+        return None
+
     def hasCommonFunctionStart(self):
-        for length in _COMMON_PROLOGUE_LENGTHS:
-            byte_sequence = self.bytes[:length]
-            if byte_sequence in COMMON_PROLOGUES[f"{length}"][self.bitness]:
-                return True
-        return False
+        return self.getFunctionStartScore() > 0
 
     def getFunctionStartScore(self):
         if self.function_start_score is None:
-            for length in _COMMON_PROLOGUE_LENGTHS:
-                byte_sequence = self.bytes[:length]
-                if byte_sequence in COMMON_PROLOGUES[f"{length}"][self.bitness]:
-                    self.function_start_score = COMMON_PROLOGUES[f"{length}"][self.bitness][byte_sequence]
-                    break
-            self.function_start_score = self.function_start_score if self.function_start_score else 0
+            # endbr64 is a CET landing pad prefixing the real prologue, not a prologue itself:
+            # strip it before matching so the prologue behind it is scored, not the raw bytes.
+            byte_window = self.bytes
+            if self.bitness == 64 and self.bytes[:_ENDBR64_LEN] == ENDBR64_BYTES:
+                byte_window = self.bytes[_ENDBR64_LEN:]
+            self.function_start_score = self._lookupPrologueScore(byte_window) or 0
         return self.function_start_score
 
     def addCallRef(self, source_addr):
--- a/src/smda/intel/FunctionCandidate.py
+++ b/src/smda/intel/FunctionCandidate.py
@@ -1,34 +1,14 @@
 from binascii import hexlify
 
-from .definitions import COMMON_PROLOGUES, ENDBR64_BYTES, MASKED_PROLOGUES_64
+from .definitions import COMMON_PROLOGUES, ENDBR64_BYTES
 
 # Hoisted: prologue lengths are checked longest-first on every candidate scoring call.
 # Pre-sort once at import time instead of re-sorting per call (hot path during CFG recovery).
 _COMMON_PROLOGUE_LENGTHS = sorted((int(k) for k in COMMON_PROLOGUES), reverse=True)
-
-
-def _buildMaskedPrologueTable(templates):
-    # scoring applies regardless of is_seedable: a candidate already found by other means
-    # (call ref, gap, symbol) still deserves credit for looking like a prologue.
-    table = []
-    for template, score, _is_seedable in templates:
-        mask = int.from_bytes(bytes(0xFF if b is not None else 0x00 for b in template), "big")
-        value = int.from_bytes(bytes(b if b is not None else 0x00 for b in template), "big")
-        table.append((len(template), mask, value, score))
-    return table
-
-
-# (length, mask, value, score) tuples, matched via (window_as_int & mask) == value.
-_MASKED_PROLOGUES_64 = _buildMaskedPrologueTable(MASKED_PROLOGUES_64)
 _ENDBR64_LEN = len(ENDBR64_BYTES)
-# lengths to probe longest-first, so a specific multi-byte match always wins over a weak
-# single-byte fallback, regardless of whether it comes from COMMON_PROLOGUES or a masked entry.
-_ALL_PROLOGUE_LENGTHS = sorted(
-    {*_COMMON_PROLOGUE_LENGTHS, *(length for length, *_ in _MASKED_PROLOGUES_64)}, reverse=True
-)
 # longest byte window a candidate ever needs to see: the widest prologue match, plus a
 # potential leading endbr64 to strip before re-matching the remainder.
-_PROLOGUE_WINDOW_SIZE = _ALL_PROLOGUE_LENGTHS[0] + _ENDBR64_LEN
+_PROLOGUE_WINDOW_SIZE = _COMMON_PROLOGUE_LENGTHS[0] + _ENDBR64_LEN
 
 
 class FunctionCandidate:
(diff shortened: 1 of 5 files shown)


danielplohmann:
how #171 is addressed is good, but I would suggest revisions for how #170 is addressed.

I think the `endbr64` CET landing pad is very good and the `push r15; push r14` also seems strong, albeit occurring way less often.

With the other two, I'd suggest something else, based on looking at the x64 Windows data set from Bao and some current libc which validated the above heuristics.

While `sub rsp, imm` appears in 27.5% of libc prologues, as a standalone byte pattern it seems a bit too generic (measured 17.3% prologue-specific in MSVC) - I know the idea is to only verify if it was identified as a potential prologue else where. But I found the following: `4053 push rbx ; 4883ec?? sub rsp, ??` - that's a really strong pattern on MSVC and could be coded as a regular 5 byte prologue pattern (40534883ec), automatically treating the imm8 as wildcarded because the wildcard then simply drops out of the current 5 byte window. 

For `mov [rsp+imm], rbx`, this is absent as a GCC prologue — System V has no shadow/home space, so GCC/Clang essentially never open this way. For that Windows data set, it's the most common first instruction and 73% of its overall occurrences are at function starts. So that's still quite strong as a signal, could work when using just the first 4 bytes (again, no wildcard needed that way). As a result we could possibly drop the wildcard mechanism using None and also the flagging that the instruction should not be used for seeding, thus avoiding a bit of complexity that would otherwise be introduced for just a special case. Still would need to check if this drives FP up significantly, might also just assign it a notably lower confidence score.

For completeness a bit more on the statistics for the shadow space as per Hy3:

```
Here are the exact MSVC numbers:
Specific pattern: 48895c24?? mov qword ptr [rsp + ??], rbx
- At function start: 21,699
- Overall (anywhere in the 7.59M instructions): 29,616
- Precision (when you see it, is it a prologue opener?): 73.3% — i.e. 26.7% (7,917) are mid-function register spills, not prologues.
- Recall: it's the #1 MSVC first instruction, opening 21.8% of all 99,616 functions.
The whole shadow-space-save family mov (qword|dword) ptr [rsp + ??], <reg>:
- At start: 43,161 (43.3% of functions) — great recall.
- Overall: 329,637 → precision only 13.1% (87% are mid-function spills).
Should we use it?
Yes, but as a weighted positive feature, not a hard trigger.
- The rbx variant alone is the best single member: 73.3% precision and 21.8% recall. If a candidate function's first decoded instruction is this, that's a solid (not conclusive) prologue hint.
- The family buys you recall (covers ~43% of MSVC functions) but at 13% precision it's far too noisy to fire on by itself — mov [rsp+imm], reg is the standard mid-function spill idiom, so 87% of its occurrences are false positives for FEP.
- Best practice for the MSVC family of heuristics: combine the two tiers —
- High-precision confirmations (use as strong/definitive signals): push rbx; sub rsp (100%), push rbp; sub rsp (100%).
- Moderate-precision recall boosters (add score, don't decide alone): the shadow-space mov [rsp+imm], reg family — weight the rbx variant higher (73%) than the rest (which dilute to ~13% at the family level).
So: keep 48895c24?? mov qword ptr [rsp + ??], rbx in the list as a recall-oriented booster, but gate any hard "this is a function start" decision on the high-precision push …; sub rsp / push rbp patterns. 
```

actually for first instruction, seems like at least for that MSVC version, the rex-prefixed pushes dominate on function entries, which is a really cool observation I've never noticed before:
```
byte    pattern	                    entry[0]	entry,!first	mid-function	start %
4053    push rbx	                11,481	    0	            0	            100.0%
4055    push rbp	                4,533	    0	            0	            100.0%
4053    push rbx ; 4883ec?? sub rsp	8,418	    0	            0	            100.0%
53      push rbx (plain)	        2	        2,041	        733	            0.1%
55      push rbp (plain)	        0	        6,089	        442         	0.0%
```

r0ny123:
Pushed the rework in ca3322b — think this covers everything.

Kept `endbr64` and `push r15; push r14` exactly as they were. Dropped the standalone `sub rsp, imm8` pattern entirely, you're right that it's too generic on its own. Added `40534883ec` and `40554883ec` as exact 5-byte patterns (`push rbx`/`push rbp; sub rsp, imm8`) — score 40, seeded as independent starts, matching your 100% precision numbers.

For `mov [rsp+disp8], rbx` I took your "weighted positive feature, not a hard trigger" framing pretty literally: it's an exact 4-byte pattern now too (no wildcard needed, same trick as the push;sub case), but I kept it scoring-only instead of seeding it. Checked the FP question you raised — ran it both ways against the Bao x64 MSVC corpus (68 binaries), and seeding it added 537 false positives corpus-wide for zero extra true positives recovered, so scoring-only was the clear call. Gave it a score of 30, below the push;sub tier.

Net effect on that corpus: precision 97.51% → 97.98% (FPs 2753 → 2216), recall unchanged at 99.6%. mirai_x64_xored and mirai_i386_xored both still land at exactly 151 functions.

Also ripped out the wildcard/`is_seedable` machinery like you suggested — `definitions.py` is just plain byte strings now, no masking logic left in `FunctionCandidate.py` or `FunctionCandidateManager.py`.

danielplohmann:
Thanks so much for the adjustments!

But this little expedition around x64 prologues raised the question for me if we should do another quick round of research and validation with modern MSVC (2022+) to see if there are more prologue constructs that are now common that were not present in the past. I'll consult some pre-built packages to gather data with symbols for that and do some evaluations, if I find anything noteworthy, I'll extend the list.

danielplohmann (approved):

r0ny123:
That's a good point; since most of these datasets were outdated, testing with the new versions makes sense.

r0ny123:
Hey Daniel — I actually ran exactly that round, figured I'd save you the trouble. Pulled a pile of MSVC 2022 x64 builds with matching PDBs and measured against the real ground truth.

One big lesson up front: don't trust the PDB `S_PUB32` "function" publics as the oracle. Half of them are padding/alignment/`int3`/`nop`/multi-byte-nop runs that the compiler tagged as functions. They're garbage for recall math. The actual clean oracle is the PE `.pdata` exception table — every x64 function MSVC emits gets a `RUNTIME_FUNCTION` entry, zero contamination. I used that.

Corpus: 55 real apps (obs-studio, WinMerge, mpc-hc, x64dbg, etc.), toolset 14.44. That's 108,420 genuine functions.

Headline: **your #177 list already nails it.** Against `.pdata`, SMDA recognizes the opener at 90.3% of all genuine function starts. Breakdown:
- 1-byte tier (`55`/`48`/`41`/`53`/`ff`/`40`/`e9`): 56.8%
- your scored `mov [rsp+imm8], rbx`: 33.5%
- unrecognized: only 9.7%

So there's no missing *high-frequency* construct from 2022. The 9.7% unrecognized is mostly three things:

1. `mov r11, rsp` (`4c 8b dc`) — real MSVC2022 opener, ~1.5% of starts, from the CFG/CFGuard + `__chkstk`/GS shadow preamble. It's genuine and rare mid-function, but it's NOT a stack-frame opener, so it'd be a low-confidence scored thing at best, never a hard seed. And at ~1.5% recall contribution it's a rounding error.

2. The rest of the `mov [rsp+imm], rX` shadow-space spill family (`48 89 4c 24 ??`, `44 89 4c 24 ??`, ...). This just confirms your own call from #177 — keep it scored, don't seed. Seeding the family is exactly what blew up FPs before (the 537-for-0 you measured). Same profile here, so leave it.

3. Thunks/trampolines — `mov edx, imm` jump stubs, `ret`-only fragments. Correctly NOT something we want as a prologue.

The long tail after that is hundreds of distinct first instructions each appearing 1–2 times. Idiosyncratic noise, nothing to chase.

Bottom line: I'd leave `DEFAULT_PROLOGUES_64` / `COMMON_PROLOGUES` exactly as #177 shipped them. If you really want the one defensible addition, a low-score (10–15) `4c 8b dc` scored-only entry might nudge CFG/GS-preamble recall by ~1.5%, but honestly it's marginal and I'd probably just not bother. Any real remaining recall gain on these binaries (SMDA sits around 79.6% overall recursive recovery vs `.pdata` — that's gap/indirect/exception-table coverage, not prologue scoring) comes from elsewhere, not new prologue bytes.

Full notes + repro scripts are in `.agents/msvc_prologue_research/FINDINGS.md` if you want to poke at the numbers yourself.

PR #179, opened 2026-07-12
feat(aarch64): resolve indirect calls across blocks and deepen jump-table recovery

r0ny123:
## Summary

The AArch64 backend recovers fewer CFG edges than Intel because its constant/PC-relative
tracking is (a) duplicated and (b) shallow: the same adrp/adr/add/mov/ldr constant-propagation
loop was independently re-implemented in `AArch64IndirectCallAnalyzer._resolveRegister` and
`AArch64JumpTableAnalyzer.getJumpTargets`, and neither could look past the single basic block
it started in.

This PR extracts one shared, maintained constant-propagation module
(`src/smda/aarch64/dataflow.py`) and uses it to close two capability gaps:

- **Indirect calls are no longer single-block.** `AArch64IndirectCallAnalyzer` now walks up to
  `block_depth` (default 3) predecessor blocks via `code_refs_to` to resolve a `blr`/`blra*`
  target register defined earlier in the function, capped per-block by
  `MAX_INDIRECT_CALLS_PER_BASIC_BLOCK` to bound cost on pathological cases (e.g. Go binaries with
  huge register-call fan-out in one block).
- **Jump tables recover more bases and sizes.**
  - the table base/index register can now resolve from a predecessor block instead of only the
    branch's own flat instruction window
  - a plain `ldr` followed by a standalone `sxtb`/`sxth`/`sxtw`/`uxtb`/`uxth` (instead of a single
    `ldrsw`) is now chased correctly — previously this broke the register-tracking chain entirely
  - table size is now also inferred from a `subs`-based bounds check, not just `cmp`
  - table reads are capped at the next known function candidate/border, in addition to the
    existing out-of-image/out-of-code checks

## Root cause / motivation

Two analyzers reimplementing the same forward dataflow pass meant fixes and improvements had to
be applied twice (and one copy, `AArch64JumpTableAnalyzer`, was missing both the `ldr`
memory-slot-dereference case and register invalidation entirely). Neither analyzer could see past
its starting block, so any indirect call or jump table whose setup instructions lived in a
predecessor block was silently unresolved.

## Changed behavior

- `src/smda/aarch64/dataflow.py` (new): `propagateConstants`, `propagateConstantsHistory` (a
  per-instruction snapshot history — needed because a jump-table base register is sometimes reused
  as the destination of the very load it addresses, so a single trailing constants map can no
  longer reflect the value that load actually read), and `gatherContextInstructions` (bounded
  predecessor-block context via `code_refs_to`).
- `src/smda/aarch64/analyzers.py`: both analyzers refit onto the shared module; see Summary above
  for the new capabilities.
- No public API or report-schema changes.

## Validation

- `ruff check .` / `ruff format --check .`: clean
- `pytest tests/ -q`: 422 passed, 1 skipped (5 new regression tests added, each with
  capstone-verified synthetic instruction encodings for: cross-block indirect-call resolution,
  the block_depth cap boundary, jump-table base recovery from a predecessor block, the
  ldr+standalone-extend chain, and subs-based size inference)
- Golden counts on the `aarch64_static` fixture (functions/blocks/instructions) are unchanged —
  verified with a before/after comparison against the pre-refactor code; this fixture doesn't
  happen to exercise the new cross-block/predecessor paths, so no rebaseline was needed
- Also verified against the 12-sample Mach-O AArch64 malware corpus already in `tests/`: no count
  change there either
- Additionally tested against a larger real-world ARM64 Mach-O sample outside the repo's test
  corpus (~12k functions): net +25 functions recovered (26 gained via the new predecessor-aware
  paths, 1 lost — very likely a stale-register false positive from the old code's lack of
  register invalidation, now correctly left unresolved)

## Notes

- Builds on top of #178 (not yet merged) — this branch is stacked on
  `feat/aarch64-phase1-accuracy`, so the diff shown here includes those commits too; only the last
  two commits (`feat(aarch64): resolve indirect calls...` and
  `test(aarch64): cover cross-block...`) are new in this PR.
- Two related raw-word (non-capstone-operand) constant trackers,
  `AArch64Backend._resolvePltGotSlot` and `FunctionCandidateManager.locateAddressRefCandidates`,
  intentionally still carry their own independent implementation and are not touched here — that's
  a separate, larger follow-up (bridging the raw-word and capstone-operand representations).
- PAC branch-target operand correctness (`bra*`/`blra*`) and additional real-corpus sourcing for
  the new code paths were both considered and intentionally left out of scope: synthetic coverage
  plus the real-sample validation above already demonstrate correctness, and PAC operand handling
  showed no measurable recovery gain worth the added surface here.

### DIFF
--- a/.github/workflows/scripts/run_perf_check.py
+++ b/.github/workflows/scripts/run_perf_check.py
@@ -45,6 +45,7 @@ def run_benchmark(iterations, output_file):
         {"name": "komplex", "filename": "komplex_xored", "base_addr": 0, "backend": "intel"},
         {"name": "njrat", "filename": "njrat_xored", "base_addr": 0, "backend": "cil"},
         {"name": "pe_export", "filename": "pe_export_label_test_xored", "base_addr": 0, "backend": "intel"},
+        {"name": "aarch64_static", "filename": "aarch64_static_xored", "base_addr": 0, "backend": "aarch64"},
     ]
 
     results = {}
--- a/src/smda/Disassembler.py
+++ b/src/smda/Disassembler.py
@@ -87,21 +87,36 @@ class Disassembler:
         finally:
             smda_report.buffer = previous_buffer
 
+    #: backends that record non-buffer-backed strings via disassembly.addStringRef
+    #: (the string only exists as a runtime/decode-time side effect, e.g. CIL's
+    #: ldstr token or an AArch64 stack-built string — there are no corresponding
+    #: bytes anywhere in the file for the generic StringExtractor to find).
+    _ANALYZER_STRINGREF_TYPES = {"dalvik": "dex", "cil": "cil", "aarch64": "stack"}
+
     def _extractStrings(self, smda_report, mode=None):
+        architecture = smda_report.architecture
+        analyzer_type = self._ANALYZER_STRINGREF_TYPES.get(architecture, "analyzer")
         for smda_function in smda_report.getFunctions():
-            if smda_report.architecture == "dalvik":
-                if smda_function.stringrefs and isinstance(smda_function.stringrefs, dict):
-                    smda_function.stringrefs = [
-                        {
-                            "string": string_value,
-                            "ins_addr": referencing_addr,
-                            "data_addr": None,
-                            "type": "dex",
-                        }
-                        for referencing_addr, string_value in sorted(smda_function.stringrefs.items())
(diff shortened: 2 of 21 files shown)


danielplohmann (approved):

r0ny123:
@danielplohmann let me fix the tests.

PR #186, opened 2026-07-16
fix(aarch64): improve function-boundary accuracy on ARM64 PE binaries

r0ny123:
## Summary

Fixes three pre-existing function-boundary bug classes in the AArch64 backend (plus one sibling in the intel gap scan), found while validating SMDA against IDA ground truth on real ARM64 PE binaries (wermgr.exe, ping.exe, robocopy.exe, bcrypt.dll — Microsoft-signed system binaries). All three classes reproduce independently of the `.pdata` feature from #185 (they occur with `USE_PE_ARM64_PDATA_CANDIDATES` off), so they are fixed here on their own branch.

## Bug classes and fixes (one commit each)

**1. Data misread as code — trap-opening gap candidates** (`449acbe`, `c1058f7`)
`nextGapCandidate()` rejected gap candidates opening with a conditional branch but not ones opening with a trap encoding. On ARM64, `udf #imm16` encodes as `0x0000_iiii`, so exception-data words in executable sections were promoted to function starts. Fix: raw-word trap predicates (`udf`/`brk`/`hlt`) in `definitions.py`, used to skip trap-opening gap candidates and to count trap words as valid boundaries for the interior-BTI check. The intel gap scan had the same blind spot for `hlt`/`ud2` trap filler (`c1058f7`).

**2. Mid-function splits after alignment NOPs** (`108c669`)
`_callFallthroughFunctionStart()` promoted any 16-aligned non-padding word after a `bl` + NOP run to a function start. MSVC ARM64 nop-aligns loop heads *mid-function*, so hot-loop heads were split into bogus functions. Fix: on PE images, the aligned word must also look like a function entry (prologue or BTI landing pad). Deliberately scoped to PE: clang/Go (Mach-O/ELF) pad between functions and emit prologue-less entries, so other formats keep the plain alignment cut — an unscoped gate cost ~150 real function starts on the Mach-O corpus.

**3. Guard stubs absorbing `.pdata`-named functions across conditional branches** (`c5b3f0f`)
`_analyzeCondBranch()` absorbed the taken target of a `b.<cond>` unless the target was already a recovered function — unlike `_analyzeUncondBranch()`, it never consulted candidate state. A small guard stub ending in `b.ne <real_function>` swallowed the entire following function even when `.pdata` explicitly named it. Fix: treat an exception-record-seeded candidate as an authoritative boundary (conditional tailcall edge) — but only when its first word is entry-shaped, because `.pdata` also names funclets/fragments whose record start is a mid-body word and those must stay absorbed (an unconditional guard added +34 false starts on bcrypt.dll).

## Results (function starts vs IDA, `.pdata` candidates enabled)

| sample | FP before → after | FN before → after |
|---|---|---|
| wermgr.exe | 87 → 71 | 22 → 21 |
| robocopy.exe | 118 → 78 | 19 → 18 |
| ping.exe | 11 → 10 | 5 → 4 |
| bcrypt.dll | 74 → 74 | 10 → 9 |

With `.pdata` candidates off, wermgr.exe improves 118 → 102 FP. Exact per-fixture function-*set* diffs (not counts) on the 12-sample Mach-O corpus and the static ELF fixture are byte-for-byte neutral for fixes 2–3; fix 1 removes only byte-verified data starts.

## Tests

Regression tests for each class (each verified to fail on pre-fix code): trap-opening gap rejection, PE call-fallthrough alignment gating (plus a non-PE control asserting the alignment cut still applies), and cond-branch exception-boundary behavior including a funclet-continuation control. `make test`: 525 passed, 1 skipped; ruff clean.

### DIFF
--- a/src/smda/aarch64/AArch64Backend.py
+++ b/src/smda/aarch64/AArch64Backend.py
@@ -142,8 +142,18 @@ class AArch64Backend(ArchBackend):
             return cursor
         if skipped_nop and cursor % 16 == 0:
             word = cls._wordAt(d, cursor)
-            if word not in (None, 0, NOP):
-                return cursor
+            if word in (None, 0, NOP):
+                return None
+            # MSVC ARM64 nop-aligns loop heads MID-function, so on PE images an
+            # aligned word after call-fallthrough padding is only a boundary when
+            # it actually looks like a function entry (real starts are seeded by
+            # .pdata/candidates anyway). clang/Go pad between functions and emit
+            # prologue-less entries, so other formats keep the plain alignment cut.
+            if d.disassembly.binary_info._getLiefType() == "PE" and not (
+                is_function_prologue(word) or is_bti_landing_pad(word)
+            ):
+                return None
+            return cursor
         return None
 
     def _analyzeCondBranch(self, d, instruction, state):
--- a/src/smda/aarch64/AArch64Backend.py
+++ b/src/smda/aarch64/AArch64Backend.py
@@ -164,8 +164,11 @@ class AArch64Backend(ArchBackend):
         target = self._branchTarget(i_op_str)
         d.tailcall_analyzer.addJump(i_address, target if target is not None else 0)
         if target is not None:
-            if target in d.disassembly.functions:
-                # taken edge into an already-recovered function: a conditional tailcall edge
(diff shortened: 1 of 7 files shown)


danielplohmann:
very nice, thx! also cool, that this led to the mirror fix in intel.

PR #195, opened 2026-07-18
fix(intel): widen PIC-hash escaping to cover 64-bit immediates

r0ny123:
## Summary

Widens Intel PIC-hash escaping to correctly handle 64-bit immediates, and adds a version-gated pic_hash recalculation path for old reports.

## Motivation

`IntelInstructionEscaper.escapeBinary`'s immediate regex was capped at 8 hex digits (32 bits): `r"0x[0-9a-fA-F]{1,8}"`. Because the quantifier is greedy but unanchored, a genuine 64-bit immediate operand (e.g. `movabs rax, 0x1122334455667788`) did not fail to match -- it silently matched only the first 8 hex digits, producing a wrong, truncated value that essentially never falls inside the checked `[lower_addr, upper_addr)` image-address range. In-image `mov r64, imm64` constants were therefore never recognized as relocatable addresses and never escaped, so `pic_hash` was not relocation-invariant on 64-bit binaries.

## Changed behavior

- Widened the regex to 16 hex digits, so a full 64-bit immediate is extracted correctly.
- `escapeBinaryValue`'s packing now tries a 4-byte little-endian pack and falls back to 8-byte on `struct.error` (mirroring `escapeBinaryPtrRef`'s existing pattern directly above it in the same file), with a dynamically-sized wildcard string instead of a hardcoded 8-character one. This also fixes a related nit: the previous 4-byte pack used native-endian `"I"` rather than explicit `"<I"`.
- Added `INTEL_PIC_HASH_ESCAPE_VERSION` (mirroring the existing `AARCH64_PIC_HASH_ESCAPE_VERSION` mechanism in `SmdaFunction.fromDict`), so reports imported from older SMDA versions get `pic_hash` recalculated for intel functions instead of trusting a stale hash computed with the old, narrower escaping. Bumped `SmdaConfig.VERSION` / `__version__` to 4.3.1 to back this gate (rebased past the maintainer's independent 4.3.0 bump that landed on master concurrently).

## Accuracy / hash-migration evidence

Ran a before/after comparison on Bao byteweight msvc10-64 (15 binaries, same set both runs):

- Predicted function address sets are byte-for-byte identical before and after, for every binary -- confirms the escaper change is fully isolated from candidate/CFG recovery, as expected (the escaper only feeds hash computation).
- The new 8-byte fallback pack path was not exercised anywhere in this 15-binary sample (0 of ~20k+ predicted functions) -- the measured real-world hash-migration footprint on this corpus is effectively zero. The fix remains necessary for the narrower case of large in-image 64-bit absolute immediates (e.g. `movabs`-heavy or PIC-disabled code), which is exercised directly by the new unit test.

## Validation

- `make lint`
- `python -m ruff format --check .`
- `make test` -- 531 passed, 1 skipped
- New regression tests: `tests/testEscaper.py` (64-bit `movabs` immediate escaping, fails against the pre-fix 8-digit regex) and `tests/testCommonModels.py` (pic_hash recalculation before/at `INTEL_PIC_HASH_ESCAPE_VERSION`, mirroring the existing AArch64 test)

No report schema, serialized field structure, or status value changes are introduced. `pic_hash` values change only for the narrow set of intel functions containing an in-image 64-bit immediate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ make test
 ´
 ## Version History
  * 2026-07-17: v4.3.0 - Format-aware `xheader` capture: `getHeaderBytes()` now stores computed, trailing-zero-trimmed, capped header regions for PE (section table), ELF (program headers), and Mach-O (active-slice load commands) instead of fixed truncations, enabling metadata recovery and binary re-synthesis. Added a normalized PE header hash (`SmdaReport.pe_header_hash`, volatile TimeDateStamp/CheckSum/SizeOfImage zeroed) for hash-busting-resistant clustering.
+ * 2026-07-18: Widened Intel PIC-hash escaping to cover 64-bit immediates (`mov r64, imm64` constants were previously truncated to their first 8 hex digits and never escaped, so PicHash was not relocation-invariant on 64-bit binaries). (THX: @r0ny123)
  * 2026-07-17: v4.2.17 - Improved function-boundary accuracy on ARM64 PE binaries (trap-data gap rejection, prologue-gated call-fallthrough alignment cuts, .pdata-authoritative conditional-tailcall boundaries) with the matching x86 gap-scan/alignment-cut fixes. (THX: @r0ny123)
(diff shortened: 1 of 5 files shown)


danielplohmann:
integrated via CLI

PR #198, opened 2026-07-21
feat(core): experimental binary synthesis (PE/ELF/Mach-O)

danielplohmann:
## Summary

Adds experimental binary synthesis to SMDA — `SmdaReport.synthesizeBinary()` rebuilds fictive PE, ELF, and Mach-O files from a recovered CFG, planting bytes per basic block at original VAs and fusing import metadata. New `smda/synthesis/` package with `BinarySynthesizer` base class and `PeSynthesizer`, `ElfSynthesizer`, `MachoSynthesizer`.

## Validation

- 12 new tests in `tests/testSynthesis.py` covering format sniffing, headered/minimal synthesis, import table wiring, planted-block verification (re-parsed via LIEF), and determinism — all passing inside the devcontainer (Python 3.10).
- Format and lint both clean (`ruff check` all checks passed, 156 files already formatted).

## Scope Notes

- **Experimental:** works on well-formed reports; not hardened against pathological/adversarial inputs. Golden fixtures cover one sample per format.
- **No new runtime deps:** leverages LIEF (already a dependency).
- **Deterministic from report content only** — does not consult `STORE_BUFFER`.
- `code_sections` names are **not preserved** through `toDict()`/`fromDict()` round-trips (they come from LIEF metadata).


src/smda/synthesis/MachoSynthesizer.py
```diff
@@ -0,0 +1,713 @@
...
+        return segments, section_metas
+
+    def _applyXheaderLayout(self, sections, layout):
+        """Transfers original segment membership and section flags onto report sections.
+        Section VAs may be shifted (fat-slice/base adjustment); a constant delta is
+        derived from matching addresses."""
+        segments, section_metas = layout
+        rep_addrs = sorted(section["va_start"] for section in sections)
+        xh_addrs = sorted(section_metas.keys())
+        adjustment = rep_addrs[0] - xh_addrs[0]
```
r0ny123, line 206: `_parseXheaderLayout` (a few lines up) returns `(segments, section_metas)` as soon as *any* `LC_SEGMENT(_64)` command is found — even if every one of those segments has `nsects == 0` (e.g. a truncated/minimal header where the only surviving segment commands are `__PAGEZERO`/`__LINKEDIT`-style entries with no sections). In that case `section_metas` is `{}`, so `xh_addrs` here is `[]` and `xh_addrs[0]` raises `IndexError`.

That propagates straight out of `synthesize()` — there's no try/except around this header-driven path the way `PeSynthesizer.synthesize()` wraps `_synthesizeFromHeader` and falls back to `_synthesizeMinimal` on any non-operational exception. Given the docstring above already anticipates "an xheader whose trailing zeros were trimmed," this is exactly the class of malformed header the feature should degrade gracefully on rather than crash on.

Suggest either requiring `section_metas` to be non-empty before `_parseXheaderLayout` returns a usable layout (falling through to the heuristic classifier), or catching this in `synthesize()` the same way `PeSynthesizer` does.

src/smda/synthesis/PeSynthesizer.py
```diff
@@ -0,0 +1,443 @@
...
+            for region in regions:
+                if region["vaddr"] <= rva and rva + ptr_size <= region["vaddr"] + len(region["raw"]):
+                    struct.pack_into("<I" if ptr_size == 4 else "<Q", region["raw"], rva - region["vaddr"], thunk_value)
+                    break
+            else:
+                self._warn("import thunk 0x%x (%s.%s) outside all sections", rva, dll, func)
+
+    def _patchDescriptors(self, raw, dll_funcs, sec_vaddr, idesc_off, dll_off, dll_str_offsets, hint_off):
+        for index, (dll_name, funcs) in enumerate(dll_funcs):
+            first_iat_rva = funcs[0][0]
```
r0ny123, line 145: `first_iat_rva = funcs[0][0]` becomes this DLL's `FirstThunk`/`OriginalFirstThunk`, and `_writeThunks` (above) only writes a non-zero value at the *known* slot RVAs — everything else in the backing region stays zero (from `bytearray(...)`). PE loaders/parsers, LIEF included, walk the thunk array starting at `FirstThunk` until they hit a zero entry and treat that as the array terminator; they have no independent count.

So if a DLL's slots in `xmetadata["imported_functions"]` aren't contiguous (a gap between two slots for the same DLL), the zero in the gap reads as "end of array" and every import past the gap for that DLL silently vanishes on reparse — no exception, no warning, just a wrong result.

The current fixtures don't hit this because `cutwail`'s real IAT is naturally contiguous per DLL, but `testElfSynthesisImportTables` already shows the pattern that would (mutating import metadata into a layout the original binary didn't have). Worth either warning + skipping non-contiguous groups the way the block-planting path already does for out-of-section blocks, or padding gaps with synthetic ordinal-0 thunks so the walk doesn't terminate early.

src/smda/synthesis/BinarySynthesizer.py
```diff
@@ -0,0 +1,100 @@
...
+    def _resolveFunctionOffsets(self, function_offsets):
+        offsets = sorted(self.report.xcfg.keys()) if function_offsets is None else list(function_offsets)
+        resolved = sorted(offset for offset in offsets if offset in self.report.xcfg)
+        skipped = len(offsets) - len(resolved)
+        if skipped:
+            self._warn("synthesis: %d requested function offsets are not in the report", skipped)
+        return resolved
+
+    @staticmethod
+    def _collectFunctionBytes(smda_function):
```
r0ny123, line 47: `_collectFunctionBytes` doesn't appear to be called anywhere — `Elf`/`Macho`/`PeSynthesizer` all use `_iterFunctionChunks` (right below) directly for per-block planting instead. Worth deleting if it's dead, or wiring it in if it was meant to be the non-block-aware entry point somewhere.

src/smda/synthesis/BinarySynthesizer.py
```diff
@@ -0,0 +1,100 @@
...
+    def _iterStringRefs(self):
+        for smda_function in self.report.getFunctions():
+            for stringref in smda_function.stringrefs or []:
+                data_addr = stringref.get("data_addr")
+                string = stringref.get("string")
+                if data_addr is None or not string:
+                    continue
+                yield data_addr, string.encode("ascii", errors="ignore") + b"\x00"
+
+    def _getPointerSize(self):
```
r0ny123, line 88: `_getPointerSize` also looks unused — each subclass computes its own bitness/pointer size instead (`ElfSynthesizer._getBitness`, `PeSynthesizer._getBitness`, `MachoSynthesizer._is64`). Consolidating onto this (or dropping it) would remove one more source of "which of these three near-identical helpers is actually in effect" confusion.

src/smda/synthesis/BinarySynthesizer.py
```diff
@@ -0,0 +1,100 @@
...
+        return 8 if self.report.bitness == 64 else 4
+
+    def _hasHeader(self, min_length=0x40):
+        return bool(self.report.xheader) and len(self.report.xheader) >= min_length
+
+    def _plantInto(self, regions, addr, content):
```
r0ny123, line 94: `_plantInto` is unused too — `Elf`/`Macho`/`PeSynthesizer` each hand-roll the same "find the region containing this VA, splice bytes in, else warn" loop instead (e.g. `ElfSynthesizer._prepareSections`'s block-planting loop, `PeSynthesizer._plantFunctions`/`_plantStrings`, the equivalent in `MachoSynthesizer._prepareSections`). That's ~30 duplicated lines across three files that this helper could absorb.

src/smda/synthesis/ElfSynthesizer.py
```diff
@@ -0,0 +1,535 @@
...
+    """Rebuilds an ELF file from an SmdaReport.
+
+    Section layout comes from the report's code_sections (name/VA ranges);
+    program headers are rebuilt around them. Imports are fused into synthetic
+    .dynsym/.dynstr/.rela(dyn)/.dynamic tables so analysis tools resolve named
+    imports at their original relocation slot addresses. Reports without
+    sections (headerless buffers) get a minimal single-segment ELF.
+    """
+
+    def synthesize(self, function_offsets=None, with_imports=True, with_strings=True) -> bytes:
```
r0ny123, line 75: For consistency: `PeSynthesizer.synthesize()` wraps its header-driven path in try/except and falls back to a minimal layout on any non-operational exception; this and `MachoSynthesizer.synthesize()` call their section/header-driven path with no equivalent fallback, so any exception raised deep in `_prepareSections`/`_buildImportTables`/`_assembleFile` propagates straight out. See the `MachoSynthesizer._applyXheaderLayout` comment for a concrete case that hits this today. Given the "experimental, not hardened" framing, applying the same defensive pattern here would make the three synthesizers behave consistently on malformed input.

src/smda/synthesis/PeSynthesizer.py
```diff
@@ -0,0 +1,443 @@
...
+
+        e_lfanew = struct.unpack("<I", self.report.xheader[0x3C:0x40])[0]
+        size_opt = struct.unpack("<H", self.report.xheader[e_lfanew + 20 : e_lfanew + 22])[0]
+        optional_header = bytearray(self.report.xheader[e_lfanew + 24 : e_lfanew + 24 + size_opt])
+        if len(optional_header) < size_opt:
+            optional_header += b"\x00" * (size_opt - len(optional_header))
+
+        size_of_code = sum(len(r["raw"]) for r in regions if r["executable"])
+        size_init_data = sum(len(r["raw"]) for r in regions if int(r["chars"]) & IMAGE_SCN_CNT_INITIALIZED_DATA)
+        struct.pack_into("<I", optional_header, 4, size_of_code)
```
r0ny123, line 273: Minor maintainability note: this whole block (`SizeOfCode`/`SizeOfInitializedData` at 4/8, `SizeOfImage` at 56, `SizeOfHeaders` at 60, `dir_base` a few lines down) writes into `IMAGE_OPTIONAL_HEADER` by bare integer offset with no named constants, unlike the section-characteristics/import-directory constants defined at module scope above. A wrong offset here produces a binary that's subtly malformed rather than an error — nothing in lint/type-checking would catch a transcription mistake. A handful of named offsets (`OPT_HDR_SIZE_OF_CODE = 4`, etc.) mirroring the existing `IMAGE_SCN_*`/`IMAGE_DIRECTORY_ENTRY_IMPORT` style would make this safer to review and modify later.

tests/testSynthesis.py
```diff
@@ -0,0 +1,212 @@
...
+        for index, first in enumerate(loads):
+            for second in loads[index + 1 :]:
+                assert not (
+                    first.virtual_address < second.virtual_address + second.physical_size
+                    and second.virtual_address < first.virtual_address + first.physical_size
+                )
+        sections = [(s.virtual_address, bytes(s.content)) for s in parsed.sections if s.name]
+        assert _verify_planted_blocks(report, sections) == []
+
+    def testElfSynthesisImportTables(self):
```
r0ny123, line 143: This test (mutating `xmetadata["imported_functions"]` into a layout the original binary didn't have) is exactly the pattern that would catch the PE non-contiguous-thunk issue flagged in `PeSynthesizer._patchDescriptors` — a PE equivalent of this test (two imports for the same DLL with a gap between their slot addresses, then asserting both names survive reparse) would be a good addition. Same idea for a Mach-O header with an `nsects == 0` segment to catch the `_applyXheaderLayout` crash noted there.

r0ny123 (commented):
Nice piece of work overall — reusing LIEF for header re-parsing and being disciplined about determinism (sorting after every set-based dedup, e.g. the `names`/`libs` lists in each `_buildImportTables`) means this reads a lot more carefully than "hand-rolled struct packing" usually does.

Left inline notes below. Two of them I'd treat as real bugs rather than nits, precisely because of the "not hardened against pathological/adversarial inputs" scope note in the description — both fail *silently* (wrong output or an uncaught crash) rather than loudly:

- `MachoSynthesizer._applyXheaderLayout` can raise an unguarded `IndexError` on a header whose segments all have zero sections, and there's no fallback-to-minimal around it the way `PeSynthesizer.synthesize()` has for its header path.
- `PeSynthesizer._writeThunks`/`_patchDescriptors` assume each DLL's import slots are contiguous in the thunk array; if they aren't, imports past a gap silently disappear on reparse (LIEF/loaders terminate the thunk walk on the first zero entry).

Also flagged: three `BinarySynthesizer` helpers (`_collectFunctionBytes`, `_getPointerSize`, `_plantInto`) that look like they were meant to be the shared implementation but aren't actually called — each subclass reimplements the same logic inline instead. And a couple of spots where named offset constants would make the header-field writes safer to maintain (a wrong offset here fails silently, not via lint/type-check).

None of this pushes back on the overall design — the per-block planting + import-fusion approach is sound, these all look fixable within the current structure. Happy to help narrow down repro cases for either bug if useful.

danielplohmann:
Thanks for the thorough review, @r0ny123! All comments are addressed and the PR is being integrated.

Recap of what landed on top of the original feature:
- **Dead code:** removed the three unused base-class helpers (`_collectFunctionBytes`, `_getPointerSize`, `_plantInto`); consolidating the duplicated per-block planting loops is deferred to a follow-up cleanup.
- **Defensive fallbacks:** `ElfSynthesizer` and `MachoSynthesizer` now mirror `PeSynthesizer`'s `try/except → minimal layout` pattern.
- **Mach-O `nsects == 0`:** `_parseXheaderLayout` returns `None` when there are no section metas, so a truncated/minimal xheader degrades gracefully instead of crashing in `_applyXheaderLayout` (`IndexError`).
- **Non-contiguous PE IAT:** gaps within a DLL's slot range are padded with ordinal-0 thunks and an explicit zero terminator is written, so all imports survive reparse even for format-violating (no-terminator) custom IATs.
- **PE optional header:** bare integer offsets replaced with named `OPT_HDR_*` / `DATA_DIR_BASE_*` constants.
- **Tests:** added `testPeSynthesisNonContiguousImports` and `testMachoSynthesisNsectsZero` to lock in the above.

Two integration notes:
- Renumbered to **4.4.1** because `master` had already shipped 4.4.0 (Labels/reporting).
- Dropped a duplicated README section the original PR accidentally copied (the “Further explanation…” / install-list boilerplate that already exists further down) — squashed into the bump commit.

Deliberately left out of scope: runtime-resolved / custom IATs and dynamic API refs not recorded among the native imported symbols. Tracked as future work in #214 so it doesn't block this PR.

Merging now — thanks again for the careful pass!

PR #226, opened 2026-08-05
ci: make the fuzzing, coverage and perf gates catch things they were missing

r0ny123:
This started as a look at where SMDA's tooling was reporting success without actually checking anything. Tightening the gates then surfaced a run of real bugs, several of them on the report-import path that MCRIT uses.

13 commits, one per area, each self-contained — no file appears in two of them, so it should read commit by commit if that's easier than the combined diff. Happy to split it up if you'd rather take it in pieces; a few parts are opinionated and I've flagged those at the bottom.

## Bugs fixed

**`lief.bad_file` doesn't exist any more.** `ElfFileLoader.getAbi` caught it, and it's gone in lief 0.17 (`hasattr` says False) — so the moment that handler was needed it would itself have raised `AttributeError`. It also read `.name` off `identity_os_abi`, but lief returns a plain int for an ABI it has no enum member for, and `.name` on an int is another `AttributeError`.

**Report deserialization accepted almost anything.** `fromDict` read fields unconditionally, so a truncated report surfaced as a `KeyError` from mid-rebuild rather than the `ValueError` the surrounding validation uses, and nothing constrained the values that did arrive. Four separate ways through, all now closed at one boundary: required fields present, containers actually containers, addresses and sizes integers rather than any number (`struct.pack_into` rejects a float outright), and every address inside the 64-bit space the synthesizers pack it into.

The subtle one was the **xcfg key** — it becomes the function offset via `int(function_addr)` with no sign check, so a negative key reached an ELF program header as `p_vaddr = -14200832`. It slipped past the image-span bound too, because that measures `max − min` and a negative floor under a positive ceiling still measured a legal 17 MB.

**A cached nesting depth was recomputed for every function:**

```python
metadata.get("nesting_depth", smda_function._calculateNestingDepth())
```

Python evaluates a `.get()` default whether or not the key is there, so every function on the cached path built a full dominator tree and discarded it. Verified by counting calls: 1 with the field present, should be 0.

**Synthesis could build an inverted segment.** The synthetic section for functions outside every known section took its floor from function offsets and its ceiling from extent ends — independent quantities at different granularities. Real binaries always have an extent end above the offset it started from; a crafted report doesn't, and `struct.pack` got `-4080` unsigned. Same shape at four sites, only one of which had crashed.

**Synthesis allocated without bound.** `MAX_IMAGE_SIZE` has been 100 MB all along and every loader enforces it — the synthesizers, which *build* images, never got the same guard, so functions 1 GB apart produced a 1 GB bytearray.

**`escapeBinaryPtrRef` couldn't survive a wide displacement.** Reached from `SmdaReport.fromDict` via `getPicHash`, so this one is on the plain report-import path. The operand regex takes a hex literal of any length; both escape helpers pack `<I`, fall back to `<Q`, and had nothing behind the fallback.

**`SmdaConfig` leaked state between instances** — class-attribute based, so `config.API_COLLECTION_FILES[name] = path` wrote through to the class and showed up in every other config in the process. That's exactly SMDA's batch/service shape.

**Windows encoding.** A previous commit switched six writes in `evaluate_runtime.py` to explicit UTF-8 and left every matching read alone, including the cache file it had just started writing as UTF-8. Swept the rest. The reads matter more than the writes: `json.dump` defaults to `ensure_ascii=True`, so SMDA's own reports are pure ASCII — a report written by another tool, or the markdown these scripts emit, is not.

**`make clean` never deleted anything** — unclosed group in its regex, so grep errored out every time.

## Added

**A synthesis fuzz target, a stricter oracle, and a committed corpus floor.** The old oracle promoted only `RecursionError` on top of the library's operational/non-operational split, so `TypeError`, `AttributeError`, `KeyError`, `IndexError` and `struct.error` were swallowed — precisely how most real Python defects present. The stricter oracle lives in `fuzzing/`, not `src/`, so the library contract is untouched.

`src/smda/synthesis/` is ~1.9k lines writing binary headers from report-derived integers, documented as unhardened, with 14 tests and no fuzzing. The new target asserts its output re-parses with LIEF — a synthesizer whose output no longer loads has failed at the one thing it's for. It found every synthesis and escaper bug above.

The accumulating corpus lived only in the Actions cache, which evicts after 7 days without a hit, so a quiet fortnight silently reset coverage to bare seeds. `tests/fuzz_corpus/` now holds a 212 KiB pruned floor. And `tests/fuzz_regressions/` had a documented convention and zero entries — it now holds the reproducers, each verified to *bisect* its fix rather than merely pass.

**Coverage measurement and gates.** Branch coverage and an exclude list existed with no `fail_under`, no `--cov` anywhere in `.github/`, and `ignore_errors` swallowing measurement errors — nobody knew the number. Now collected on one leg, published to the step summary, floored at 75 (measured 81), and gated on *diff* coverage, since a global floor on 24.7k lines barely moves when a new module lands untested.

**The suite now runs on Windows and macOS.** Previously Linux-only across four Python versions, with the other platforms appearing in a job that imports `smda` and prints versions. Both passed first time — which the UTF-8 sweep is what made true.

**Supply-chain scanning.** `pip-audit` against the dependency set resolved from `pyproject.toml`, plus `zizmor` over the workflows, on a weekly cron. It immediately earned its place: the setuptools bound in the `Makefile` and `README` pinned contributors *below* the fix for PYSEC-2026-3447. Also pinned the one floating action, which ran under `pull_request_target`. And `SECURITY.md`, which didn't exist for a library whose whole job is parsing hostile input.

**A peak-memory channel in the perf gate**, via `tracemalloc` in its own pass so tracing overhead can't skew the timing channel. Gated at 10%. The time threshold drops 50% → 30%, but a fixture must now also lose 0.15s absolute before failing — at 30% alone, 68 ms of variance on a 0.2 s fixture failed the build, and the new memory channel is what proved it was noise (10.3 MiB both sides, +0.01%).

**`py.typed`**, so an installed `smda` stops handing consumers zero types. `ty` also now covers `fuzzing/`, `profiling/` and `.github/workflows/scripts/` — the last holds `evaluate_runtime.py`, 955 lines of statistics gating the perf benchmark that nothing but its own tests had checked.

**Test ergonomics** — `filterwarnings = ["error"]`, a Hypothesis profile with `deadline=None`, and a `slow` marker turning a ~150 s suite into a ~15 s subset for iteration.

## Removed

**`version_history.md`**, folded into a new `CHANGELOG.md`. The two had genuinely forked — 71 versions only in README's list, 94 only in `version_history.md`, 15 in both. README keeps a pointer and drops from 289 to 201 lines.

**`requirements.txt`** — consumed by nothing, and missing `hypothesis`, `ty` and `diff-cover`.

**Eight methods in `intel/FunctionCandidateManager.py`** re-declaring their own parent's body character for character. Compared parsed bodies rather than eyeballing: six other overrides in the same class genuinely differ and are untouched. Report identity hashes and function/block/instruction counts across six fixtures are unchanged.

**`SmdaConfig.LOG_PATH`** — defined, never read. Public attribute though, so shout if you'd rather keep it.

## Validation

| Check | Result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | clean, 217 files |
| `python -m ty check src/smda/ fuzzing/ profiling/ .github/workflows/scripts/` | clean |
| `make test` | 929 passed, 1 skipped, 793 subtests |
| coverage | 81% (gate 75) |
| `diff-cover --fail-under=100` | 100%, 0 missing |
| local fuzzing | 21.5M executions across `synthesis` + `report`, 0 findings |

Report identity hashes on six fixtures are unchanged across the whole branch — none of this was supposed to move recovery output, and it doesn't.

## Things you might want to push back on

- Deleting `SmdaConfig.LOG_PATH` is a public-attribute removal, even though nothing reads it.
- Deserialization errors now come out as `ValueError` where some were `KeyError` or `TypeError`. Anything catching `KeyError` around `fromDict` would stop catching.
- Folding the README's version history into `CHANGELOG.md` is a docs-layout opinion more than a fix — easy to drop that commit.
- `diff-cover --fail-under=100` is strict. It only measures `src/smda/**`, so changes under `tests/`, `fuzzing/` or the CI scripts pass by not being measured; there's a note in the workflow saying so, since it reads stricter than it is.

## Not done

- `py.typed` ships, but only `Disassembler`'s public API is annotated — the model classes are still mostly inferred (~71 of 75 public methods). Deliberately not guessing at 71 signatures under `all = "error"`.
- `ty` doesn't cover `tests/` yet; ~187 diagnostics there need their own pass.
- No `pytest-xdist`. `SmdaConfig`'s mutable-default leak is fixed, but `lief.logging.disable()` still runs as a constructor side effect and several `lru_cache`s key on binary content. Separately, the Mach-O synthesis tests are intermittently flaky under load — they pass in isolation, and a baseline with this branch's changes stashed shows the same behaviour, so it predates this work.


danielplohmann:
Thanks, this is good work, and the report-import fixes are the kind I want caught before MCRIT sees them.
I verified the report identity hashes myself across five fixtures (intel PE+ELF, Mach-O, AArch64, Dalvik) and they are unchanged, and the eight removed FunctionCandidateManager overrides really are AST-identical to the parent.
Merging once the points below are addressed.

**1. Please reword commit 96e32e3's nesting-depth paragraph.**

The commit message says `fromDict` passed `_calculateNestingDepth()` as a `dict.get()` default, so every function on the cached path built and discarded a dominator tree.
The line the commit actually replaces is:

    smda_function.nesting_depth = function_dict["metadata"]["nesting_depth"]

a plain subscript, so zero wasted calls.
`git log -S` finds no `.get()` form anywhere in that file's history, so I think the measurement came from a different local state.

The change itself is worth keeping, just for a different reason: a report whose metadata lacks `nesting_depth` now recomputes instead of raising `KeyError`, which covers reports written between 2020-04-30 and 2020-06-22 (nesting_depth landed in aeedbaf).
Please restate it that way and drop the performance claim.
This one goes into the history permanently, unlike the PR description.

Same for the PR body while you are at it: the committed corpus floor is 39 KiB across 54 files, not 212 KiB.

**2. Delete `.devcontainer/` in this PR.**

You removed every doc pointer to it, correctly, but the directory is still pinned at `ARG PYTHON_VERSION=3.10` and still calls itself "SMDA (Python 3.10)", which has been unusable since 4.4.0 required 3.11+.
It dates from when I was on an older Ubuntu.
It has no reason to exist now, and removing the references while leaving the broken container is the one state worse than either.

**3. `pip-audit` should not block on push and PR.**

It earned its place with the setuptools bound.
As wired, though, it turns unrelated PRs red whenever a CVE is published against a transitive dep, with nothing in the diff to fix.
Please `continue-on-error: true` on the push/PR triggers and keep it hard-failing on the weekly cron.

**4. Same for the diff-coverage gate, for now.**

`--fail-under=100` on `src/smda/**` is a real policy change and I would rather see the number before committing to it.
`continue-on-error: true` on that step keeps it in the step summary on every PR without holding merges up.
If it sits near 100 on its own over the next few PRs, we make it blocking then.

**5. Keep `filterwarnings = ["error"]`, but get the suppression out of `src/`.**

Warnings-as-errors in the suite is a good addition and I want it.
What I do not want is `warnings.catch_warnings()` in `MachoBinary.py` and `MachoFileLoader.py`, because that is library behaviour changed to satisfy a test setting, and `catch_warnings` swaps the global filter list, so it is not thread-safe for a library consumed inside services.
Please drop both blocks and add a narrow ignore for that LIEF `RuntimeWarning` to the `filterwarnings` list in `pyproject.toml` instead.

**Accepted as-is:** `SmdaConfig.LOG_PATH` (I traced it: it came in with the 2018 initial commit and no file in the repo's history has ever read it), the SECURITY.md response window, the `ValueError` change to the fromDict contract, and folding the version history into CHANGELOG.md.
On the last one I checked that 180 versions went in and 180 came out, with nothing dropped.

I will handle the version bump and the CHANGELOG entry on merge, so do not add one.
The new `test_the_changelog_documents_the_current_version` couples the two and I would rather not have you rebase over it.

---

Separately, and not something to fix here: your suite surfaced a pre-existing bug in `safe_lief_parse`.
It catches only `MemoryError`, but on lief 1.0.0 the same allocation failure also arrives as `ValueError: vector::_M_fill_insert`, which is nanobind's mapping of `std::length_error`.
So `MachoFileLoader.getBaseAddress` and `ElfFileLoader.getArchitecture` raise instead of degrading.
It reproduces on master with `b"."`, and CI never sees it because we pin `lief>=0.16.0` with no upper bound and the runner resolves an older lief.
Fold the `except (MemoryError, ValueError)` widening into #227 with the rest of the loader work, and let us talk there about whether the matrix should carry a lief 1.0 leg.

r0ny123:
@danielplohmann comments are addressed.

PR #227, opened 2026-08-07
fix: repair correctness defects found by a whole-tree audit

### LINKED ISSUE #216: Synthesis: PE IAT gap-padding overwrites planted code (1221/1611 instructions corrupted on cutwail)
## Summary

`PeSynthesizer._writeThunks` pads gaps inside each DLL's IAT group by walking `va` from `min(rva)` to `max(rva)` in `ptr_size` steps and writing an ordinal-0 thunk into every slot not present in `import_rvas`. `_writeThunkAt` (`PeSynthesizer.py:135-140`) accepts **any** region that contains the address — there is no executable/data check, and the padding loop has no cap on the span it may cover.

When a report's `imported_functions` are not densely contiguous, this walks straight through the planted code and overwrites it. A warning is emitted, but the bytes are already gone.

## Reproduction

Against the repo's own `tests/cutwail_xored` fixture on `master` (4.4.1), giving the report two IAT slots 0x1000 apart — the sparse/dumped-sample shape:

```python
report.xmetadata["imported_functions"] = {
    base + 0x1000: ("kernel32.dll", "CreateFileA"),
    base + 0x2000: ("kernel32.dll", "ExitProcess"),
}
synth = report.synthesizeBinary(output_format="PE", with_strings=False)
```

Comparing each planted instruction in the synthesized image against the report's own bytes:

```
baseline (no imports):      1611 instructions checked,    0 corrupted
sparse IAT 0x1000/0x2000:   1611 instructions checked, 1221 corrupted
```

with the warning `1023 gap slot(s) in IAT group for kernel32.dll padded with ordinal-0 thunks`.

So ~76% of the recovered code in the synthesized binary is replaced by `00 00 00 80` ordinal thunks.

## Why it matters

The trigger is precisely the memory-dump/unpacked-sample case SMDA targets: multiple or sparse IATs, or a single mis-attributed slot far from its DLL group. Both the padding loop and the primary thunk writes look like they need a gate — the slot should have to land in a non-executable region, and the group span should be bounded.

Note the asymmetry already present in the same file: the `.smdaIAT` path at `PeSynthesizer.py:~397` *does* perform an overlap check before adding its region, so the intent exists, just not on this path.

## Relationship to #214

#214 tracks recording runtime-resolved/custom IATs. This is narrower and separable: it is a memory-corruption bug in the *existing* native-IAT padding path, reproducible with ordinary contiguous-format input that merely has a gap. #214's third bullet ("Decide whether custom IAT slots that fall inside code/data regions should be tolerated without corrupting those bytes") is the policy question; this issue is the concrete present-day corruption.

---

Found during a whole-repo audit of `src/smda/**` and re-verified against `master` before filing. Synthesis is documented as experimental, so filing rather than patching — happy to send a PR if you'd like a particular gating approach.

r0ny123:
## What this is

A batch of correctness fixes from a read-through of the whole tree. There are no new features here and nothing changes how SMDA is meant to work — every commit repairs a spot where the code already does something different from what the surrounding code clearly intends.

34 fix commits in one PR is a lot to look at, so a few things to make it easier: each commit is self-contained (one problem, one explanation of the mechanism, one regression test), each test was run against the code *before* its fix to confirm it actually fails there, and the commits are independent enough to cherry-pick if you'd rather take some and leave others. The grouping below is by what goes wrong rather than by file, since that's probably the more useful way in.

Anything that changes a number you might be tracking is called out in its own section near the bottom. There's exactly one, plus one recall improvement.

## 1. One bad input kills the whole run

These are the ones I'd care about most: SMDA returns an error report with zero functions for a file it could otherwise have analysed almost completely.

| Where | What sets it off |
|---|---|
| PE loader | A section at `PointerToRawData == 0x200` — the single most common value — took a `>` comparison that should have been `>=`, so the whole raw file got copied to RVA 0 |
| .NET | A corrupted `BSJB` signature, a malformed `MethodDef` row, or a missing `#US` heap each took the entire run down instead of skipping one method |
| Rust symbols | One malformed legacy symbol name produced `int('')`, and the resulting `ValueError` wasn't in any of the guarded exception sets |
| Dalvik | A crafted `call_site_ids` entry walks an arbitrary encoded-value tree; only `ValueError`/`struct.error` were caught, so nested arrays that blow the stack or index out of range escaped |
| Go symbols | A truncated function table unpacked past the end of the buffer and threw out of the provider |
| Report loading | A version string like `4.4.4-dev`, a try-range with handlers but no bounds, or a function offset that isn't an instruction start each raised out of `fromDict` or the xref helpers |
| Delphi symbols | A method entry pointing below the image base still read successfully, because the bounds check happened after `+ 2` had already made the offset positive |

## 2. Analysis says `ok`, but the output is wrong

No error, no warning — just quietly incorrect data in the report.

| Where | What was wrong |
|---|---|
| PE loader | A section with `VirtualSize == 0` produced a zero-width code area, which flipped `isInCodeAreas` into a filter that rejects every address: status `ok`, zero functions |
| Basic blocks | `is_jmp` is a one-instruction latch, but nothing cleared it when the instruction that set it was never booked, so an unrelated fall-through edge got recorded as a jump target and split a block nothing branches into |
| Basic blocks | Removing a fall-through ref deleted an unrelated branch target from `jump_targets`, so its block stopped being a block start |
| .NET blocks | Blocks ran straight through `throw`/`rethrow`/`endfinally`/`endfilter` into a separately seeded handler, double-counting and double-hashing those instructions |
| .NET | Every method was reported as a leaf — nothing ever cleared the flag. On the bundled njrat fixture that was 64 of 64, despite 1244 call instructions |
| .NET | Method names were used as if they identified a method, so the eight `.ctor` rows collapsed to one address and seven call operands were rewritten to point at a different constructor |
| AArch64 | One variable served as both the jump table's address and the anchor targets are rebased on, so every recovered target was off by `table - anchor` |
| Strings | Image-VA arithmetic was applied to a raw-file buffer that still validated against the mapped size |
| Strings | Pointer chains were always followed 4 bytes at a time, so on a 64-bit image whose base doesn't fit in 32 bits the chain died at its first link |
| Register calls | `lea` was treated as a load (the register got the *value at* the address, not the address), and a rip-relative slot was read 4 bytes wide and then had rip added to it a second time |
| Intel | A far `lcall` booked a code ref to its segment selector, because capstone separates the far pair with a comma and only `ljmp` uses a colon |
| Analysis state | Reverting a function undid three of the seven things it publishes, so it stayed in `leaf_functions` and kept its symbol entry while vanishing from the function list. The Dalvik copy of that code had the same gap plus one more field — no caller reaches it today, but it's fixed rather than left to drift |
| PDB | A PDB passed to `disassembleFile` was seeded into the providers and then wiped by the first analysis pass, which made the argument a no-op |
| Batch mode | Report stems are path-relative so same-named samples can't collide, but each input path was relativised against itself, so two directories both holding `sample.bin` overwrote each other |
| Dalvik | Five-byte sleb128 values lost their sign (`-1` came back as 34359738367), a brace inside a string constant swallowed the operand separator, and array descriptors like `[B` were never escaped |
| Dalvik | An unrenderable type came out as a lief object repr — complete with a memory address, so the report wasn't reproducible run to run |
| IDA export | Out-refs stayed anchored on the macro head rather than the sub-instruction that actually leaves the block |
| Escaping | Segment-qualified memory operands (`dword ptr es:[edi]`, anything fs/gs-relative) escaped as `CONST` instead of `PTR`, and AVX-512 registers passed through unescaped into the hash |
| Synthesis | A Mach-O section could resize the segment buffer it was written into and shift every following file offset; a code section outside `__TEXT` was mapped writable-not-executable; scattered import slots padded the gap between them with a quarter of a million thunks |

## 3. Things that never worked in the first place

Code that reads as if it does a job but structurally cannot, usually a small mismatch nothing exercised.

| Where | Why it couldn't work |
|---|---|
| Language detection | The Delphi locales check tested for exact membership in the extracted-string list *and* used the wrong case, so it could only fire on synthetic input. Separately, the Delphi string scan used `.`, which doesn't match `0x0a` outside DOTALL, so any string of exactly ten characters was skipped |
| Intel terminators | `RET_INS`/`END_INS` listed a bare `iret`, which capstone never emits — it's `iretd`/`iretq`, and `retf` becomes `retfq` under REX.W. `ud2` was missing entirely, even though the gap scan already treats its encoding as a trap and the AArch64 side already lists `udf` |
| .eh_frame | The augmentation loop bounds-checked the cursor before every character including `S` and `B`, which consume nothing, so `zRS`, `zRB` and a bare `zS` were rejected along with every FDE referencing them |
| .eh_frame | A record in the 64-bit DWARF format had its extended length read but its CIE pointer still read as 4 bytes |
| AArch64 escaping | The verification test compared `expected_mnemonic_group()` against a function that delegates straight to it — a function compared with itself, so the gate could never fail. Fixing that surfaced five mnemonics in no group table at all and the advanced-SIMD scalar forms |
| Candidate queue | `BracketQueue` sorts each bracket so the highest-scoring candidate pops first, but only the constructor ever did it, so anything added or moved later jumped the queue |
| Dominators | Path compression recursed once per ancestor link, so a long chain with a back edge near the root blew the recursion limit |
| Loader | The PE section table was located at a fixed offset picked from the bitness instead of the size the COFF header actually declares |
| Everywhere | Sixteen `struct` reads used a native-endian format. Nothing SMDA parses is host-endian — PE headers, x86 rel32 offsets, Go string words, Delphi VMT pointers are all little-endian by spec — so these decode correctly on x86 and silently wrongly anywhere else. A test now asserts the convention against the source, since a suite run on x86 can't catch it |
| Intel terminators | `sysret`/`sysexit` were missing too. capstone puts both in its own `iret` group; neither has a fallthrough successor |

## What moves

Measured across the bundled fixtures with `disassembleUnmappedBuffer` and
`WITH_STRINGS=True`, this branch against #226:

| Fixture | Change |
|---|---|
| `aarch64_switch_macho_O0` | functions 4 → 6, blocks 76 → 86, instructions 217 → 322; pic/opc digests and the function set all change |
| `komplex` | blocks 723 → 706, strings 39 → 85 |
| `cutwail` | blocks 258 → 253, strings 12 → 11 |
| `mirai_x64` | blocks 1822 → 1813 |
| `mirai_i386` | blocks 1906 → 1898 |
| `aarch64_static` | blocks 3525 → 3497 |
| `njrat` | instructions 8454 → 8453, pic digest changes |
| `bashlite` | strings 126 → 144 |

**Function sets are not identical on every fixture.** `aarch64_switch_macho_O0`
recovers two more functions, and that is the jump-table anchor fix doing exactly
what it was written to do — a switch fixture is where it should show up. The
instruction count moves with it because the two recovered functions bring their
own bodies.

**The string counts have two causes, not one.** The pointer-width fix only
explains `komplex`, which is based at `0x100000000`. `bashlite` is 64-bit but
based at `0x400000`, where the high dword is zero and pointer width cannot
matter, yet it gains 18. `cutwail` is 32-bit, where the word size is 4 on both
sides, and it loses one. Both of those are d0de195 reading strings from the
mapped image instead of the raw file, which moves counts in both directions.

## What I deliberately did not change

Over-detection is left alone throughout — none of this is a precision fix.

To be clear about one overlap: this PR does **not** close #216, even though it touches that exact padding loop. What I bounded here is the runaway output growth — a DLL whose recovered slots are scattered across the image no longer gets a quarter of a million placeholder thunks. The corruption #216 describes has a different cause: `_writeThunkAt` accepts any region containing the address, including executable ones. That issue's reproduction (two slots exactly `0x1000` apart) still reproduces on this branch — I re-ran it to be sure. I left it alone deliberately, because #216 asks which gating approach you'd prefer and that seemed like your call rather than something to slip into an unrelated PR.

One fix I wrote and then reverted, which is worth knowing about: the jump-table docstring documents a pattern-B bound check of the form `cmp [ebp+arg_4], size`, and `RE_CMP_SIZE` only matches a register operand, so that shape is never recognised. Widening the regex to accept a memory operand looked obvious but measurably regressed things — `_findJumpTableSize` takes the first `cmp` in the backtrack window without tying it to the jump index, so unrelated compares started matching, and on the bundled `mirai_x64` fixture that enabled a bogus direct table and cost a recovered function (151 → 150 functions, 1813 → 1933 blocks). Recognising pattern B properly needs index tracking rather than a wider pattern, so I left the regex alone and recorded the constraint and the measurement in a comment next to it.

## Validation

```
make test       # 1023 passed, 1 skipped
make lint       # ruff check + ruff format --check clean
make typecheck  # exit 0
```

Diff coverage is 100% against `upstream/master` (412 changed lines, 0 uncovered). Every bundled fixture was re-run before and after to compare recovered function sets and block counts; the only movement is the section above.

## Note on the diff

This branch is based on the branch of #226, which is still open against the same base, so until that lands this PR's file list shows both sets of changes. The 39 commits listed here are the content of this PR, and the diff will shrink on its own once #226 merges.


danielplohmann:
This is a lot of careful work and the commit discipline makes it reviewable: one defect per commit, the mechanism spelled out, and several commits volunteering what they deliberately did not fix.
The reverted pattern-B jump-table attempt with its measurement recorded at the regex is exactly the kind of note I want left behind.

I checked a few of the trickier ones by hand and they hold up.
The iterative `COMPRESS` is equivalent to the recursion it replaces, since the recursion mutates only on unwind and collecting the path first changes nothing.
The `rust_v0` const backref now matches the three sibling `backref_printer()` call sites verbatim, where the old form dropped the child printer's output on the floor.
`0x18 + SizeOfOptionalHeader` is the right way to reach the section table, and it agrees with what the old `0xF8`/`0x108` constants actually meant.

The suite here runs 1015 passed, 7 failed, 2 skipped, 818 subtests in 5m36s.
All 7 failures are the `safe_lief_parse` gap, not yours: I re-ran `test_fuzz_loaders.py` from this branch against #226's `src` and got the same failures, 9 of them.
More on that below.

Two things before this goes in.

**1. The AVX-512 widening regresses the masked form.**

`re.search("zmm[0-9]+", op_field)` became `re.fullmatch(r"(?:[xyz]mm|k)[0-9]+", op_field)`, and a masked operand no longer matches:

```
                             #226          #227
vaddps zmm0, zmm1, zmm2      XREG          XREG
vaddps zmm0 {k1}, ...        XREG          zmm0 {k1}   <- regression
kmovw  k1, eax               k1            XREG        <- fix
vaddps ymm16, ymm17, ymm18   ymm16         XREG        <- fix
```

The substring search used to catch `zmm0 {k1}`, and the anchored pattern does not, so a raw register name now leaks into the escaped operand.
That is the exact leak the commit message says it closes.
It needs an optional `\s*\{[^}]*\}` tail plus the `{z}` form, and a test: nothing in `tests/` covers a masked operand today.

**2. Please rewrite "The one number that moves" against a full sweep.**

The section promises exactly one moved number plus one recall improvement.
Measured across the bundled fixtures with `disassembleUnmappedBuffer` and `WITH_STRINGS=True`, comparing this branch against #226:

| Fixture | Change | In the PR body |
|---|---|---|
| `aarch64_switch_macho_O0` | functions 4 to 6, blocks 76 to 86, instructions 217 to 322, pic/opc/function-set all change | no |
| `komplex` | blocks 723 to 706 | no |
| `mirai_i386` | blocks 1906 to 1898 | no |
| `njrat` | instructions 8454 to 8453, pic digest changes | no |
| `bashlite` | strings 126 to 144 | no |
| `cutwail` | strings 12 to 11, downward | no |
| `cutwail` / `mirai_x64` / `aarch64_static` | blocks 258 to 253, 1822 to 1813, 3525 to 3497 | yes |
| `komplex` | strings 39 to 85 | yes |

The `aarch64_switch_macho_O0` movement is the jump-table anchor fix doing its job, and recovering two more functions on a switch fixture is the point of that commit.
It does contradict "Function sets are identical on every fixture", though, so it has to be in the section rather than left for someone to find.

The string counts have a second cause the body does not name.
`komplex` is based at `0x100000000`, so its 39 to 85 really is the pointer-width fix as described.
`bashlite` is 64-bit but based at `0x400000`, where the high dword is zero and by your own reasoning the pointer width cannot matter, yet it gains 18.
`cutwail` is 32-bit, where the word size is 4 on both sides, and it loses one.
Both of those are d0de195 reading strings from the mapped image instead of the raw file, which is the right fix and moves counts in both directions.

**3. Fold the `safe_lief_parse` widening in here after all.**

I asked for this in #226 and said it could wait for this branch, and the numbers argue for doing it now rather than later.
On #226 two of these tests failed; from this branch, nine.
The count grows because every local run seeds more falsifying inputs into `.hypothesis`, so it gets worse for anyone working in the tree, while CI stays green only because the runner resolves an older lief.
`except (MemoryError, ValueError)` closes it.

**Checked and fine as it stands.**

The segment-qualified operand fix (`dword ptr fs:[0x30]` going from `CONST` to `PTR`) is correct and does not touch the stored hashes.
`pic_hash` runs through `escapeBinary` and `opc_hash` through `escapeToOpcodeOnly`, and neither reaches `escapeField`, which the unchanged Intel pic/opc digests on cutwail, bashlite, mirai_x64 and mirai_i386 confirm.
So no escape-version bump is needed for it.
`getEscapedOperands()` is public and capa consumes SMDA as a backend, so it still deserves a changelog line.

Four notes, none of which should hold this up.

The CIL `END_INS` additions change .NET block structure, so pic_hash values for methods containing `throw`/`rethrow`/`endfinally`/`endfilter` will not match previously stored ones.
This is unlike the v4.4.2 Dalvik case: that was an escaping change, so recalculating from stored blocks on import produced the corrected value, while here the CFG itself changed and recalculation reproduces the old hash.
Only re-analysis fixes it, so leave `CIL_PIC_HASH_ESCAPE_VERSION` alone and give it the same bold warning v4.4.2 got in the changelog.

`BracketQueue._sort_bracket` now runs on every add and every update, rebuilding the bracket dict each time, which makes filling a bracket quadratic against `MAX_FUNCTION_CANDIDATES = 200000`.
The fix is right and the placement is expensive; a dirty flag sorted at pop time keeps both.
Low urgency, since `CANDIDATE_QUEUE` defaults to `PriorityQueue`, which also means the perf gate cannot see it.

`_UNUSABLE_DEMANGLERS` blacklists permanently and process-wide on `subprocess.TimeoutExpired` as well as `OSError`.
One timeout under load then disables the demangler for every later analysis in an MCRIT worker, so restricting the blacklist to `OSError` would be safer.

`_analyzeCall` returns early on any `lcall`, including the memory-indirect far form that used to reach the indirect arm.
Rare, but it is wider than the fix the commit describes.

Merge order is #226, this, then #228.

danielplohmann:
Merged, thanks.

This went in by rebase rather than through the button, since #226 was force-pushed with rewritten commits before it landed and this branch still carried the originals, which GitHub reported as a conflict.
The rebase dropped all 13 of them as already applied and replayed this PR's 40 commits with no conflicts, so the content that landed is the branch as you pushed it.
The only file that differs from your head is `.pre-commit-config.yaml`, where master's newer `ruff-pre-commit` pin at v0.16.1 won.

Master is now at de8eb95.
The suite runs 1026 passed, 2 skipped, 822 subtests, and `ruff check` plus `ruff format --check` are clean on 229 files.

Both fixes check out.
The escaper pattern now covers the zeroing form as well, which I had not thought to ask for: `zmm0`, `zmm0 {k1}` and `zmm0 {k1} {z}` all escape to XREG, `k1` and `ymm16` with them, and `dword ptr fs:[0x30]` still escapes to PTR.
Widening `safe_lief_parse` to `(MemoryError, ValueError)` also cleared the seven fuzz-loader failures this branch was carrying locally.

Every number in the rewritten "What moves" section reproduces here.
One leftover for whoever edits it next: the intro still says "There's exactly one, plus one recall improvement", which the section below it now contradicts.

I will write the changelog entry, including the note that .NET pic_hash values for methods containing throw, rethrow, endfinally or endfilter no longer match previously stored ones.

#228 is next, and it needs the same rebase.

PR #228, opened 2026-08-09
fix: close the open synthesis, import recovery and Go symbol issues

### LINKED ISSUE #214: Synthesis: handle runtime-resolved / custom IATs and dynamic API refs (future work)
## Context

PR #198 adds experimental binary synthesis (`SmdaReport.synthesizeBinary()`), which rebuilds import tables from `xmetadata["imported_functions"]`. The current review fixes (see PR #198) harden the **native IAT** path (non-contiguous slot padding, defensive fallbacks, etc.).

This issue tracks a **broader, separate** limitation that should NOT be addressed in PR #198:

- SMDA records imports from the **native IAT** only. Malware frequently resolves APIs at runtime and caches them in **custom / secondary IATs** that the sample builds and maintains itself, often with format violations (no zero terminators, scattered / non-contiguous slots, possibly interleaved with code or data).
- These dynamic API references are **not listed among the imported symbols**, so the synthesized binary's import view is inconsistent with what the sample actually uses.

## Open questions / future work

- Detect and record runtime-resolved API references (e.g. from dataflow-resolved indirect calls) so they appear in `imported_functions`.
- Make synthesis robust to custom IAT layouts (e.g. a contiguous synthetic IAT backed by `.bspack`, rather than depending on the original possibly-broken IAT structure for import enumeration).
- Decide whether custom IAT slots that fall inside code/data regions should be tolerated without corrupting those bytes.

## Scope

Out of scope for PR #198 — tracked here so the simpler review fixes can land without blocking on this larger, partly-open-ended problem.

### LINKED ISSUE #216: Synthesis: PE IAT gap-padding overwrites planted code (1221/1611 instructions corrupted on cutwail)

### LINKED ISSUE #217: Synthesis: ELF section header table drops unnamed sections (stripped ELF re-loads with base 0, loses functions)
## Summary

`ElfSynthesizer._assembleFile` builds the section header table from
`named_sections = [s for s in sections if s["name"]]` (`ElfSynthesizer.py:~329`).

A stripped ELF — where SMDA recovers real VA ranges but every `code_sections` entry has `name == ""` — therefore synthesizes to a file with `e_shnum == 2` (the null section plus `.shstrtab`) even though the real ranges are still emitted as PT_LOADs. Re-loading that file gives `base_addr = 0` and loses functions.

## Reproduction

`tests/mirai_x64_xored` on `master` (4.4.1), with the section names cleared to model the stripped case:

```python
report.code_sections = [("", s, e) for _, s, e in report.code_sections]
synth = report.synthesizeBinary(output_format="ELF", with_strings=False)
```

```
all-unnamed code_sections -> e_shnum = 2 | names: ['', '.shstrtab']
  PT_LOADs emitted: 4
re-loaded: 145 functions (original 151), base 0x0 (original 0x400000)
```

Four PT_LOADs describing the real layout, but no section headers describing them; **6 of 151 functions are lost** on re-disassembly and the base address is wrong.

(For contrast, the unmodified fixture has 9 of 10 sections named and synthesizes with `e_shnum == 10`.)

## Suggested direction

Fabricating names for unnamed ranges (`.smda0`, `.smda1`, …) would preserve the section header table and keep the synthesized file self-describing. That also removes the `base_addr = 0` case, which is one of the two paths into the wrong-base problem filed separately.

---

Found during a whole-repo audit of `src/smda/**` and re-verified against `master` before filing. Synthesis is documented as experimental, so filing rather than patching.

### LINKED ISSUE #218: Synthesis: first segment not congruent with the image base, so SMDA's own loaders derive the wrong base_addr
## Summary

Both `ElfSynthesizer` (`:~318`) and `MachoSynthesizer` (`:~539`) place the first segment with

```python
file_cursor = align_up(file_cursor, PAGE_SIZE) + va_start % PAGE_SIZE
```

That is page-congruent, but it starts the first segment *after* the header page rather than mapping the header at the image base, which is what the ELF and Mach-O conventions expect. SMDA's own loaders derive the base address from `min(section.va - section.offset)`, so they read back a base that is off by a page or entirely wrong.

## Reproduction

`tests/mirai_x64_xored` on `master` (4.4.1), unmodified, round-tripped through ELF synthesis:

```
original:  151 functions, base 0x400000
re-loaded: 174 functions, base 0x3fe000
```

The base comes back 0x2000 low, and every derived RVA in the re-disassembled report shifts with it.

For Mach-O, every synthesized segment ends up with `vmaddr - fileoff == 0xFFFFF000`, so `MachoFileLoader._calculate_base_address` returns `0xFFFFF000` instead of `0x100000000`. (Note that on current `master` a negative/again-wrong base is now filtered by the loader, but the synthesized layout is still not what the format requires.)

The Mach-O output is additionally not dyld-valid: `__TEXT` must start at file offset 0 and cover the mach header.

## Related, same subsystem

While looking at this I also noticed `MachoSynthesizer` writes its LC_SYMTAB/LC_DYSYMTAB payloads past the last segment — inside no segment at all — and emits no `__LINKEDIT` (nor `__PAGEZERO`). LIEF parses it, but dyld and any consumer that resolves the symbol table through segment mapping will see nothing. Mentioning it here rather than as a separate issue since it is the same "output parses but is not format-valid" theme.

---

Found during a whole-repo audit of `src/smda/**` and re-verified against `master` before filing. Synthesis is documented as experimental, so filing rather than patching.

### LINKED ISSUE #219: Synthesis: minimal-PE string-coverage section can get RVA 0 and overlap .text/.smdaIAT
## Summary

In the minimal-PE path, the synthetic string-coverage section is added with no overlap or lower-bound check (`PeSynthesizer.py:~415-440`):

```python
str_vaddr = align_down(min_s, section_alignment)
str_size  = align_up(max_s, section_alignment) - str_vaddr
regions.append({"name": ".rdata", "vaddr": str_vaddr, ...})
```

The `.smdaIAT` path immediately above (`:~397`) *does* check `overlaps` before appending its region, so the asymmetry looks unintentional rather than deliberate.

Consequences when a string ref sits low in the image:

- `align_down` of a low RVA (e.g. 0x200) yields **RVA 0**, which is below `SizeOfHeaders`. A section overlapping the headers is invalid per the PE spec and Windows rejects it.
- The resulting range can fully contain `.text` and `.smdaIAT`, so strings planted under `.text`'s range are silently lost on mapping.

With string refs at `base+0x200` and `base+0x50000`, LIEF reports `.rdata va=00000000 vs=00051000` containing both `.text va=00001000` and `.smdaIAT va=00003000`.

## Note on verification

Unlike the sibling issues I filed alongside this one, I confirmed this one by reading the code and by the asymmetry with the `.smdaIAT` overlap check rather than by re-running it end to end — my attempts to drive the minimal-PE path on the repo fixtures kept landing in the full-PE path, which has a different layout routine. Flagging that explicitly so you can weigh it accordingly; the mechanism is straightforward but I would not call it reproduced by me.

## Suggested direction

Mirror the `.smdaIAT` overlap check, and clamp `str_vaddr` to at least `SizeOfHeaders` (or the first section's RVA) so the coverage section can never start at 0 or straddle an existing region.

## Related, same subsystem

`PeSynthesizer._buildOptionalHeader` (`:~296-316`) copies the original optional header wholesale, so the export/resource/reloc/TLS/debug/IAT/delay-import directory RVAs survive while their targets are now zero-filled; only the IMPORT entry is rewritten. Parsers that follow those directories read zeroed structures instead of failing cleanly. Same subsystem, same "output parses but is not internally consistent" theme.

---

Found during a whole-repo audit of `src/smda/**`. Synthesis is documented as experimental, so filing rather than patching.

### LINKED ISSUE #223: Validate Go pclntab symbol extraction & function seeding across all pclntab versions
### Context
PR #220 hardened `GoLabelProvider._parse_pclntab` (finding 12): it now catches
`struct.error` (previously a dead `except ValueError`, since `struct.error` is not a
`ValueError`, so one bad entry silently dropped *every* Go symbol) and switched the
parser to explicit little-endian (`"<I"`/`"<Q"`), since it was using native-endian
while the detector happily accepts big-endian magic.

That fix makes the parser **fail gracefully**, but it does **not** prove the parser is
**correct** across the range of Go `pclntab` formats SMDA may encounter. We have a
corpus of Go binaries compiled with (almost) every Go version — this issue tracks
validating symbol recovery and function-candidate seeding against that corpus.

### What needs verifying (per pclntab version)
1. **Offset discovery** — `.gopclntab` section (ELF/PE) and `__gopclntab` (Mach-O),
   plus the `runtime.pclntab` symbol fallback, resolves to a usable header.
2. **Symbol extraction** — `getFunctionSymbols()` / `getSymbol()` return the expected
   function names/offsets (no truncated-table drops, no silent `delete=True` discards).
3. **Function candidate seeding** — recovered Go symbols actually seed FEPs / get
   resolved onto functions via `symbol_providers` (`_updateLabelProviders` →
   `getSymbolCandidates` / `resolveSymbol`), i.e. recall is correct, not just the dict.

### Coverage gaps to close
- Current `_PCLNTAB_VERSIONS` marker map only recognizes **1.12 / 1.16 / 1.18 / 1.20**
  (`0xFFFFFFFB/FB/FA/F0/F1`). Go has shipped many versions since, and the pclntab magic /
  field layout has shifted across 1.2 → 1.16 → 1.18 → 1.20+; newer toolchains (1.21–1.2x)
  and any pre-1.12 formats are **unverified**. Map every version in the corpus to its
  marker and confirm handling, adding markers/branches as needed.
- Variation axes to sweep in the corpus:
  - Go version (full range available)
  - `GOOS` (linux / windows / darwin) and `GOARCH` (amd64 / 386 / arm64)
  - stripped vs. unstripped
  - PIE vs. non-PIE / CGO on/off

### Acceptance
- A reproducible harness that disassembles each corpus sample and emits a matrix of
  (go-version × os × arch × stripped) → symbols-recovered / functions-seeded counts,
  with a manual or reference baseline (e.g. `runtime` symbol names, or `go tool nm`).
- Any version that currently parses wrong / drops symbols gets a regression test added
  to `tests/` (mirroring the per-finding tests in PR #220).
- Document the supported pclntab version range (or "best effort" caveats) in the report
  once validated.

### References
- `src/smda/common/labelprovider/GoLabelProvider.py` (`_parse_pclntab`, `getPcLntabInfo`,
  `getPcLntabOffset`, `update`)
- PR #220 (the struct.error / endianness hardening this validation is meant to confirm)

r0ny123:
Closes #214, closes #216, closes #217, closes #218, closes #219, closes #223.

This stacks on #227. Its 52 commits are included here unchanged and are not what this PR proposes; the 13 commits from `fix(core): keep synthesized PE sections and thunks off recovered code` onward are. Merging #227 first reduces this PR to those 13. The dependency is real rather than accidental: the synthesis fixes below build on files #227 already corrects, and cherry-picking them onto master conflicts.

## Synthesis

**#216 — the synthetic IAT overwrote recovered code.** Gap-padding and terminator slots were planted at addresses the report never listed as imports, so they landed on planted code bytes. Those writes are now confined to non-executable regions. On the PE fixture: 218 of 253 recovered blocks corrupted, now 0.

**#217 — unnamed ELF ranges never reached the section header table.** A serialized report carries no section names, so the file described its `PT_LOAD`s with no sections at all and re-loaded at base 0. Ranges now get non-colliding `.smdaN` names: `e_shnum` 2 to 10, re-load base `0x0` to `0x400000`.

**#218 — synthesized ELF and Mach-O images mapped at the wrong base.** Each segment now takes the lowest page-congruent file offset that clears what precedes it. The ELF program header table moves to the file tail so only the 64-byte header sits below the first segment, and Mach-O `__TEXT` maps from offset 0. ELF base `0x3fe000` to `0x400000`; Mach-O `vmaddr - fileoff` `0xfffff000` to `0x100000000`, with 35 of 35 functions recovered.

**#219 — the minimal-PE string section straddled the headers.** A section at RVA 0 overlapped the header block and swallowed `.text` and `.smdaIAT`. It is now clamped to section alignment and skipped when it would overlap an existing section. The same class caused an abort in `struct.pack_into` when an import slot below the image base gave `.smdaIAT` a negative RVA; that is fixed alongside it.

## #214 — runtime-resolved API references

`imported_functions` was built purely from the static import table, so a sample that constructs its own import table at runtime described none of the APIs it actually uses. On a memory dump, where the mangled IAT does not parse at all, the report claimed zero imports while analysis had already resolved hundreds — and a binary synthesized from that report inherited the same empty view.

Every API resolved during analysis now carries the pointer slot it was reached through, and those merge into `imported_functions` at report construction with the static table winning on collision. Analysis itself still reads only the static map, so the ELF GOT-base oracle keeps a stable input.

`call/jmp dword|qword ptr [<reg> + <disp>]` is also resolved now. It was previously dropped as "a minority of calls", but it is how a runtime-built table is indexed. The base register comes from the existing backward pass and the slot is base plus displacement. Worth noting for review: capstone prints a displacement below 10 as a bare decimal digit, so matching only `0x`-prefixed operands would have missed the common pointer strides.

This path records the API and its slot only — it books no code reference and no function candidate — so CFG recovery cannot move on a heuristic register value. Measured on the `asprox` dump fixture: 0 to 95 imports, all 95 resolvable by LIEF after synthesis, with function, block and instruction counts unchanged.

A sweep for the same class found two further sites that had a slot address in hand and discarded it: the aarch64 `ldr x8, [slot]; blr x8` shape, which is the normal way an import is called on an ISA with no absolute `call [addr]` form, and the IDA exporter, whose import map is already keyed by the import entry. Both now record it. The rip-relative register-load pattern also only ever matched `[rip + 0x...]`, so a slot behind the instruction stopped the backward pass resolving the call at all.

One shape stays off by default. Some binaries never call through an import slot at all: a language runtime that manages its own syscall dispatch loads the slot into a register, stores the pointer into a per-thread structure, and issues its one indirect call from a single shared routine, so a call/jmp-driven pass attributes nothing to any call site. `RECORD_IMPORT_SLOT_LOADS` adds an opt-in pass that records the API against the loading instruction instead. On a 64-bit Go PE whose import table lists 48 kernel32 entries that no instruction ever calls through, it recovers 102 API references over 46 distinct imports in 57 functions. It is off by default because it widens an apiref from "calls this API" to "references this API", which consumers should choose rather than inherit; like the paths above it books no code reference and no function candidate, so function, block and instruction counts are identical with it on and off.

## #223 — Go pclntab recovery

The premise that Go 1.21+ shipped unrecognised markers does not hold: Go defines exactly four table magics and the 1.20 one is still current, so the marker map was already complete. The layout handling behind it was not.

A 1.16 `_func` record opens with a pointer-wide entry address, where 1.18 and later carry a `uint32` offset from `textStart`. The scan locating a function's record read four bytes for both, then skipped four more on 64-bit to reach the name offset. That works only when the entry's high half is zero and the target is little-endian, so a 64-bit Mach-O image — based at `0x100000000`, where no entry fits in 32 bits — dropped every symbol, and a big-endian target matched the low half second and read `args` instead of `nameoff`, yielding truncated names against correct addresses. The entry is now read at its own width. Symbol names also render `U+00B7` as `.` rather than `:`, matching `debug/gosym`, so wrapper names read as `go tool nm` prints them.

Separately, the text start was read from the pclntab header, a field Go no longer maintains — it is a placeholder now, dropped because keeping it accurate needed a relocation, which is why `debug/gosym` takes the value from its caller and `cmd/internal/objfile` reads it off the container. Only Go's own linker left the old field usable; hand the layout to the system linker, as any cgo build does, and it reads 0, so every recovered address was short by the whole text base. It now comes from the container — a `runtime.text` symbol when present, otherwise the text section address — with the header field kept as the fallback for a buffer that has no container behind it. Across Go 1.18 to 1.25 that moves cgo builds from 0.0 to 0.986 symbol recall, and leaves internally linked builds untouched because there the two sources agree.

`tests/go_corpus/build_matrix.py` is the reproducible harness the issue asks for. It builds a sample with each Go toolchain across `GOOS`/`GOARCH` and link mode, compares recovered symbols against `go tool nm` — which reads the same table through `debug/gosym` — and deletes each binary as it is measured. A Go hello-world is roughly 2 MB and the sweep builds a few hundred, so only the result matrix is committed.

The recorded run covers **322 cells**, sweeping every axis the issue lists: Go version, `GOOS`/`GOARCH`, stripped, PIE, and cgo. All four table layouts are exercised, across ELF/PE/Mach-O, both byte orders and both pointer sizes. Go published no darwin/arm64 toolchain before 1.16, so the oldest layout is covered instead by the `gofmt` shipped in the Go 1.15 distributions, itself a binary built by that toolchain. `tests/testGoCorpusMatrix.py` asserts the recorded results, and `tests/go_corpus/README.md` documents the supported version range and how to regenerate the matrix.

Every cell reaches 0.9855 recall or better with one documented exception, which a test pins as the only cell allowed to fall short: a pre-1.18 table in an externally linked Mach-O recovers nothing, because those layouts store a pointer-wide entry per function and Apple's linker writes pointer-wide fields as dyld chained-fixup entries — `0x0020000000001a50` is how `0x100001a50` is stored. That is a Mach-O rebase-map concern rather than a pclntab one, and is left for a separate change; `AArch64FunctionCandidateManager._machoFixupState` already builds such a map and declines chained slots for the same reason.

Five differences between the two readings are presentation rather than disagreement and are normalized away in the harness: `nm`'s `.abi0` suffixes, its section boundary markers, Mach-O's leading underscore on assembly symbols, the C functions an externally linked binary carries that the pclntab never describes, and the interior entry points such as `runtime.duffzero+8` that external linking puts in the symbol table for a function the table lists once.

## Validation

`make test`: 1094 passed, 1 skipped. `make lint` and `make typecheck` clean. Diff coverage 100% over 631 changed lines. The synthesis and report fuzz corpora replay clean.

Recovered CFG output is byte-identical to master across the frozen `tests/*_xored` fixtures — function, block and instruction counts plus per-function shape — so none of this moves recovery results. The three regression tests added for the pclntab layout fixes were each confirmed failing on the pre-fix tree.


danielplohmann:
On the evidence this is the cleanest of the three, and the Go work is the strongest single piece in the stack.
The reasoning about why the pclntab header's text start went stale matches what `debug/gosym` and `cmd/internal/objfile` actually do, and the endianness handling, the entry-width fix and taking the value from the container all read correctly.
I have no findings on that part.

What I measured against #227, so the numbers below are this PR's 13 commits only.

Recovery does not move anywhere.
Across seven fixtures with `WITH_STRINGS=True`, the function count, block count, instruction count, apiref count, `imported_functions`, function-set digest and pic digest are identical on every one, with no measurable change in wall time.
That is a stronger result than #227 delivered, and it is what makes this one comfortable to take.

The #214 headline reproduces exactly.
On the asprox dump, imports go from 0 to 95 while apirefs stay at 546 and functions, blocks and instructions stay at 106, 2121 and 15714 with an unchanged pic digest.
One thing to add to the PR body: this only fires with `API_COLLECTION_FILES` configured.
Under SMDA's default empty config `resolveApi` has nothing to resolve against, and the same fixture measures 0 to 0.

#218 for Mach-O is confirmed.
The synthesized Mach-O for komplex moves its image base from `0xfffff000` to `0x100000000`, and LIEF's `imagebase`, `min(p_vaddr - p_offset)` and `min(sh_addr - sh_offset)` all agree on it.
Re-parsing the synthesized images and comparing every instruction back against the report also shows cutwail and asprox fully intact in all three formats.

Two things before this goes in.

**1. The ELF base fix overshoots on one fixture.**

cutwail goes from `0x3fff000` to `0x4000000`, which is exactly right.
komplex goes from `0xffffe000` to `0x100001000` against a report base of `0x100000000`, so the error moves from two pages low to one page high rather than disappearing.
All three metrics agree on that, and `min(p_vaddr)` is `0x1000012b0`, so the first recovered content on this fixture sits above the base page.
Is that the intended outcome there, or should the ELF path reserve the base page the way the Mach-O path now does for `__TEXT`?

**2. `resolveComputedImportSlots` needs the per-block cap its sibling has.**

`resolveRegisterCalls` carries `MAX_INDIRECT_CALLS_PER_BASIC_BLOCK` with a comment naming the sample that forced it, a Go binary with 130k register calls.
The new pass runs the same backward walk with no such backstop.
Measured on the bundled corpus it tops out at 19 sites in a single function, with 51 across asprox, 4 in cutwail and none in komplex or njrat, so nothing here demonstrates a problem.
However, `call dword ptr [<reg> + <disp>]` is the C++ virtual dispatch shape and will be far denser on other samples than `call <reg>` ever was.
I would rather close the asymmetry now than wait for the sample that finds it.

**Notes, none of them blocking.**

`RESOLVE_COMPUTED_IMPORT_SLOTS` defaults on while `RECORD_IMPORT_SLOT_LOADS` defaults off.
The split is well argued, and the on-by-default one still changes `xmetadata.imported_functions` for everyone who re-analyses, so it wants a changelog line.

komplex loses 2 of 4914 instructions through ELF and Mach-O synthesis on both branches.
That is pre-existing rather than introduced here, but it means the synthesis round-trip is not lossless on that fixture.

`_mergeImportedFunctions` lowercases the DLL name for the dynamic entries, and `parse_pe_imports` already lowercases the static ones, so the two halves of the merge agree.
Checked rather than assumed.

`testGoCorpusMatrix.py` asserts the recorded matrix instead of rebuilding it, so it pins the claim without re-verifying it in CI.
The PR body and `tests/go_corpus/README.md` both say so plainly, and given a Go hello-world is roughly 2 MB and the sweep builds a few hundred, that is the right trade.

**Suite.**

1091 passed, 2 failed, 2 skipped, 1466 subtests in 5m47s.
Both failures are the `safe_lief_parse` gap I asked you to fold into #227, not anything here.

The count across the three branches is worth knowing: 2 failures on #226, 7 on #227, 2 on #228, for the same defect and the same class of input.
The number tracks how many falsifying examples each worktree's `.hypothesis` database happens to hold rather than anything about the code.
A fresh checkout starts at zero and climbs as you work in the tree, which is exactly the shape that keeps a defect invisible in CI.

Merge order is #226, then #227, then this.

danielplohmann:
Merged, thanks. This one was the cleanest of the three to take.

Same procedure as #227: rebased rather than pushed through the button, since the branch still carried the pre-merge commits of #226 and #227, which GitHub reported as a conflict.
The rebase dropped those as already applied and replayed this PR's 13 commits with no conflicts.
The only file differing from your head is `.pre-commit-config.yaml`, where master's newer `ruff-pre-commit` pin at v0.16.1 won.

Master is now at 5313782, with the release entry on top at 2f6207f.
The suite runs 1098 passed, 2 skipped, 1470 subtests, and `ruff check` plus `ruff format --check` are clean on 233 files.

Both findings are closed properly.
The per-block cap mirrors `resolveRegisterCalls`, and keying it on the block's start address rather than on the instruction tuple is the right call.
The ELF base now reserves the gap between the report's own `base_addr` and the first segment, which is the general form of the fix rather than a patch for one fixture: measured here, `komplex` moves from `0xffffe000` to `0x100000000`, `cutwail` from `0x3fff000` to `0x4000000`, and `asprox` stays correct at `0x8d0000`.
Bounding it with `MAX_BASE_PROLOGUE` is a good instinct, since nothing else stops a report from claiming megabytes of empty space below its first section.

One thing I chased and dropped, in case it saves someone else the detour: LIEF reports `imagebase = 0x4001000` for the synthesized Mach-O of `cutwail`, which looks like the same defect surviving on that path.
It is not. `__TEXT` sits at vmaddr `0x4001000` with fileoff `0x1000`, so `min(vmaddr - fileoff)` is `0x4000000` and correct.
LIEF's Mach-O `imagebase` reports the first segment's vmaddr, which is a different quantity from the one the ELF fix targets.

Still open, and not from this PR: `komplex` loses 2 of its 4914 instructions through ELF and Mach-O synthesis.
It predates all three PRs and reproduces on master, so the synthesis round-trip is not lossless on that fixture yet.
Worth its own issue rather than a fourth PR on this stack.

The release entry records what moved for consumers, including that stored CIL `pic_hash` values for methods containing `throw`, `rethrow`, `endfinally` or `endfilter` no longer match a freshly analyzed one.

PR #231, opened 2026-08-10
perf: cut a third of the per-instruction work out of the disassembly pipeline

r0ny123:
> **Stacked on #228.** This branch is cut from `fix/upstream-issues-2026-08-09`, so GitHub currently shows that PR's commits here as well. **The 11 commits below are the ones this PR asks you to review** — everything else belongs to #228 (which itself stacks on #227) and will disappear from this diff once #228 merges. Merge order: #227, then #228, then this.

## Summary

A profile-driven pass over the disassembly pipeline. Every change is a pure performance change: recovered functions, blocks and instructions are unchanged, and `computeReportIdentityHash` is byte-identical on every binary tested.

Against the branch point, primitive calls across the eight benchmark fixtures drop **9,673,670 → 6,420,772 (-33.6%)**. Against `master`, the same suite is **-33.3% primitive calls and -25.8% wall time** with identical function counts on all eight fixtures.

## The 11 commits

| | commit | change |
|---|---|---|
| `a41116c` | `perf(dalvik)` | skip dead claimed intervals in the orphan `code_item` scan (a quadratic) |
| `028192c` | `perf(intel)` | give the stub-chain block patterns a literal prefix so sre can fast-search |
| `2342547` | `perf(common)` | memoize the Delphi and C++ language evidence scores |
| `13bb5b3` | `perf(common)` | hoist loop invariants out of the block derivation loop |
| `33bb1c2` | `perf(utility)` | run large inputs on a memory-capped worker pool |
| `11ac7d9` | `perf(common)` | cut per-instruction call overhead out of the CFG and report paths |
| `3495f35` | `perf(intel)` | dispatch instructions by table and precompile the escaper regexes |
| `2ac224c` | `perf(cil)` | share one metadata parse and cut per-operand and per-opcode overhead |
| `74530dd` | `perf(labels)` | anchor the language scans and build each report dict once |
| `c581c65` | `perf(labels)` | find the pclntab header by literal anchor instead of scanning every offset |
| `05d322d` | `perf(loaders)` | drop the LIEF-keyed caches and shallow-decode the Dalvik start sweep |

Each commit body carries its own measurement and the argument for why it cannot move output.

## Motivation

The hot regions were taken from a cProfile run over all backends rather than picked by inspection, and the work falls into four recurring shapes:

- **Per-instruction call overhead.** `analyzeFunction` called four one-line state accessors per instruction; `addCodeRef` used `dict.get` where `k in d` compiles to a single `CONTAINS_OP` with no call; `getBlocks` re-read the same attributes and recomputed `len(...) - 1` every iteration.
- **Regex patterns that cannot fast-search.** The stub-chain and pclntab patterns both begin with an alternation, so the engine entered the matcher at every offset of the mapped image. Rewriting them to start with a literal lets sre emit a prefix search.
- **Work computed and then discarded.** The Dalvik start sweep decoded every offset in full — operand strings, register lists, reference resolution — and used two fields. The CIL module was parsed twice per analysis, once by the symbol provider and once by the disassembler.
- **A cache that cost more than the work.** Eight loader helpers were memoized with `lru_cache` keyed on a LIEF object. LIEF's `__hash__` is a deep visitor: measured on a real ELF, hashing costs 0.259 ms while the cached work — sorting the section list — costs 0.056 ms. Every cache *hit* was ~4.6x more expensive than recomputing.

## Changed behavior

None intended, and none observed. `DisassemblyResult.addCodeRefs` and `SmdaInstruction.getEscapedBinary` stay in place as public API; only the hot paths stop routing through them.

The one deliberate widening is in `BatchProcessor`: inputs at or above a size threshold now run on a separate pool whose concurrency is capped by a memory budget derived from physical RAM, with workers recycled per file. Peak worker RSS scales with input size, so a corpus mixing multi-megabyte binaries with small ones previously sized its whole pool for the worst case.

## Validation

Output identity, all at the default configuration:

- **8 benchmark fixtures** (intel, AArch64, Dalvik, CIL): identical `report_hash`, `num_functions`, `num_instructions`, `num_blocks`, and per-function block/instruction map.
- **251 real binaries** (the bundled malpedia corpus, 264 samples of which 251 analyse): **276,793 functions / 13,023,531 instructions / 2,220,554 basic blocks, zero delta, zero hash mismatches.** The 13 that do not analyse fail identically on both sides.
- **An 8.1 MB Go PE**: byte-identical report (8,687 functions), 23.1 s → 19.8 s.

Ground-truth accuracy, branch point vs this branch, identical to four decimals on every row:

| corpus | n | PPV | TPR | F1 |
|---|---|---|---|---|
| Plohmann malpedia:itw | 57 | 92.9279 | 98.4626 | 95.2799 |
| Bao msvc10-64 (O1+O2) | 34 | 99.1987 | 99.6863 | 99.4397 |
| Bao msvc10-32 (O1+O2) | 34 | 92.2314 | 99.1319 | 95.4390 |
| Bao_Dumped 64 (O1+O2) | 28 | 99.1068 | 97.6134 | 98.3496 |
| Bao_Dumped 32 (O1+O2) | 28 | 91.2758 | 98.8967 | 94.7985 |

Per-fixture speedup against `master`, same benchmark driver run against both trees:

| fixture | primitive calls | wall time |
|---|---|---|
| cutwail | 2,419,558 → 1,513,272 (-37.5%) | 250.9 → 219.7 ms |
| aarch64_static | 2,291,421 → 1,725,333 (-24.7%) | 248.0 → 194.2 ms |
| blockblast | 1,426,062 → 1,048,428 (-26.5%) | 204.0 → 121.1 ms |
| asprox | 1,153,625 → 709,086 (-38.5%) | 98.6 → 83.7 ms |
| njrat | 1,103,337 → 628,933 (-43.0%) | 65.7 → 46.4 ms |
| bashlite | 767,054 → 484,483 (-36.8%) | 144.0 → 82.7 ms |
| komplex | 461,157 → 309,611 (-32.9%) | 48.8 → 38.4 ms |
| **total** | **9,624,022 → 6,420,740 (-33.3%)** | **1060.7 → 786.7 ms (-25.8%)** |

Batch memory, measured on two binaries of 8.1 MB and 7.4 MB at two workers: aggregate peak child RSS **3.39 GiB → 2.55 GiB (-24.8%)** for +0.9% wall time, which is noise. A single 8.1 MB analysis peaks at ~2.5 GB, so an uncapped pool of ten workers can reach ~25 GB.

Gates:

```bash
python -m pytest tests/ -q          # 1143 passed, 1 skipped, 1473 subtests
python -m ruff check .              # clean
python -m ruff format --check .     # 223 files already formatted
python -m ty check src/smda/ fuzzing/ profiling/ .github/workflows/scripts/   # exit 0, no new errors
diff-cover coverage.xml --compare-branch=origin/master --fail-under=100       # 100%, 1137 lines
```

## Notes for review

Two rewrites are equivalences rather than refactors, and both are pinned by tests rather than argued:

- **pclntab scan.** Candidates are located by the literal `\xff\xff\xff` both alternatives share and then confirmed with the *same compiled pattern*, so the accepted set is identical by construction. Two matches of that pattern cannot overlap, so the non-overlapping `finditer` this replaces could not have returned a different count either — which matters because the caller only accepts a table when the scan yields exactly one hit. Checked against the old scan on a real 8.1 MB Go PE, 271 real binaries, and 4,000 randomized buffers.
- **Dalvik shallow decode.** The sweep's decoder performs the same three validity checks in the same order; where an opcode declares a payload the real format decoder still produces the target, gated on the opcode table's own `payload_kind` rather than a hardcoded format list. Equivalence is asserted over **all 247,668 offsets** of the bundled DEX — same accepted offsets, same size and payload, raising on exactly the same offsets — and that check is kept as a test.

`binweight` is serialized inside the report hash, so the report-materialization commit keeps it a float with a bit-identical value; that is deliberate and is why the sum is taken over raw byte lengths rather than halved hex lengths.

Diff coverage is 100%, which required covering several pre-existing branches this work touched: the ELF `.gopclntab` lookup, malformed-header rejection, the 32-bit pclntab path, the orphan decode-attempt cap, and the PLT and `.plt.sec` stub-chain shapes that no bundled fixture exercises.

## Not included

Three candidates from the same profile were deliberately left out, listed here so they are not mistaken for oversights:

- **An interval-based replacement for the byte-granular `code_map` / `ins2fn` indexes.** The surface is wide, and it should not land without an end-to-end CPU win demonstrated against the dict baseline rather than lower memory alone.
- **Switching the `DisassemblyResult` reference containers from sets to tuples.** A tuple iterates in insertion order where a set iterates in hash order, and report identity here is order-sensitive.
- **Releasing the byte maps before report materialization.** This measured only -1.7% peak RSS at 8 MB scale against -16% on the small fixtures, because freeing a dict returns memory to Python's allocator pools rather than to the OS. It also breaks post-analysis introspection through the retained `Disassembler.disassembly` handle, which an existing test relies on.

### DIFF
--- a/batch_analyze.py
+++ b/batch_analyze.py
@@ -94,7 +94,8 @@ if __name__ == "__main__":
 
     os.makedirs(ARGS.output_path, exist_ok=True)
 
-    TOTAL = len(collectInputFiles(ARGS.input_paths))
+    COLLECTED = collectInputFiles(ARGS.input_paths)
+    TOTAL = len(COLLECTED)
     WORKERS = ARGS.workers if ARGS.workers > 0 else getDefaultWorkerCount()
     print(f"Disassembling {TOTAL} file(s) with {WORKERS} worker(s) into {ARGS.output_path}")
     if ARGS.timeout is None:
@@ -107,6 +108,7 @@ if __name__ == "__main__":
         timeout=ARGS.timeout,
         resume=ARGS.resume,
         max_tasks_per_child=ARGS.max_tasks_per_child,
+        collected=COLLECTED,
     )
 
     START = time.time()
--- a/src/smda/DisassemblyResult.py
+++ b/src/smda/DisassemblyResult.py
@@ -282,12 +282,19 @@ class DisassemblyResult:
             verified_refs = sorted(block_starts.intersection(self.code_refs_from[last_ins_addr]))
             if verified_refs:
                 block_refs[block[0][0]] = verified_refs
+        # block_refs was seeded from sorted(block_starts), so it only needs re-sorting if the
+        # exception pass introduced a key that was not already a block start
+        added_key = False
         for block_start, successors in self._getExceptionSuccessors(func_addr, blocks).items():
             merged_successors = set(block_refs.get(block_start, []))
             merged_successors.update(successors)
             block_refs[block_start] = sorted(merged_successors)
             for successor in successors:
-                block_refs.setdefault(successor, [])
+                if successor not in block_refs:
+                    block_refs[successor] = []
+                    added_key = True
+        if not added_key:
+            return block_refs
         return {block_start: block_refs[block_start] for block_start in sorted(block_refs)}
 
     def getInRefs(self, func_addr):
@@ -297,11 +304,9 @@ class DisassemblyResult:
         return sorted(in_refs)
 
     def getOutRefs(self, func_addr):
-        ins_addrs = set()
         code_refs_from = self.code_refs_from
-        for block in self.functions[func_addr]:
-            for ins in block:
-                ins_addrs.add(ins[0])
+        blocks = self.functions[func_addr]
+        ins_addrs = {ins[0] for block in blocks for ins in block}
         # function may be recursive
         ins_addrs.discard(func_addr)
         out_refs = {}
--- a/src/smda/aarch64/AArch64Backend.py
+++ b/src/smda/aarch64/AArch64Backend.py
@@ -1,6 +1,5 @@
 #!/usr/bin/python
 
-import contextlib
 import logging
 import re
 
@@ -53,13 +52,19 @@ _HEX_OPERAND = re.compile(r"0x[0-9a-fA-F]+")
 _IMM_TOKEN = re.compile(r"#?-?0x[0-9a-fA-F]+|#-?\d+")
 
 
-def _immediate_tokens(op_str):
-    values = set()
+def _hasDataRefImmediate(op_str, base_addr, upper_addr, is_in_code_areas):
+    """Whether any printed immediate could name a data address, short-circuiting on the first."""
+    # the token pattern cannot match without one of these two markers
+    if "0x" not in op_str and "#" not in op_str:
+        return False
     for match in _IMM_TOKEN.finditer(op_str):
-        token = match.group().lstrip("#")
-        with contextlib.suppress(ValueError):
-            values.add(int(token, 0))
-    return values
+        try:
+            value = int(match.group().lstrip("#"), 0)
+        except ValueError:
+            continue
+        if base_addr <= value < upper_addr and not is_in_code_areas(value):
+            return True
+    return False
 
 
 class AArch64Backend(ArchBackend):
@@ -268,10 +273,7 @@ class AArch64Backend(ArchBackend):
         binary_info = d.disassembly.binary_info
         base_addr = binary_info.base_addr
         upper_addr = base_addr + binary_info.binary_size
-        if not any(
-            base_addr <= value < upper_addr and not binary_info.isInCodeAreas(value)
-            for value in _immediate_tokens(i_op_str)
-        ):
+        if not _hasDataRefImmediate(i_op_str, base_addr, upper_addr, binary_info.isInCodeAreas):
             # no printed immediate can yield a data ref, so the detail re-decode
             # below would record nothing.
             return
--- a/src/smda/aarch64/AArch64InstructionEscaper.py
+++ b/src/smda/aarch64/AArch64InstructionEscaper.py
@@ -9,6 +9,14 @@ from smda.aarch64.AArch64CapstoneVerification import (
 
 LOGGER = logging.getLogger(__name__)
 
+_KEEP_MASK_MEMO = {}
+# mirrors the _shift_extend class attribute below; additions must be made in both places
+_SHIFT_EXTEND_KEYWORDS = ("lsl", "lsr", "asr", "ror", "sxt", "uxt")
+_SHIFT_EXTEND_IMMEDIATE_RE = re.compile(
+    r"\b(?:lsl|lsr|asr|ror|sxtw|uxtw|sxtb|uxtb|sxth|uxth|sxtx|uxtx)\s*#-?(?:0x[0-9a-fA-F]+|\d+)",
+    re.IGNORECASE,
+)
+
 
 class AArch64InstructionEscaper:
     _aritlog_group = {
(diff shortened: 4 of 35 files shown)


danielplohmann:
Merged, thanks. Master is now at 9e341ac.

This one needed a different rebase from #227 and #228.
Those two were sitting on the amended #226, so the duplicate commits dropped out as already applied; this branch was cut before that amendment, so a plain rebase hit conflicts inside the #226 commits themselves.
Replaying only the 12 commits after `4c4ad108` with `git rebase --onto origin/master 4c4ad108` was the cleaner path, and 11 of them applied untouched.

The twelfth, `perf(loaders)`, conflicted on one import line in `MachoFileLoader.py`: master had dropped `import warnings` when the `catch_warnings` suppression moved into `pyproject.toml`, and your commit dropped `from functools import lru_cache` along with the four decorators.
Both lines go, since neither name is used any more.
That is the only place I touched your work.

## What I verified

The output-identity claim holds against today's master, which is the version that matters now that #226 through #228 have landed.
Ten fixtures across all four backends, default config plus strings and the ApiScout database: function, block and instruction counts, the per-function block/instruction map, and `computeReportIdentityHash` are identical on every one, and wall time drops 6.20s to 5.09s, or -17.9%.
Per fixture the wins run from -5% on `aarch64_static` to -52% on `cutwail`, which is the shape a profile-driven pass should have.

Worth recording for anyone repeating this: measured before the rebase, the raw hashes differ on every fixture, because the v4.4.5 release commit bumped `smda_version` and that field is serialized into the report and therefore into the hash.
Excluding that one field made them identical there too.

`make test` runs 1147 passed, 2 skipped, 1477 subtests, and `ruff check` plus `ruff format --check` are clean on 233 files.
The single `test_macho_get_bitness_never_crashes` failure this branch carried was the `safe_lief_parse` gap, which the rebase picks up from #227.

## Reading notes

The two rewrites that are equivalences rather than refactors both hold up.
The pclntab anchor checks `position - 1` and `position` for every `\xff\xff\xff` run, which is exactly the little-endian and big-endian spelling of the pattern, and the dedup through `sorted(set(...))` is needed for the overlapping-`0xff` case rather than decorative.
The intel dispatch table raising at import when the mnemonic tables stop being disjoint is the right place to catch a future edit, since nothing else would notice.

`format_operand` returning early on `None` is equivalent: it used to fall through every isinstance arm to reach the same empty string, seven of them through `ABCMeta`.
Dropping eight LIEF-keyed `lru_cache` decorators cannot change results for functions that sort section lists, and a cache hit costing more than the work is the sort of thing only a profile finds.

## Three follow-ups we will take ourselves

None of these is a request; recording them so they are not lost.

`_analyzeCondJmpInstruction` and `_analyzeLoopInstruction` now parse the operand with `int(i_op_str, 16)` where `getReferencedAddr` returned 0 for an operand carrying no hex literal, and the surrounding `if jump_destination:` then skipped it.
For Jcc and loop capstone emits rel8/rel32 as a bare `0x...`, so this should be unreachable, but the argument deserves to sit next to the code: if it is ever reached the report degrades to `status="error"`, and the fuzzing oracle will not flag it, because `ValueError` is on the allowlist.

`getMemoryBudget` reads `SC_PHYS_PAGES`, which reports the host rather than a cgroup limit.
Under Docker or Kubernetes, which is how MCRIT workers usually run, the budget can be several times the memory actually available: measured here, 22.9 GiB of RAM gives an 11.4 GiB budget, and a 4 GiB container would compute the same number.
Reading `/sys/fs/cgroup/memory.max`, falling back to the v1 path and to `SC_PHYS_PAGES`, closes it.
`LARGE_INPUT_RSS_FACTOR` and `MEMORY_BUDGET_FRACTION` also belong on `SmdaConfig` beside `MAX_FUNCTION_CANDIDATES` and `MAX_IMAGE_SIZE`, so an operator can move them without editing the module.

The comment in `_runPartitions` says ordering by path makes a rerun yield summaries identically, but it only orders within one `wait()` batch, so ordering across batches still follows completion.
The behaviour is fine, since the only consumer aggregates counts and every summary carries its path, and buffering to the end would cost the streaming progress the batch runner exists to give.
The comment is what wants softening.

Issue #229, opened 2026-08-10
PeSymbolProvider recovers no names from PE COFF symbol tables (Symbol.section is always None; use section_idx)


danielplohmann:
## Summary

`PeSymbolProvider.parseSymbols()` recovers **zero** function names from a PE that has a full COFF symbol table, because it skips every symbol at

```python
if symbol.section is None:
    continue
```

and LIEF returns `Symbol.section is None` for **all** PE symbols. The needed information is still available as `Symbol.section_idx`.

## Reproduction

Test binary: a Rust program (`regex` + `flate2`) built for `x86_64-pc-windows-gnu` with `debuginfo=2`, linked by mingw-w64, so the PE carries a complete COFF symtab.

```python
import lief
b = lief.parse("probe.exe")
syms = list(b.symbols)
print(len(syms))                                                    # 16019
print(sum(1 for s in syms if s.section is None))                    # 16019  (100%)
print(sum(1 for s in syms if s.complex_type.name == "FUNCTION"))    # 5788
```

Consequently:

```python
from smda.Disassembler import Disassembler
r = Disassembler().disassembleFile("probe.exe")
named = [f for f in r.getFunctions()
         if f.function_name and not f.function_name.startswith("sub_")]
print(len(list(r.getFunctions())), len(named))    # 5111 functions, 1 named
```

The single name is `original_entry_point`, set by `setBaseAddress`/OEP handling — not by `parseSymbols`.

## Not a LIEF 1.0 regression

I initially assumed this was new in LIEF 1.0, and that was wrong. Checked both:

| LIEF | symbols | `section is None` | `COMPLEX_TYPE.FUNCTION` | valid `section_idx` |
|---|---|---|---|---|
| 0.17.6 | 16019 | 16019 (100%) | 5788 | **5788** |
| 1.0.0 | 16019 | 16019 (100%) | 5788 | **5788** |

So the behaviour is the same on both sides of the 1.0 boundary, and `section_idx` is populated correctly on both. SMDA declares `lief>=0.16.0`, so this affects the whole supported range.

## Suggested fix

Resolve the section through the 1-based `Symbol.section_idx` instead of `Symbol.section`:

```python
def parseSymbols(self, lief_binary, base_addr=None):
    active_base = self._resolve_base_addr(lief_binary, base_addr)
    sections = list(lief_binary.sections)
    function_symbols = {}
    for symbol in lief_binary.symbols:
[...]

PR #233, opened 2026-08-11
fix: recover function starts lost to padding, alignment and jump-table heuristics

r0ny123:
## Summary

Seven defects in function-start recovery, each one a heuristic discarding a start the binary itself identifies. They were found by attributing the false negatives and false positives of the accuracy corpora one class at a time, and every change is gated on the same rule: **true-positive rate may not fall on any configuration.** It never does — and no single binary in any configuration loses a true positive.

Macro-averaged over the all-samples population (`Os`/`Od`/`Ox` included), against `master` at `0ac4a55`:

| dataset | n | master | this branch | ΔF1 | master TPR | branch TPR | ΔTP | ΔFP |
|---|---|---|---|---|---|---|---|---|
| ByteWeight MSVC x86 | 68 | 93.3326 | **94.7409** | **+1.4083** | 97.6271 | 97.8613 | +223 | −5,585 |
| ByteWeight MSVC x86, dumped | 56 | 92.3140 | **94.0371** | **+1.7231** | 97.2125 | 97.4969 | +223 | −5,608 |
| malpedia ITW | 57 | 95.2799 | **95.4551** | **+0.1752** | 98.4626 | 98.5302 | +14 | −58 |
| ByteWeight MSVC x64 | 68 | 99.3420 | **99.4542** | **+0.1122** | 99.7294 | 99.8352 | +159 | −309 |
| ByteWeight MSVC x64, dumped | 56 | 98.2983 | **99.3740** | **+1.0757** | 97.7518 | 99.8117 | +1,987 | −184 |
| **mean of five** | | 95.7134 | **96.6123** | **+0.8989** | | | | |

Precision and recall both rise on every configuration.

## The seven commits

| commit | change |
|---|---|
| `4fee6a1` | `fix(intel)` stop seeding the body behind an MSVC hot-patch pad |
| `611697b` | `fix(intel)` scan an explicit jump table whose bound was not recovered |
| `69b813f` | `feat(intel)` recover x64 exception records from a headerless image |
| `9aced4b` | `fix(intel)` stop skipping a hot-patch pad that is not alignment filler |
| `2bc0a97` | `fix(core)` let an exception record outrank the inferred alignment floor |
| `1923b48` | `fix(intel)` recognize the x64 entry pads the gap scan skipped as filler |
| `cf9c09c` | `fix(intel)` claim a relative jump table's bytes as data |

## Root causes

**`mov edi, edi` is both alignment filler and a function's first instruction, and the two were never separated.** `DEFAULT_PROLOGUES` holds the padded forms *and* the bare `55 8b ec`, so the prologue scan seeded two candidates per hot-patched function — the pad, which is the real entry, and the body two bytes behind it (`4fee6a1`). The gap scan had the mirror-image problem: it skipped the pad as an effective NOP unless the five-byte window matched exactly `8b ff 55 8b ec`, so a pad in front of any other prologue mislocated the start (`9aced4b`, and `1923b48` for the x64 encodings `66 90` and `8b c9`).

The separator is alignment, and on this corpus it is exact. Filler exists only to reach the next 16-byte boundary, so a pad that neither ends on one nor is followed by more padding is the entry itself: 221 of 223 pads that ground truth names as starts end off the boundary, and all 14 sites where the pad really is filler end on it. `isUnalignedEntryPad` (`src/smda/intel/FunctionCandidateManager.py:105`) is deliberately narrow — the sets are per bitness, because widening x86 beyond `mov edi, edi` was measured to cost true positives: in real-world 32-bit images the two bytes before a genuine start are an effective NOP by coincidence often enough that the ITW set has 318 such sites, spread across every alignment including odd addresses.

**A sentinel mismatch abandoned unresolved switch tables.** `_findJumpTableSize` reports "bound not recovered" as `0`, but `_resolveExplicitTable` (`src/smda/intel/JumpTableAnalyzer.py:180`) defaulted only on `None`, so `0` meant `range(0)` and the table was never read. `/Od` builds hit this constantly, since spilling the switch value produces a memory-operand compare the size regex deliberately does not match. With no targets the case bodies never enter `code_map`, each becomes a gap, and the gap scan books each one as a one-block function — a cascade worth 3,565 false positives on the x86 set for this commit alone (`611697b`).

**A relative jump table's own bytes were never claimed as data.** `_resolveExplicitTable` marks every entry it consumes; `_extractRelativeTableOffsets` (`src/smda/intel/JumpTableAnalyzer.py:133`) never did — its `addDataRef` call had been commented out and the function did not take `state`. On x64, where switch tables are int32 offsets, the table bytes stayed unclaimed and the gap scan seeded a function at any entry pair that happened to look like code. Measured on one binary, 40 of its 161 false positives sit inside a table the analyzer had already walked, and none of its 5,861 true positives do (`cf9c09c`).

**Headerless x64 dumps lost `.pdata` entirely.** `locateExceptionHandlerCandidates` needs the section table, and a dumped image has no parseable header, so every guaranteed function start went missing — 4,369 of them in one sample. `_carveExceptionRecords` (`src/smda/intel/FunctionCandidateManager.py:639`) finds the table without one: `RUNTIME_FUNCTION` is three DWORD-aligned image-relative ULONGs, the table is sorted and non-overlapping, and `UNWIND_INFO` byte 0 packs Version with Flags such that only eight byte values are legal there. The carve runs only when `getSections()` yields nothing at all, so every file-loaded binary keeps exactly its current behaviour (`69b813f`).

**An inferred alignment floor outranked the image's own declaration.** `USE_ALIGNMENT` derives a per-binary floor from where call-referenced candidates happen to sit, and `getNextFunctionStartCandidate` (`src/smda/common/FunctionCandidateManager.py:150`) then skipped every candidate below it — including candidates created from `.pdata`. Fifty-nine ground-truth starts scored 5040 and were never handed to analysis at all. Exception records are now exempt; symbols are declarations of the same kind but are deliberately **not** exempt, because exempting them was measured to cost three true positives on the x86 set and one on the ITW set (`2bc0a97`).

## Validation

`.pdata` carving was not accepted on a corpus delta. Every dumped sample has an on-disk twin whose real `.pdata` parses from its section table, so the carved table was diffed against exact ground truth for the structure itself: **56/56 tables recovered, 103,815/103,815 entries, zero invented.** Only then was accuracy measured.

The same parse also confirms the pre-existing `UNW_FLAG_CHAININFO` skip is sound: of 30,592 chained entries across the x64 set, **none** is a ground-truth function start, while 59 primary entries were being lost.

- Full suite green (1,176 passed, 1 skipped), `ruff check`, `ruff format --check`, and `ty` clean with zero errors.
- Diff coverage 100% on the changed lines. Twenty-five regression tests were added; several were verified to fail on the pre-fix code.
- Benchmark fixtures: primitive calls +0.00% to +0.13%, and **one** of eight `report_hash` values moves (`cutwail`, a 32-bit PE where the hot-patch fix fires). The other seven are unchanged.
- Recall guard on 264 real malware samples: the recovered-function count falls 276,793 → 275,661, which is what a precision fix looks like here — 890 of the dropped addresses have the pad recovered two bytes away, and switch-case pseudo-functions collapse into their owning function. Classified by instruction coverage instead, the 88 changed samples gain 13,326 addresses and lose 3,105, **net +10,221 more code disassembled**.

## Notes for review

Three levers were confirmed and then rejected rather than shipped, because each drops recall somewhere: enabling `RESOLVE_TAILCALLS` (+87 true positives for +564 false positives), promoting `jmp` targets by entry shape (best rule: 128 recovered for 909 new false positives), and an entry-shape score gate on gap candidates (removes 1.2 true positives per false positive). The two "obvious" widenings of the changes above — symbols in the alignment exemption, and the other effective-NOP encodings on x86 — are also measured rejections, and are marked as such in the source so they are not re-attempted.

The MSVC EH funclet question is deliberately untouched. The two ground truths disagree about it on byte-identical code: PDB-derived ByteWeight folds a funclet into its parent, IDA labels it a function. Suppressing that shape gains roughly 1.5 macro F1 on the x86 sets and costs hundreds of true positives on the ITW set, so it is a labeling-convention decision rather than a defect.

### DIFF
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -152,7 +152,15 @@ class FunctionCandidateManager:
             if not (candidate.isFinished() or candidate.getScore() == 0):
                 if self.language_candidates_only and candidate.lang_spec is None:
                     continue
-                if self.identified_alignment and candidate.alignment < self.identified_alignment:
+                if (
+                    self.identified_alignment
+                    and candidate.alignment < self.identified_alignment
+                    and not candidate.is_exception_handler
+                ):
+                    # The alignment floor is inferred from where call-referenced candidates
+                    # happen to sit, so it only ever bounds a guess, and an exception record is
+                    # the image's own declaration of a function start. Symbols are declarations
+                    # too but are deliberately NOT exempt: measured, that costs true positives.
                     continue
                 yield candidate
 
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -157,10 +157,9 @@ class FunctionCandidateManager:
                     and candidate.alignment < self.identified_alignment
                     and not candidate.is_exception_handler
                 ):
-                    # The alignment floor is inferred from where call-referenced candidates
-                    # happen to sit, so it only ever bounds a guess, and an exception record is
-                    # the image's own declaration of a function start. Symbols are declarations
-                    # too but are deliberately NOT exempt: measured, that costs true positives.
+                    # The floor is inferred from call-referenced candidates, so it only bounds
+                    # a guess; an exception record is the image's own declaration. Exempting
+                    # symbols as well was measured to cost true positives -- do not widen this.
                     continue
                 yield candidate
 
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -49,6 +49,17 @@ _RE_PLTSEC_BLOCK = re.compile(
 )
 _RE_PLTSEC_ENTRY = re.compile(b"\xf3\x0f\x1e\xfa\xf2\xff\x25(?P<function>.{4})", re.DOTALL)
 
+_PDATA_ENTRY_SIZE = 12
+_PDATA_MIN_ENTRIES = 16
+_PDATA_SEED_ENTRIES = 4
+# A sample only finds a table if it lands inside one with a whole seed window left, so this
+# must stay <= (_PDATA_MIN_ENTRIES - _PDATA_SEED_ENTRIES) * _PDATA_ENTRY_SIZE.
+_PDATA_SAMPLE_STRIDE = 128
+_PDATA_MAX_FUNCTION_SIZE = 0x100000
+# UNWIND_INFO byte 0 packs Version (bits 0-2, always 1) and Flags (bits 3-7), and only
+# EHANDLER/UHANDLER/CHAININFO are defined -- so just these eight byte values are legal.
+_UNWIND_INFO_FIRST_BYTES = frozenset(0x01 | (flags << 3) for flags in range(8))
+
 
 class FunctionCandidateManager(_CommonFunctionCandidateManager):
     CANDIDATE_CLASS = FunctionCandidate
@@ -483,7 +494,9 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
             self.pdata_end_addresses = set()
             record_pdata_ends = self.config.USE_PE_X64_PDATA_ENDS
             base_addr = self.disassembly.binary_info.base_addr
+            has_sections = False
             for section_info in self.disassembly.binary_info.getSections():
+                has_sections = True
                 section_name, section_va_start, section_va_end = section_info
                 if section_name == ".pdata":
                     rva_start = section_va_start - self.disassembly.binary_info.base_addr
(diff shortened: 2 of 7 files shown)


danielplohmann:
Merged, thanks. Master is now at e394a45.

Rebased onto master rather than pushed through the button, as with the three before it; the seven commits replayed with no conflicts.
The suite runs 1180 passed, 2 skipped, 1480 subtests, and `ruff check` plus `ruff format --check` are clean on 234 files.

The accuracy tables were produced at 0ac4a55, one commit short of c8b9950, which fixed PE COFF symbol recovery (#229) and therefore seeds candidates on exactly the MSVC-style binaries these heuristics target.
I have not asked for a re-measurement, since the table is documentation and the direction of the result is not in doubt.
Worth knowing when the numbers are next quoted, though.

## What I verified, and where it differs from the description

The two hunks worth challenging are both pinned by tests rather than argued, which is the right way round.
I checked the reconstructed data-ref address by hand: the read is at `getRawBytes(rebased + index * 4)` with `rebased = off_jumptable + bonus_offset - base_addr`, so `off_jumptable + bonus_offset + index * 4` is exactly the VA that was read, bonus case included.
The `.pdata` carve is gated on `getSections()` returning nothing at all, so a file-loaded binary keeps its behaviour by construction, and the sample stride of 128 respects the bound its own comment states.

The output movement is narrower than "one of eight report_hash values moves", in two directions.

On cutwail, under the benchmark's own path (`disassembleBuffer` at base 0x4000000), all 504 function offsets are identical and so are the function, block and instruction counts.
What moves is `statistics.num_failed_functions`, 20 to 22, and nothing else in the report.
That is most likely the jump-table bytes no longer decoding as code, which is the intended effect, but it reads better stated than left as a moved hash.

On the unmapped-buffer path nine of ten fixtures are byte-identical, and the tenth is `tests/rust_pe_gnu_xored`, which did not exist when this PR was written.
Its CFG does not move at all: 2354 functions, 30124 blocks, 154207 instructions and the same per-function shape.
What moves is the data claim, `xdata_refs_from` 232 to 450 keys and `xdata_refs_to` 206 to 3388, plus one function's `stringrefs` - a gimli DWARF parser whose newly claimed table bytes now resolve as data references.
That fixture is a mingw-linked Rust PE, so it exercises the hot-patch and jump-table paths directly; it is worth having in the "what moves" accounting from here on.

## One thing to keep an eye on

`jumptable_size is not None` becoming a truthiness test is the change whose reach exceeds its diff: a table whose bound genuinely resolves to zero now scans up to 255 entries rather than none, bounded only by the per-entry image check.
The measured 3,565 false positives it removes on the x86 set pays for that comfortably, and the sentinel argument is correct.
It is still the change most likely to behave differently on a sample that looks unlike the accuracy corpora, so it is the first place I would look if a switch-heavy binary starts over-recovering.

PR #234, opened 2026-08-11
fix(core): plant a block that overflows its section instead of dropping it

### LINKED ISSUE #232: Synthesis round-trip loses two instructions on the komplex fixture (ELF and Mach-O)
Synthesizing the `komplex` fixture to ELF or Mach-O and re-parsing the result loses 2 of its 4914 recovered instructions.
PE round-trips intact, and the other fixtures round-trip intact in all three formats.

This predates #226, #227 and #228 and reproduces on master, so it is not fallout from the synthesis work in #228.
Filing it separately rather than stacking a fourth PR on that branch.

## Reproducing

```python
import lief
from pathlib import Path

raw = Path("tests/komplex_xored").read_bytes()

for output_format in ("pe", "elf", "macho"):
    parsed = lief.parse(list(bytes(report.synthesizeBinary(output_format=output_format))))
    intact = corrupt = 0
    for function in report.getFunctions():
        for block in function.getBlocks():
            for instruction in block.getInstructions():
                expected = bytes.fromhex(instruction.bytes)
                try:
                    got = bytes(parsed.get_content_from_virtual_address(instruction.offset, len(expected)))
                except Exception:
                    got = b""
                intact, corrupt = (intact + 1, corrupt) if got == expected else (intact, corrupt + 1)
    print(output_format, intact, corrupt)
```

## Measured

| Fixture | PE | ELF | Mach-O |
|---|---|---|---|
| `komplex` | 4914 / 4914 | 4912 / 4914 | 4912 / 4914 |
| `cutwail` | 1611 / 1611 | 1611 / 1611 | 1611 / 1611 |
| `asprox` | 15714 / 15714 | 15714 / 15714 | 15714 / 15714 |

`komplex` is the Mach-O fixture based at `0x100000000`, which is also the one that exposed the segment base handling in #218, so the two are plausibly neighbours.
The remaining two instructions are worth identifying before assuming which side is wrong: the planting could be dropping them, or the section they fall in could be described in a way that stops `get_content_from_virtual_address` reading them back.

Both formats losing exactly the same count points at the shared layout code in `BinarySynthesizer` rather than at either synthesizer.

r0ny123:
Fixes #232.

## Which side was wrong

Neither, as it turns out — the report describes the layout correctly and the planting logic is what drops the bytes.

The two instructions are the last block of the function at `0x100006db0`: a `call` at `0x100006ee0` and the `nop` it falls through to at `0x100006ee5`. `komplex`'s `__text` ends at `0x100006ee5` and `__stubs` begins at `0x100006ee6`, so the call sits **entirely inside** `__text` while the nop lands in the single byte of padding between the two sections. Checked against the fixture's own Mach-O headers with lief, the report's section list matches exactly — `__text` `0x1000012b0`+`0x5c35`, `__stubs` at `0x100006ee6` — so nothing is misdescribed. Recovery simply followed a fall-through one byte past the end of a section, and that byte is still inside the mapped `__TEXT` segment.

What loses them is that planting was all-or-nothing per block:

```python
if section["executable"] and section["va_start"] <= block_offset and block_offset + len(chunk) <= section["va_end"]:
```

The block is both instructions, six bytes, `0x100006ee0`–`0x100006ee6`. One byte past `__text` fails the test, so the whole block is skipped — taking the five-byte call that fits perfectly with it. The `9090909090` the reproducer reads back at `0x100006ee0` is the synthesizer's own NOP fill, never overwritten.

Your instinct about the shared layout code was right, with a wrinkle: the loop is not in `BinarySynthesizer`, it is **byte-identical copies** of it in `ElfSynthesizer` and `MachoSynthesizer`, which is exactly why both formats lose the same two. PE never had the problem because it ignores the report's section list and builds one `.text` spanning `min_rva` to `align_up(max_rva)` across every function extent.

## The change

Blocks are now planted **per overlapping range** rather than whole, so bytes that fit are written even when the block crosses a boundary. A section whose block runs past its end **grows to cover it**, capped at the following section's start, so growth only ever claims address space no other section describes — for `komplex` that is exactly the one padding byte between `__text` and `__stubs`. Bytes that still land nowhere are reported with a count rather than the block vanishing.

The shared version lives in `BinarySynthesizer` and replaces both copies. PE is untouched.

## Validation

Every bundled fixture round-tripped through all three formats, before and after:

| fixture | format | before | after |
|---|---|---|---|
| `komplex` | ELF | 4912 / 4914 | **4914 / 4914** |
| `komplex` | Mach-O | 4912 / 4914 | **4914 / 4914** |
| `komplex` | PE | 4914 / 4914 | 4914 / 4914 |

**Every other fixture is byte-identical before and after** — `cutwail`, `bashlite`, `mirai_i386`, `mirai_x64`, `aarch64_static`, `aarch64_switch_macho_O0`, `pe_export_label_test`, `rust_pe_gnu`. The ELF synthesizer now emits no warning at all for `komplex`, where it previously reported `block 0x100006ee0 of function 0x100006db0 fits no executable section, skipped`.

Full suite green (1,187 passed, 1 skipped), `ruff check`, `ruff format --check` and `ty` clean with zero errors, diff coverage 100%. Benchmark fixtures unchanged to 0.000% of primitive calls with no `report_hash` moving, as expected — synthesis is not on the disassembly path.

Six regression tests, all verified to fail on `master`: the overflow growth, growth capped at a neighbouring section with the fitting bytes still planted, growth taking only the overflow rather than the whole gap, a block starting in a gap being reported, a block below its section being clipped rather than wrapping to the end of the buffer, and the `komplex` end-to-end round-trip in all three formats.

## Unrelated observations from the sweep

Worth separate issues rather than folding in here, all pre-existing and untouched by this change:

- PE synthesis reads back fewer instructions than ELF/Mach-O on several fixtures — `aarch64_static` 18963/19881, `bashlite` 7733/8767, `mirai_i386` 13365/14758, `mirai_x64` 11762/12564 — while ELF and Mach-O are 100% on all four.
- `rust_pe_gnu` round-trips fully through PE and ELF but reads back 0/154207 through Mach-O.
- `synthesizeBinary` raises `ValueError` for reports whose architecture did not resolve (the non-x86 `mirai_*` fixtures, `asprox`).

Happy to file those separately if useful.

### DIFF
--- a/src/smda/synthesis/BinarySynthesizer.py
+++ b/src/smda/synthesis/BinarySynthesizer.py
@@ -81,6 +81,58 @@ class BinarySynthesizer:
         extent_end = max(self._functionExtentEnd(self.report.xcfg[offset]) for offset in offsets)
         return va_start, align_up(max(extent_end, max(offsets) + 1), end_alignment)
 
+    def _growSectionForOverflow(self, sections, section, block_end):
+        """Extends an executable section's end towards block_end, stopping at the next section.
+
+        Recovered code can run past the end of the section it starts in: a call at the tail of
+        __text whose fall-through lands in the padding before __stubs is still part of the
+        function, and the byte is still inside the mapped image, just not inside any section.
+        Growth is capped at the following section's start, so it only ever claims address space
+        no other section describes.
+        """
+        successors = [other["va_start"] for other in sections if other["va_start"] > section["va_start"]]
+        limit = min(successors) if successors else block_end
+        new_end = min(block_end, limit)
+        if new_end <= section["va_end"]:
+            return
+        section["raw"].extend(self._nopFill(new_end - section["va_end"]))
+        section["va_end"] = new_end
+
+    def _plantFunctionChunks(self, sections, offsets):
+        """Plants every block of every function into the executable sections covering it.
+
+        Blocks are written per overlapping range rather than all-or-nothing: a block that runs
+        one byte past its section would otherwise be dropped whole, taking the instructions that
+        do fit with it.
+        """
+        ordered = sorted(sections, key=lambda section: section["va_start"])
+        executable = [section for section in ordered if section["executable"] and section["raw"] is not None]
+        for offset in offsets:
+            for block_offset, chunk in self._iterFunctionChunks(self.report.xcfg[offset]):
+                block_end = block_offset + len(chunk)
+                for section in executable:
+                    if section["va_start"] <= block_offset < section["va_end"] < block_end:
+                        self._growSectionForOverflow(ordered, section, block_end)
+                        break
+                planted = 0
+                for section in executable:
+                    start = max(block_offset, section["va_start"])
+                    end = min(block_end, section["va_end"])
+                    if start >= end:
+                        continue
+                    section["raw"][start - section["va_start"] : end - section["va_start"]] = chunk[
+                        start - block_offset : end - block_offset
+                    ]
+                    planted += end - start
+                if planted < len(chunk):
+                    self._warn(
+                        "block 0x%x of function 0x%x: %d of %d bytes fit no executable section",
+                        block_offset,
+                        offset,
+                        len(chunk) - planted,
+                        len(chunk),
+                    )
+
     def _nopPadding(self):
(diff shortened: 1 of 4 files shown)


danielplohmann:

PR #237, opened 2026-08-11
fix(core): read the stored header only when it is this format's own

### LINKED ISSUE #235: Mach-O synthesis emits a 32-bit image for a 64-bit report when the source was not a Mach-O
Synthesizing a 64-bit report to Mach-O emits a **32-bit** Mach-O whenever the report came from something that is not a Mach-O, so every address above 4 GiB is truncated mod 2^32 and none of the planted code can be read back.

`rust_pe_gnu` is based at `0x140000000`, so all 154207 of its recovered instructions are lost this way. PE and ELF round-trip it intact.

## Reproducing

```python
import struct, lief
from pathlib import Path

raw = Path("tests/rust_pe_gnu_xored").read_bytes()
print("report bitness", report.bitness, "base 0x%x" % report.base_addr)

blob = bytes(report.synthesizeBinary(output_format="macho"))
print("emitted magic 0x%08x" % struct.unpack("<I", blob[:4])[0])   # 0xfeedface, the 32-bit magic
print([hex(s.virtual_address) for s in lief.parse(list(blob)).sections][:3])
```

```
report bitness 64 base 0x140000000
emitted magic 0xfeedface
['0x40001000', '0x4009b000', '0x4009c000']
```

`0x140001000 & 0xFFFFFFFF == 0x40001000`, so the section addresses are the real ones with the top 32 bits cut off.

## Cause

`MachoSynthesizer._is64` (`src/smda/synthesis/MachoSynthesizer.py:101`) reads the stored `xheader` as a Mach-O header without first checking that it is one:

```python
def _is64(self):
    if self._hasHeader(4):
        magic = struct.unpack("<I", self.report.xheader[0:4])[0]
        return magic in (MH_MAGIC_64, MH_CIGAM_64)
    return self.report.bitness != 32
```

`_hasHeader` (`src/smda/synthesis/BinarySynthesizer.py:109`) only tests that an `xheader` exists and is long enough — it says nothing about the format. For a PE-origin report the first four bytes are `4d 5a 90 00`, which is not `MH_MAGIC_64`, so the method returns `False` and the correct `report.bitness` fallback on the last line is never reached.

`_getCpuType` (`src/smda/synthesis/MachoSynthesizer.py:107`) has the same shape and reads `xheader[4:8]` as a Mach-O `cputype`. For `rust_pe_gnu` those bytes are `03 00 00 00`, so the synthesized Mach-O also claims `CPU_TYPE_I386` for a 64-bit x86-64 report.

Both are only reached when an `xheader` is present but belongs to another format, which is why the Mach-O fixtures are unaffected: `komplex` supplies a real `MH_MAGIC_64` and takes the intended path.

## Measured

| fixture | origin | base | PE | ELF | Mach-O |
|---|---|---|---|---|---|
| `rust_pe_gnu` | PE | `0x140000000` | 154207 / 154207 | 154207 / 154207 | **0 / 154207** |
| `komplex` | Mach-O | `0x100000000` | 4914 / 4914 | 4914 / 4914 | 4914 / 4914 |
| `cutwail` | PE | `0x400000` | 1611 / 1611 | 1611 / 1611 | 1611 / 1611 |

`cutwail` hides it: it is also a PE-origin report that synthesizes to a 32-bit Mach-O, but its base is below 4 GiB so nothing truncates. Only a report based above 4 GiB whose `xheader` is not a Mach-O exposes the bug, which is why `rust_pe_gnu` is the only fixture that fails.

Measured with the round-trip loop from #232 against `master` at `5ad5504`.

## Suggested direction

Gate both methods on the `xheader` actually being a Mach-O — the file already has `MH_MAGIC`/`MH_CIGAM`/`MH_MAGIC_64`/`MH_CIGAM_64` and a `_parseSegments` path that validates the magic before trusting the rest — and otherwise fall through to `report.bitness` and the architecture. Happy to put a PR together if the direction looks right.

r0ny123:
Fixes #235.

## Cause

A report keeps the header of the binary it came from, whatever format that was, and `_hasHeader()` only checks that one exists and is long enough. The Mach-O and ELF synthesizers read their own header fields out of it regardless, so a report from another format decodes whatever those bytes happen to hold.

For Mach-O this is not cosmetic. `_is64()` reads bytes 0:4 as a Mach-O magic; a PE report's `MZ` header is not `MH_MAGIC_64`, so it answers `False` and the correct `report.bitness` fallback on the next line is never reached. A 64-bit report is then written as a 32-bit Mach-O and every address is truncated mod 2³². `rust_pe_gnu` is based at `0x140000000`, so all 154207 of its instructions land at `0x40000000` and none can be read back. `_getCpuType()` has the same shape and stamped `CPU_TYPE_I386`, taken from bytes 4:8 of the DOS header, onto an x86-64 report.

## The ELF sibling, found while sweeping the class

The same defect is present in `ElfSynthesizer`, milder because it does not corrupt addresses: `e_ident[7:16]` was copied from a foreign header and `e_type` read from its bytes 16:18. Measured on `master`, a PE-origin report produces `e_type = 0xb8` and a Mach-O-origin one `e_type = 0x17` — neither is a defined ELF type — with a non-zero `EI_PAD` that the spec requires to be zero. Both are fixed here, since it is one bug class rather than two.

`PeSynthesizer` needs no change: it parses the stored header with lief first and falls back to a minimal header when that fails, which is the behaviour the other two were missing. `MachoSynthesizer._parseSegments` already validated the magic the same way — the four accessor methods simply did not.

## Measured

| fixture | origin | base | Mach-O before | Mach-O after |
|---|---|---|---|---|
| `rust_pe_gnu` | PE | `0x140000000` | **0 / 154207** | **154207 / 154207** |
| `cutwail` | PE | `0x4000000` | 1611 / 1611 | 1611 / 1611 |
| `komplex` | Mach-O | `0x100000000` | 4912 / 4914 | 4912 / 4914 |

Round-trip loop from #232, against `master` at `5ad5504`. Every bundled fixture was swept in all three formats and `rust_pe_gnu`'s Mach-O column is the only one that moves.

`cutwail` shows the fallback behaving: it is also PE-origin, and it correctly stays a 32-bit Mach-O because its report says `bitness = 32`. `komplex` is unchanged at 4912/4914 because that is the separate #232 defect, fixed in #234 — the two PRs are independent and touch different files.

## Validation

Full suite green (1,188 passed, 1 skipped), `ruff check`, `ruff format --check` and `ty` clean with zero errors, diff coverage 100%.

Seven regression tests, six of which fail on `master`: the emitted magic for a PE-origin 64-bit report, the CPU type not coming from DOS-header bytes, the round-trip above 4 GiB, a real Mach-O header still selecting the word size in both directions, a foreign header falling back to `report.bitness` in both directions, the ELF identity not inheriting foreign bytes, and a real ELF header still supplying it.

### DIFF
--- a/src/smda/synthesis/ElfSynthesizer.py
+++ b/src/smda/synthesis/ElfSynthesizer.py
@@ -95,8 +95,17 @@ class ElfSynthesizer(BinarySynthesizer):
     def _getBitness(self):
         return 64 if self.report.bitness == 64 else 32
 
+    def _hasElfHeader(self, min_length):
+        """True only when the stored header is long enough AND is itself an ELF header.
+
+        A report keeps the header of the binary it came from, whatever format that was, so
+        reading these fields off a PE or Mach-O header yields whatever those bytes happen to
+        hold -- an e_type of 0xb8, say, which is not a defined ELF type at all.
+        """
+        return self._hasHeader(min_length) and bytes(self.report.xheader[0:4]) == b"\x7fELF"
+
     def _getMachine(self):
-        if self._hasHeader(0x14):
+        if self._hasElfHeader(0x14):
             machine = struct.unpack("<H", self.report.xheader[0x12:0x14])[0]
             if machine:
                 return machine
@@ -380,9 +389,9 @@ class ElfSynthesizer(BinarySynthesizer):
         ehdr[4] = 2 if is_64 else 1
         ehdr[5] = 1
         ehdr[6] = 1
-        if self._hasHeader(16):
+        if self._hasElfHeader(16):
             ehdr[7:16] = self.report.xheader[7:16]
-        e_type = struct.unpack("<H", self.report.xheader[16:18])[0] if self._hasHeader(18) else ET_EXEC
+        e_type = struct.unpack("<H", self.report.xheader[16:18])[0] if self._hasElfHeader(18) else ET_EXEC
         machine = self._getMachine()
         entry = 0
         if self.report.oep:
--- a/src/smda/synthesis/MachoSynthesizer.py
+++ b/src/smda/synthesis/MachoSynthesizer.py
@@ -98,14 +98,25 @@ class MachoSynthesizer(BinarySynthesizer):
             self._warn("Mach-O synthesis from sections failed (%s), falling back to minimal layout", exc)
             return self._synthesizeMinimal(offsets)
 
+    def _hasMachoHeader(self, min_length):
+        """True only when the stored header is long enough AND is itself a Mach-O header.
+
+        A report keeps the header of the binary it came from, whatever format that was, so
+        reading these fields off a PE or ELF header yields whatever those bytes happen to hold.
+        """
+        if not self._hasHeader(min_length):
+            return False
+        magic = struct.unpack("<I", self.report.xheader[0:4])[0]
+        return magic in (MH_MAGIC, MH_CIGAM, MH_MAGIC_64, MH_CIGAM_64)
+
     def _is64(self):
-        if self._hasHeader(4):
+        if self._hasMachoHeader(4):
             magic = struct.unpack("<I", self.report.xheader[0:4])[0]
             return magic in (MH_MAGIC_64, MH_CIGAM_64)
         return self.report.bitness != 32
 
     def _getCpuType(self):
-        if self._hasHeader(8):
+        if self._hasMachoHeader(8):
             cpu_type = struct.unpack("<I", self.report.xheader[4:8])[0]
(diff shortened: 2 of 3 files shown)


r0ny123:
Heads-up on merge order: this and #238 touch no source file in common, but both append a test class to `tests/testSynthesis.py` and both add `import types` there, so whichever merges second will conflict in that file only. Verified by merging the two branches locally — the conflict is confined to the two appended classes and the import line, and the resolution is to keep both. Happy to rebase whichever one you take second.

danielplohmann:

PR #238, opened 2026-08-11
fix(core): lower the synthetic image base so low-RVA functions fit

### LINKED ISSUE #236: PE synthesis drops functions below RVA 0x1000, losing the entry region of ELF-origin reports
PE synthesis cannot place a function that sits below RVA `0x1000`, so every ELF-origin report whose code starts just past the ELF header loses its first functions. ELF and Mach-O synthesis keep all of them.

The blocks are reported and skipped rather than dropped silently, so the log does say what happened — but the synthesized PE is missing real code, which matters for the round-trip property #232 is about.

## Reproducing

```python
import lief
from pathlib import Path

raw = Path("tests/bashlite_xored").read_bytes()

parsed = lief.parse(list(bytes(report.synthesizeBinary(output_format="pe"))))
print("text starts at abs 0x%x" % (parsed.optional_header.imagebase + parsed.sections[0].virtual_address))
print("lowest function  0x%x" % min(report.xcfg))
print("functions below RVA 0x1000:", sum(1 for o in report.xcfg if o - report.base_addr < 0x1000))
```

```
text starts at abs 0x401000
lowest function  0x4000e8
functions below RVA 0x1000: 20
```

The synthesizer emits `block 0x4000e8 of function 0x4000e8 fits no executable section, skipped` for each one.

## Cause

`PeSynthesizer._synthesizeMinimal` (`src/smda/synthesis/PeSynthesizer.py:430`) floors the `.text` start at one section alignment:

```python
text_vaddr = max(align_down(min_rva, section_alignment), section_alignment)
```

which is correct as far as it goes — a PE section cannot start at RVA 0, that range belongs to the headers. But the image base is then left at the report's own base, so any function whose RVA is below `0x1000` has no section that can hold it and is skipped at planting time.

ELF binaries put their entry stub immediately after the program headers, so this is the normal layout for them rather than an edge case: all four affected fixtures are ELF-origin, with lowest RVAs between `0x94` and `0x158`.

## Measured

| fixture | base | lowest RVA | functions below `0x1000` | PE | ELF | Mach-O |
|---|---|---|---|---|---|---|
| `bashlite` | `0x400000` | `0xe8` | 20 | **7733 / 8767** | 8767 / 8767 | 8767 / 8767 |
| `mirai_i386` | `0x8048000` | `0x94` | 13 | **13365 / 14758** | 14758 / 14758 | 14758 / 14758 |
| `mirai_x64` | `0x400000` | `0x100` | 5 | **11762 / 12564** | 12564 / 12564 | 12564 / 12564 |
| `aarch64_static` | `0x400000` | `0x158` | 15 | **18963 / 19881** | 19881 / 19881 | 19881 / 19881 |

Measured with the round-trip loop from #232 against `master` at `5ad5504`. PE-origin fixtures are unaffected, since a PE's own code already starts above its headers.

## Suggested direction

The image base is the free variable: choosing it so that the lowest function extent lands at or above `section_alignment` — rather than inheriting the report's base and clamping `.text` — would let every function be planted at a correct RVA, at the cost of a synthetic base that differs from the report's. That is a visible trade-off, so it seems worth your call before anyone implements it; the alternative of keeping the base and accepting the loss is defensible too, in which case the skip is arguably worth surfacing more loudly than a log line.

r0ny123:
Fixes #236.

## The trade-off, first

The issue notes this one has a visible decision in it, so to be explicit about what this PR chooses: **the synthesized PE's `ImageBase` no longer always equals `report.base_addr`.** It is lowered, to a 64K boundary, when and only when the lowest function would otherwise fall below the first section-aligned RVA. Every function keeps its original absolute address — the RVAs shift up by exactly the amount the base came down — so a reader of the image sees the same addresses it would have seen, and the field that changes is one a synthetic image cannot really claim to preserve anyway.

The alternative, keeping the base and accepting the loss, is left available: nothing is lowered for a report that already clears the headers, so PE-origin reports are byte-identical to today.

## Cause

`_synthesizeMinimal` floors `.text` at one section alignment, which is correct — a PE section cannot start at RVA 0, that range belongs to the headers:

```python
text_vaddr = max(align_down(min_rva, section_alignment), section_alignment)
```

but the image base stayed at the report's, so a function below RVA `0x1000` had no section that could hold it and was skipped at planting time. ELF binaries put their entry stub immediately after the program headers, so this is their normal layout rather than an edge case.

`aarch64_switch_macho_O0` shows the sharper end of it: every one of its functions sits below the floor, `text_size` comes out non-positive, and synthesis raises `functions do not fit into a synthetic .text section`.

## Measured

| fixture | base | lowest RVA | functions below `0x1000` | PE before | PE after |
|---|---|---|---|---|---|
| `bashlite` | `0x400000` | `0xe8` | 20 | 7733 / 8767 | **8767 / 8767** |
| `mirai_i386` | `0x8048000` | `0x94` | 13 | 13365 / 14758 | **14758 / 14758** |
| `mirai_x64` | `0x400000` | `0x100` | 5 | 11762 / 12564 | **12564 / 12564** |
| `aarch64_static` | `0x400000` | `0x158` | 15 | 18963 / 19881 | **19881 / 19881** |
| `aarch64_switch_macho_O0` | `0x100000000` | `0x460` | all | `ValueError` | **322 / 322** |

Round-trip loop from #232, against `master` at `5ad5504`. Every bundled fixture was swept in all three formats; those five PE columns are the only ones that move, and no ELF or Mach-O column changes.

## A regression this caught on itself

The first version moved the entry point. `oep` is relative to the report's own base, and the PE path resolved it against the synthetic one:

```python
entry_rva = self.report.oep - base if self.report.oep >= base else self.report.oep
```

so lowering the base by `0x10000` moved the absolute entry with it — `bashlite` went from `0x400194` to `0x3f0194`. It is now resolved to an absolute address first and turned back into an RVA, which is the form `ElfSynthesizer` already used. Verified across every fixture: the absolute entry point and the section layout are byte-identical to `master`, including for the reports whose base is lowered.

## Validation

Full suite green (1,186 passed, 1 skipped), `ruff check`, `ruff format --check` and `ty` clean with zero errors, diff coverage 100%. Image bases stay 64K-aligned as the format requires, no synthesized image has overlapping sections, and all of them still parse.

Five regression tests, three of which fail on `master`: the low-RVA functions being planted, the lowered base staying 64K-aligned and below the report's, the entry point keeping its absolute address, a report that already clears the headers keeping its base unchanged, and a base-0 report never being lowered below zero.

Independent of #234 and #237 — different files, no shared lines.

### DIFF
--- a/src/smda/synthesis/PeSynthesizer.py
+++ b/src/smda/synthesis/PeSynthesizer.py
@@ -37,6 +37,8 @@ MACHINE_I386 = 0x014C
 MACHINE_AMD64 = 0x8664
 MACHINE_ARM64 = 0xAA64
 
+IMAGE_BASE_ALIGNMENT = 0x10000
+
 MAX_SYNTHETIC_SECTION_SPAN = 0x2000000
 
 
@@ -381,7 +383,7 @@ class PeSynthesizer(BinarySynthesizer):
         )
         return bytes(header) + b"".join(bytes(r["raw"]) for r in regions)
 
-    def _buildMinimalOptionalHeader(self, regions, bitness, size_of_headers, image_end, entry_rva, import_dir):
+    def _buildMinimalOptionalHeader(self, regions, bitness, size_of_headers, image_end, entry_rva, import_dir, base):
         is_64 = bitness == 64
         dir_base = DATA_DIR_BASE_PE32_PLUS if is_64 else DATA_DIR_BASE_PE32
         size = 0xF0 if is_64 else 0xE0
@@ -395,9 +397,9 @@ class PeSynthesizer(BinarySynthesizer):
         text_rva = min((r["vaddr"] for r in regions if r["executable"]), default=0)
         struct.pack_into("<I", opt, 20, text_rva)
         if is_64:
-            struct.pack_into("<Q", opt, 24, self.report.base_addr)
+            struct.pack_into("<Q", opt, 24, base)
         else:
-            struct.pack_into("<I", opt, 28, self.report.base_addr & 0xFFFFFFFF)
+            struct.pack_into("<I", opt, 28, base & 0xFFFFFFFF)
             data_rva = min((r["vaddr"] for r in regions if not r["executable"]), default=0)
             struct.pack_into("<I", opt, 24, data_rva)
         struct.pack_into("<II", opt, 32, 0x1000, 0x200)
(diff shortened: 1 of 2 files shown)


r0ny123:
Heads-up on merge order: this and #237 touch no source file in common, but both append a test class to `tests/testSynthesis.py` and both add `import types` there, so whichever merges second will conflict in that file only. Verified by merging the two branches locally — the conflict is confined to the two appended classes and the import line, and the resolution is to keep both. Happy to rebase whichever one you take second.

Also added `53db5e5` after opening: the `cutwail` fixture carries a real PE header, so it synthesizes from that layout and never reaches the lowering — the test named for "a report that already clears the headers" was actually pinning the xheader path. Renamed to say so, and covered the synthetic path's own no-op case directly, since no bundled fixture exercises it with a function that already clears the headers.

danielplohmann:

PR #245, opened 2026-08-13
feat(labels): rebuild the PDB symbol provider on purepdb

### LINKED ISSUE #230: PdbSymbolProvider silently disabled without pdbparse (GPL, unmaintained); llvm-pdbutil is a permissive alternative
## Summary

`PdbSymbolProvider` disables itself silently when `pdbparse` is absent, and `pdbparse` is both **GPL-licensed** (SMDA is BSD-2-Clause) and **unmaintained since 2020**. In practice this means PDB-bearing PEs get no names at all, with no indication that anything was skipped.

## Current behaviour

```python
try:
    import pdbparse
    from pdbparse.undname import undname
except ImportError:
    pdbparse = None
    LOGGER.debug("3rd party library pdbparse (use fork @ ...) not installed - ...")
```

`pdbparse` is not a declared dependency, so a default install never has it. The message is at `debug` level, so the default logging configuration hides it. Result: a PE with a perfectly good PDB beside it yields 1 named function (`original_entry_point`), and nothing says why.

## The dependency itself is problematic

| | |
|---|---|
| PyPI `pdbparse` licence | **GPL** (`License :: OSI Approved :: GNU General Public License (GPL)`) |
| SMDA licence | **BSD 2-Clause** |
| Last `pdbparse` release | **2020-02-01** (v1.5) |
| Upstream state | the working variant is a third-party fork (VPaulV), not the PyPI package |

An optional import softens the licence question, but recommending a GPL library in the docstring of a BSD project is still worth a second look.

## I checked whether anything better exists in 2026 — it does not

You mentioned the original goal was pure-Python with no native dependency. Re-surveying:

- **`pdbpy`** ("Pure python implementation of parsing PDB debug information files") — a single `0.0.1` release, and it does not import: `ModuleNotFoundError: No module named 'dtypes'`. Not viable.
- **`lief.pdb`** — the API exists (`lief.pdb.load`, `DebugInfo`, `Function`, `PublicSymbol`), which would be ideal since LIEF is already a dependency. But it requires **LIEF "extended"**, the commercial build; the open PyPI wheel returns `None` with *"Debug info not available for this build."*
- **`libpdb`** (shareef12) — C, Linux, would need bindings.
- **`pdbparse`** — as above.

So there is still no maintained, permissively-licensed, pure-Python option.

## Suggestion: an optional `llvm-pdbutil` provider

`llvm-pdbutil` is maintained, ships in distro LLVM packages, and reads `rust-lld`-produced PDBs correctly (verified). Parsing its output requires no new Python dependency and no GPL code:

- `llvm-pdbutil dump --publics <pdb>` → `S_PUB32` records, raw mangled names, with `addr = SECT:OFFSET`
- `llvm-pdbutil dump --symbols <pdb>` → `S_GPROC32` / `S_LPROC32` records, with `addr = SECT:OFFSET, code size = N`

Virtual address = `imagebase + sections[sect-1].virtual_address + offset`.

Two findings from implementing this that are worth having in SMDA regardless of the parser used:

1. **Publics alone are a poor source.** On a Rust msvc binary, publics gave 31.4% naming; adding the `S_GPROC32`/`S_LPROC32` procedure records took it to 54.8%. The publics stream includes imports and data and omits many functions.

2. **Most of the remainder are fragments, not anonymous functions.** MSVC-style codegen outlines cold paths and EH funclets into separate chunks, and SMDA reports each chunk as its own function. Of 1470 functions with no symbol at their entry, **1439 (97.9%) fall inside a procedure's `[addr, addr+code size)` range**. Labelling those from the containing procedure took naming from 54.8% to **99.0%** (99.0% of functions above the 10-instruction threshold).

   Worth noting for any implementation: label such a fragment relative to its parent (`<parent>$+0x<delta>`) rather than with an absolute offset, so the label is stable across builds and does not break MCRIT's `(pic_hash, function_name)` dedup.

## Measured impact

Same binary (Rust `regex` + `flate2`, `x86_64-pc-windows-msvc`, `debuginfo=2`, linked with `rust-lld`), 3391 functions:

| | named |
|---|---|
| SMDA 4.4.4 as shipped | 1 (0.0%) |
| publics only | 1066 (31.4%) |
| publics + proc records | 1858 (54.8%) |
| + fragment attribution | **3356 (99.0%)** |

## Minimal ask

Even without changing the parser, raising the "pdbparse not installed" message from `debug` to `warning` — or warning when a `.pdb` sits next to the input and no PDB provider is active — would turn a silent no-op into something diagnosable.

Versions: SMDA 4.4.4, Python 3.12.13, LLVM 14, Linux x86_64.

r0ny123:
Closes #230.

## Dependency

Requires `purepdb>=0.3.0`, which is the first release carrying the API this is written against: `Function.module` and `.aliases`, `PDB.functions(code_publics=...)`, and `diagnose().warnings`. The floor matters more than usual here — `0.2.0` installs cleanly and shares its version number with a much smaller API, so nothing but the pin separates them, and no runtime guard could.

Everything below was re-measured against the release installed from PyPI, not a source checkout.

## What was wrong

`PdbSymbolProvider` was gated on an optional `pdbparse` import. `pdbparse` is GPL-licensed against a BSD-2-Clause project, unmaintained since 2020, and was never a declared dependency — so a default install had no PDB support at all. The import failure was reported at `debug`, which the default logging configuration hides, so a PE with a perfectly good PDB beside it came back with no names and nothing said why.

## What this changes

`purepdb` replaces it as a declared, permissively-licensed, pure-Python dependency with no native code.

Names now come from three merged record sources rather than the global symbol stream alone: `S_GPROC32`/`S_LPROC32` procedures, publics, and `S_THUNK32` thunks. Each address is resolved through the section table the symbols are actually expressed against — including OMAP-translated images — instead of being recomputed here from `imagebase + section.virtual_address + offset`.

**Fragment attribution** is the part that moves the number. Most of what remains unnamed after the record merge is not anonymous functions but fragments: MSVC-style codegen outlines cold paths and EH funclets into separate chunks, and SMDA reports each chunk as a function of its own. A function with no symbol at its entry that falls inside some procedure's `[rva, rva + code_size)` is now labelled from that procedure as `<parent>$+0x<delta>`. The label is parent-relative rather than absolute so that it survives a rebuild and a different load address — which is what MCRIT's `(pic_hash, function_name)` dedup depends on. Import thunks are excluded from being fragment parents, identified by the linker input the PDB records for the address (`Import:foo.dll` in the Section Contribution table) rather than by the shape of the name.

**MSVC-decorated names are reported as the PDB stores them.** purepdb does not demangle, by design. `pdbparse` supplied `undname`, so this is a deliberate behaviour change, not an oversight: an MSVC-decorated name now reaches the report decorated, and there is no maintained, permissively-licensed MSVC demangler to substitute. A decorated name is also stable, which an approximate expansion would not be. Thunk records partly cover the gap on x86, where the thunk carries the undecorated import name against the public's decorated one; purepdb keeps both and the extra spellings land in `Function.aliases`.

**Rust names are demangled**, by the demangler this codebase already applies to ELF, Mach-O and PE symbol tables. It could not reach PDB names on its own, because it is driven from a parsed binary and a PDB arrives as a separate file path. Current toolchains write readable names into their proc records, so most names are already legible without this — but not all of them: on the rust-lld/MSVC x64 image measured below, 20 of 399 functions have no proc record and arrive with the mangled public in the name slot, and a PDB carrying only publics would see all of them. Names are parsed before they are rewritten, so `_RTC_Initialize`, `_ReadFile@20` and `_RoInitialize@4` — real names from a sqlite3 PDB, all matching a bare `_R` prefix — are left alone.

**The silent no-op is fixed.** A `.pdb` sitting next to the input while no PDB has been loaded now warns; so does a file supplied as a PDB that is not one; so does a PDB that fails to parse.

Failing to parse is not the only way to come back empty, though — every failure mode purepdb has on a real file is an *empty result* rather than an exception, so the silence has a second door. A PDB with no section-header stream parses cleanly and resolves every RVA to `None`; a managed (.NET) PDB carries records keyed by metadata token with no address to resolve. Both yield zero symbols from a successful parse. Both now warn, quoting the reasons `diagnose()` gives, so the message says which case it was. The third shape degrades rather than fails: publics but no procedure records — `/DEBUG:FASTLINK` and pre-2010 toolchains — still names entry points, but with no code sizes there is nothing to attribute a fragment to, and that is worth saying out loud given how much it costs. A PDB with both symbols and procedure records stays silent; verified on all the fixtures below, whose naming is unchanged.

## Measured

`Disassembler.disassembleFile(exe, pdb_path=pdb)`, naming counted over every reported function. The issue's binary is not in hand, so these are different binaries and different numbers.

| binary | functions | named before | named after | of which fragments |
|---|---|---|---|---|
| Rust `x86_64-pc-windows-msvc`, rust-lld | 477 | 1 (0.2%) | 442 (92.7%) | 46 |
| sqlite3 x64, `link.exe` | 4045 | 273 (6.8%) | 3698 (91.4%) | 97 |
| sqlite3 x86, `link.exe` | 4103 | 273 (6.7%) | 3730 (90.9%) | 111 |
| Rust i686, rust-lld, `no_std` | 6 | 1 | 6 (100%) | 0 |

Function totals move because recovered symbols also seed candidates.

On the Rust x64 binary the 35 that stay unnamed all lie past the last procedure record, in ranges the PDB carries no proc, no public and no section contribution for — precompiled `core::fmt` bodies from the std rlibs. That is the ceiling for that file, not a gap in the attribution.

Stability, checked on all four: analyzing the same input twice produces an identical `{offset: name}` map; no label contains a virtual address; and a fragment label is unchanged when the same PDB is applied at base `0x400000` versus `0x10000000`.

## Also in this change

`_parseOep` is dropped. It only ever saw the first 16 bytes of its target while PE entry-point extraction needs at least `0x40` (`e_lfanew` lives at `0x3C`), so it could not produce a symbol under any input; `PeSymbolProvider` is what actually supplies `original_entry_point`.

`tests/rust_pe_msvc_i686{,_pdb}_xored` is a new fixture pair, xored like the other binary fixtures. It is purepdb's own i686 groundtruth build — `#![no_std] #![no_main]`, linked by rust-lld with `/nodefaultlib`, so it links against no CRT and no import library and involves no Microsoft-licensed material — regenerable from the `build.sh` and `main.rs` beside it there. 84 KB for the pair, and paired so the PDB can be checked against the image it describes.

## Validation

Against `purepdb==0.3.0` from PyPI: `ruff check .` and `ruff format --check .` clean; `make typecheck` exits 0 with no new diagnostics; `pytest tests/` 1207 passed, 1 skipped, 1480 subtests; `diff-cover --fail-under=100` reports 60/60 changed lines covered. A PE with no PDB beside it is unaffected — covered by a test that removes the PDB and asserts the previous naming.

## Known follow-up

The same "confirmed the format, then failed silently" shape exists outside this provider: the CIL, Rust, Mach-O, Delphi and Go symbol providers and all three file loaders report a whole-parse failure at `debug`, and two of those sites log nothing at all. Those are left out here to keep this change to one subject; they need failure-path fixtures the suite does not currently have.


r0ny123:
Pushed a follow-up commit that demangles Rust names on this new provider.

## Why it belongs here rather than in its own PR

This PR's provider reports every name as the PDB spells it, which for a Rust image means `_RNvCs8JfRDQSkzJ8_15rust_pe_symbols4main`. The demangler this codebase already applies to ELF, Mach-O and PE symbol tables cannot reach those names: `RustSymbolProvider` is driven from a LIEF-parsed binary, and a PDB never becomes one — it arrives as a separate file path. So without this, a function whose only name lives in a PDB stays mangled no matter what else is loaded.

## What it does

Each recovered name goes through the same evidence gate the other providers use — exact prefix, full parse, and a hash-suffix requirement for the legacy form — so a name is parsed before it is rewritten. A bare `_R` prefix also matches `_RTC_Initialize`, `_ReadFile@20` and `_RoInitialize@4`, all real names carrying the function flag in a sqlite3 PDB, and corrupting a good name is worse than leaving a mangled one alone. Anything that fails to parse, demangles to an empty path, or raises keeps the spelling the PDB gave it.

Fragment labels follow the entry point they are attributed to, so a fragment inside a Rust procedure reads `rust_pe_symbols::main$+0x60` instead of carrying the mangled parent.

MSVC decoration is deliberately untouched — `?mangled@@YAXXZ` still comes back verbatim, and there is a test asserting that. Dropping `pdbparse` removed the only `undname` available here, so that half needs its own answer.

## Measurements

Against the fixtures in the purepdb repo — a rust-lld/MSVC x64 image and sqlite3 x86/x64:

| check | result |
|---|---|
| names this rewrites, vs `rustc-demangle` 0.1.28 | byte-identical, 0 mismatches |
| non-Rust names altered | 0 |
| empty labels produced | 0 |
| sqlite3 x86/x64 names (7221) | 0 touched |

Worth recording how few names this actually affects on that image, because it is not obvious: `rustc` on the MSVC toolchain writes proc records whose names are already readable, and the mangled `_R` spelling exists as a *public* at the same address, which `functions()` keeps as an alias since procs win the name slot. So of 399 functions only 20 arrive with a mangled name in the name slot — the ones that have no proc record. A publics-only PDB would see all of them.

Local gates: full suite green, `make lint`, `make typecheck` and `ruff format --check` clean.

r0ny123:
Follow-up on the one gap this PR states it leaves open.

The body says MSVC decoration is deliberately untouched because *"there is no maintained, permissively-licensed MSVC demangler to substitute"*, and that the half needs its own answer. It has one now: #255 writes that demangler — it reads all 609 names of the reference corpus exactly as `llvm-undname` spells them, with none given a different spelling — and #257 stacks on this branch and #255 to route the PDB provider through it.

Nothing here needs to change. The reasoning in the body was right for this PR at the time, and the decorated-name behaviour it describes is what a reader should still expect from this branch alone. Recording the pointer so the gap does not read as open.

danielplohmann:
Superseded by #289 — rebased onto current master (post #282–#288).

Issue #247, opened 2026-08-14
Function discovery finds only 44% of functions in one 32-bit MSVC PE (gap scanning contributes nothing)


danielplohmann:
## Summary

On one `i686-pc-windows-msvc` binary, SMDA reports **1167 functions** where the companion PDB describes **2656 distinct function entry addresses** (2296 `S_[GL]PROC32` procedure records, 1790 publics).

The ratio is not the interesting part — the shape of the miss is:

```
SMDA functions                      : 1167
PDB entry addresses (procs+publics) : 2656  (procs 2296, publics 1790)
SMDA functions with a PDB entry     : 1167  (100.0%)
SMDA functions WITHOUT one          : 0
PDB entries SMDA did not find       : 1489
```

**Every function SMDA found is one the debug info also describes, and it found none beyond that set.** SMDA is not reading the PDB here — it names 1 function of 1167 — so this is not a naming effect. It means that on this binary, whatever normally recovers functions that are *not* reachable by direct call contributed nothing at all.

A binary where discovery is working normally has a substantial population of functions with no symbol at their entry (outlined exception funclets, cold chunks). Here that population is exactly empty, which is what makes it look like a discovery problem rather than an accounting one.

## Versions

Reproduced identically on **4.4.4** and **4.4.7** (current release): 1167 functions in both. Python 3.12.13, lief 1.0.0.

## Reproducer

Bundle attached separately (`hyperfine.exe` + `hyperfine.pdb` + two scripts, ~4 MB):

```
pip install smda lief
python3 repro.py hyperfine.exe hyperfine.pdb            # the counts
python3 entry_coverage.py hyperfine.exe hyperfine.pdb   # the diagnostic above
```

`entry_coverage.py` needs `llvm-pdbutil` on PATH for the PDB side.

## Probably not the target triple

The same build recipe, toolchain and target applied to **ripgrep 14.1.1** gives **15,370 functions**, with the expected large population of unnamed-at-entry functions. So `i686-pc-windows-msvc` as such is fine; it is this binary that comes out empty.

## One property that may matter

[...]

PR #248, opened 2026-08-14
fix(labels): stop Delphi method-table walks at the first unreadable entry

r0ny123:
Rebased onto 802e627 to clear a merge conflict, and the conflict turned out to be informative: half of what this PR proposed had already landed by another route, so that half is now dropped and only the part that is still missing remains.

## What the rebase resolved

The original PR changed two `continue`s to `break`s in `JumpTableAnalyzer` and two more in the Delphi walks. #285's jump-table rewrite reached the same conclusion for the intel half independently, and did it better: the short-read stop is there at the entry read, and the `rebased < 0` check was hoisted out of the loop into an early `return` rather than being tested once per declared entry, which is right because the condition is loop-invariant.

Both conflicts were resolved in favour of master. `src/smda/intel/JumpTableAnalyzer.py` is now byte-identical to master, and this PR makes no intel source change at all.

## What is left, and why it is still needed

The Delphi half is unchanged on master:

- `DelphiPythiaProvider._extract_dynamic_methods` reads a method pointer and never checks for `None`, so an unreadable entry falls through to `function_offset += self._ptr_size` and the walk continues to its declared `table_length`. The two neighbouring walks in the same file already stop on exactly this condition, so this site is the odd one out rather than a deliberate choice.
- `DelphiReSymProvider` skips an unreadable method-entry pointer and keeps going, walking the whole declared table when the first failure already establishes the table is not there.

Same root cause as the intel half: a declared table length comes from the binary, so treating an unreadable entry as skippable spends that length as an iteration count.

## Tests

The intel tests are kept even though the intel fix is gone. They pass against master's implementation unmodified, and nothing on master pins the behaviour today, so they turn a property two separate changes arrived at independently into one a later refactor cannot quietly undo. Their commit is now `test(intel)` rather than `fix(intel)`, since it no longer changes any source.

| test | pins |
| --- | --- |
| `test_direct_table_scan_stops_at_the_image_edge` | direct scan stops at the image edge |
| `test_relative_table_scan_stops_at_the_buffer_end` | relative scan stops at the end of the buffer |
| `test_dynamic_method_walk_stops_at_the_image_edge` | Pythia dynamic-method walk stops |
| `test_method_definition_table_walk_stops_at_the_image_edge` | Pythia method-definition walk stops |

## Validation

| command | result |
| --- | --- |
| `ruff check .` | clean |
| `ruff format --check .` | 246 files already formatted |
| `make typecheck` | exit 0 |
| full test suite | 1791 passed, 1 skipped, 2579 subtests |
| diff coverage vs master | 100%, 3/3 lines |

### DIFF
--- a/src/smda/common/labelprovider/DelphiPythiaProvider.py
+++ b/src/smda/common/labelprovider/DelphiPythiaProvider.py
@@ -366,6 +366,8 @@ class DelphiPythiaProvider(AbstractLabelProvider):
         function_offset = dynamic_offset + 2 + (2 * table_length)
         for _ in range(table_length):
             method_addr = self._read_ptr(function_offset, self._ptr_size)
+            if method_addr is None:
+                break
             if method_addr and self._binary_info.isInCodeAreas(method_addr):
                 function_offsets.add(method_addr)
             function_offset += self._ptr_size
--- a/src/smda/common/labelprovider/DelphiReSymProvider.py
+++ b/src/smda/common/labelprovider/DelphiReSymProvider.py
@@ -596,7 +596,7 @@ class DelphiReSymProvider(AbstractLabelProvider):
             me_addr = self._read_ptr(me_ref_offset)
 
             if me_addr is None:
-                continue
+                break
 
             me_offset = self._addr_to_offset(me_addr)
             method_info = self._extract_method_info(me_offset)
--- a/tests/testDelphiPythiaProvider.py
+++ b/tests/testDelphiPythiaProvider.py
@@ -189,5 +189,55 @@ class TestDelphiReSymNegativeOffsets(unittest.TestCase):
         self.assertIsNone(parser._extract_method_info(-1))
 
 
+class TestDelphiTableWalksStopAtTheImageEdge(unittest.TestCase):
+    """Both walks take their entry count from a 16-bit field in the image, so a corrupt or
+    hostile table declares up to 65535 entries. The read offset only grows, so the first
+    unreadable entry means every later one is unreadable too: the walk has to stop there
+    instead of running out the declared count."""
+
+    def test_dynamic_method_walk_stops_at_the_image_edge(self):
+        provider = DelphiPythiaProvider.__new__(DelphiPythiaProvider)
+        provider._base_addr = 0x400000
+        provider._bitness = 32
+        provider._binary_info = SimpleNamespace(isInCodeAreas=lambda addr: True)
+        table_length = 0xFFFF
+        # the pointer array follows the 2-byte count plus one word per declared entry
+        pointers_offset = 2 + 2 * table_length
+        binary = bytearray(pointers_offset + 8)
+        binary[0:2] = struct.pack("<H", table_length)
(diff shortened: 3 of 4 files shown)


danielplohmann:
Independent corroboration on a **second, unrelated binary**, since the repro attached to #247 was a single artefact and this confirms the diagnosis generalises.

`win.spank_rat/a665646e` — a real-world 32-bit sample, `i686-pc-windows-msvc`, rustc **1.94.0**, no relation to the hyperfine reproducer (different project, different toolchain, different provenance). It hit the same stall, and I only noticed because the resulting measurement was wrong rather than because anything looked broken.

| | status | functions | wall |
|---|---|---|---|
| 4.4.7 unpatched, default `TIMEOUT=300` | **`timeout`** | 1182 | — |
| 4.4.7 unpatched, `TIMEOUT=3000` | `ok` | **4160** | 1381.6 s |
| **this PR's head (f9d2349), default `TIMEOUT=300`** | `ok` | **4160** | **4.4 s** |

So on this binary too: **output-equivalent, and ~314× faster.** The identical function count between the raised-budget unpatched run and the patched run is the same "identical set" property your validation table reports, reproduced independently.

The `1182` is worth noting alongside your `1167` — two different i686 Rust PEs truncating to almost the same size is consistent with the stall being in float-formatting code that both link from `std` rather than anything project-specific.

## Why it is easy to miss

The truncated report does not look damaged. In my case it produced a plausible-but-wrong measurement that I filed as a **negative** result: I was measuring what fraction of each binary's functions are EH cleanup funclets, and this sample came out at **0.8%** — the lowest of its entire class, which I recorded as "this one is not affected". Re-run with the budget raised it is **29.2%**, comfortably in the affected group.

The reason it lands that way is exactly the mechanism in your root-cause section: analysis stops after the call-graph pass, so the surviving functions are the call-reachable ones — which are precisely the *non*-funclets, since a cleanup funclet is reached only through the unwinder. The truncation therefore biases the result in a specific, plausible-looking direction rather than producing obvious garbage.

Not a request for anything in this PR — the fix is the fix. Recording it because "a truncated report reads like a complete one" is the part a downstream consumer is most likely to get wrong, and I did.

r0ny123:
Thanks — the second sample is the useful half of this, and the funclet observation is worth more than the timing numbers.

The bias direction is the part I had not stated explicitly: truncation keeps the call-reachable population, and an EH cleanup funclet is reached only through the unwinder, so it is systematically the first thing to disappear. Any per-binary ratio whose numerator is "functions not reached by a direct call" is therefore not merely noisy on a truncated report, it is skewed toward zero — which is exactly the shape of a result that gets filed as a negative rather than as a failure.

That is what makes "a truncated report reads like a complete one" worth fixing on its own, so I have opened #253 for it: one warning at default log level when the analysis is cut short, emitted on the timeout callback every backend polls, plus an incompleteness marker in the report's string form. `status`, `message` and the serialized schema are untouched, so nothing downstream has to adapt — it only stops the truncation from being silent.

The `1182` vs `1167` similarity fits the mechanism: the stall is in float-formatting code both binaries link from `std`, and both stop at the end of the same pass, so what survives is each binary's call-reachable set rather than anything project-specific.

PR #252, opened 2026-08-15
fix(labels): decode non-ASCII identifiers in Rust v0 symbols

### LINKED ISSUE #251: Rust v0: non-ASCII identifiers render as a punycode placeholder
A Rust v0 symbol with a non-ASCII identifier is reported with a placeholder in place of the identifier:

```
_RNqCs4fqI2P2rA04_11utf8_identsu30____7hkackfecea1cbdathfdh9hlq6y

smda      : utf8_idents::punycode{__-7hkackfecea1cbdathfdh9hlq6y}
rustc-demangle 0.1.28 : utf8_idents::საჭმელად_გემრიელი_სადილი
```

This is wrong output rather than a clean failure, which makes it worse than leaving the symbol mangled: the reported name is neither the mangled spelling nor the real one, so it cannot be matched against either.

`Ident.display()` in `rust_demangler/rust_v0.py` already tries to decode and falls back to the `punycode{...}` form when `try_small_punycode_decode()` returns `None`, which is what happens here. I have not root-caused why the decode fails, so I would not want the fallback simply removed — that would turn a wrong name into a crash-free but empty one. Worth checking whether the hand-rolled decoder can be replaced with the `punycode` codec in the standard library, which would also delete a chunk of hand-maintained state machine.

## How it was found

Cross-checking the demangler against the 18 mangled symbols in `rustc-demangle`'s own test corpus: 8 matched, 9 raised `UnableTov0Demangle` (the `#[splat]` grammar, which this port does not implement — raising is the safe outcome, the caller keeps the mangled name), and this one disagreed.

Worth repeating the check after any demangler change; it takes the reference crate's test file as input and needs no fixtures.

r0ny123:
Fixes #251.

A Rust v0 symbol carrying a punycode-encoded identifier came back with a placeholder where the identifier should be:

```
_RNqCs4fqI2P2rA04_11utf8_identsu30____7hkackfecea1cbdathfdh9hlq6y

before : utf8_idents::punycode{__-7hkackfecea1cbdathfdh9hlq6y}
after  : utf8_idents::საჭმელად_გემრიელი_სადილი
```

That is worse than leaving the symbol mangled, because the reported name matches neither the mangled spelling nor the real one, so it cannot be looked up either way.

## Root cause

The decoder was never broken. Running it on that identifier fills its output buffer with the correct text — the failure is purely in how it reports having done so.

`punycode_decode()` leaves its loop through a bare `return` when the input is exhausted, which yields `None`. Each of its seven failure paths also returns `None`. `try_small_punycode_decode()` then tests `if r is None` and cannot distinguish a completed decode from a rejected one, so every decode — successful or not — took the placeholder branch in `Ident.display()`.

The `-> Optional[None]` annotation says the same thing in the signature: the function had no way to express success.

The fix is to return `True` on completion and widen the annotation. A decode that genuinely fails still falls back to the placeholder, which now has its own test so the fallback is not lost to a later simplification.

## A second defect the fix made reachable

While every decode was being discarded, a second divergence from the reference was unobservable: the code point is built with `chr()`, which accepts the surrogate range `U+D800..U+DFFF`, where Rust's `char::from_u32` refuses it. `_RNvCu4_ib9b4main` decoded to a name holding a lone surrogate, and such a name cannot be encoded as UTF-8 by whatever consumes the report — `SmdaReport` itself survives because `json.dumps` escapes it, but the name is handed to every downstream consumer as a `str`.

Non-scalar values are now rejected explicitly, and the fuzz target encodes every result it gets so the class stays closed rather than resting on this one case.

## Scope of the sweep

The class is *a success path returning the value the failure paths use as their sentinel*. An AST pass over `src/smda/**` for functions mixing a bare `return` with an explicit `return None` finds no other instance, and no other `Optional[None]` annotation remains in the tree.

## Validation

Cross-checked against the 18 mangled symbols in `rustc-demangle` 0.1.28's own test corpus:

| | before | after |
|---|---|---|
| match exactly | 8 | **9** |
| raise, keeping the mangled name | 9 | 9 |
| disagree | 1 | **0** |

The nine that raise are the `#[splat]` grammar, which this port does not implement; raising is the safe outcome there, since the caller then keeps the name the binary gave it.

Separately, on 147 real `_R` symbols read out of a rust-lld/MSVC x64 image, 146 match the reference byte for byte and the remaining one is the unrelated ABI spacing bug fixed in #249 — nothing regressed here.

```
python -m pytest tests/ -q                  1184 passed, 1 skipped
```

Independent of #249; the two touch different parts of the same file and both are needed for the output to match the reference exactly.

### DIFF
--- a/src/smda/common/labelprovider/rust_demangler/rust_v0.py
+++ b/src/smda/common/labelprovider/rust_demangler/rust_v0.py
@@ -101,7 +101,7 @@ class Ident:
         self.out[i] = c
         return True
 
-    def punycode_decode(self) -> Optional[None]:
+    def punycode_decode(self) -> Optional[bool]:
         count = 0
         punycode_bytes = self.punycode
         try:
@@ -151,10 +151,12 @@ class Ident:
             n += i // lent
             i %= lent
 
-            try:
-                c = chr(n)
-            except (ValueError, OverflowError):
+            # mirror char::from_u32: a scalar value, so no surrogate and nothing past the
+            # last code point. chr() accepts surrogates, and a name carrying one cannot be
+            # encoded as UTF-8 by whoever consumes the report.
+            if not 0 <= n <= 0x10FFFF or 0xD800 <= n <= 0xDFFF:
                 return None
+            c = chr(n)
 
             if not self.insert(i, c):
                 return None
@@ -163,7 +165,7 @@ class Ident:
             try:
                 punycode_bytes[count]
             except IndexError:
-                return
+                return True
 
             delta = delta // damp
             damp = 2
--- a/tests/testRustSymbolProvider.py
+++ b/tests/testRustSymbolProvider.py
@@ -11,6 +11,7 @@ from smda.common.labelprovider.rust_demangler import demangle
 from smda.common.labelprovider.rust_demangler.rust import TypeNotFoundError
 from smda.common.labelprovider.rust_demangler.rust_legacy import LegacyDemangler, UnableToLegacyDemangle
 from smda.common.labelprovider.rust_demangler.rust_v0 import (
+    Ident,
     Parser,
     Printer,
     UnableTov0Demangle,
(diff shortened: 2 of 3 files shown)


danielplohmann:
Already contained in master — `try_small_punycode_decode` returns `Optional[bool]` with surrogate rejection via `from_u32`, matching this PR exactly.

PR #255, opened 2026-08-15
feat(labels): demangle MSVC decorated symbol names

r0ny123:
MSVC-decorated names were the one mangling scheme SMDA carried no demangler for, so a C++ binary built with the Microsoft toolchain reported `?__crt_rotate_pointer_value@@YAIIH@Z` where a GCC-built one reported a signature. This adds one and wires it into the PE provider.

Part of a linear stack: #249 → #250 → #254 → **#255**. They are ordered that way because #249 and #254 both touch `RustSymbolProvider` and would otherwise conflict in a way that silently breaks `import smda` if resolved the obvious way, and because #250 creates the `NOTICE` this PR extends. Merging in that order needs no conflict resolution.

## The contract

A name comes back either as the reference implementation spells it, or exactly as it went in. Never as a third spelling.

That matters more here than coverage does. A decorated name is still a usable identity — it can be matched against a symbol server, an import table, or another report. A confidently wrong expansion is worse than no expansion, because it matches neither the decorated form nor the real one. So every construct the parser does not fully model declines rather than guessing, and the tests assert this over the whole corpus rather than sampling it.

## Measurements

`llvm-undname` is the reference. The corpus is the one LLVM tests its own demangler against — 609 unique names drawn from `llvm/test/Demangle/ms-*.test` (the upstream files repeat 15), which is a compiler stress suite rather than a sample of ordinary binaries: it is full of lambdas, anonymous namespaces, member pointers and deeply nested templates.

| | at PR open | now |
|---|---|---|
| spelled exactly as `llvm-undname` spells it | 383 (63%) | **609 (100%)** |
| returned unchanged | 226 | **0** |
| **given a different spelling** | **0** | **0** |

Every name in that corpus is now pinned individually as its own subtest, rather than by an aggregate count. That distinction is not cosmetic: a single count is how a common C++ construct — every function returning a class by value — sat unread in the declined bucket for the first half of this PR while both a 0-wrong corpus and 100% branch coverage stayed green.

On the MSVC names carried by real PDBs (the sqlite3 x86/x64 and rust-lld builds): 14 names, 14 exact, 0 wrong.

And on a PE built by the toolchain this is meant to read, all four decorated exports come back as signatures. `tests/msvc_cxx_pe_xored` is that binary, added here because none was available: no PE on hand carries MSVC-decorated names — sqlite3's 277 exports are all C — so this arm of the PE provider had only synthetic symbols behind it. It is a small translation unit built for `x86_64-pc-windows-msvc` by clang-cl 22.1.7, exporting free functions, a namespace, a class returned by value and an `extern "C"` name.

Its first real export was one the demangler could not read, and the reason generalised. A return type is the one position carrying a qualifier of its own, and that qualifier is not decoration: every function returning a class by value is spelled with it, `?A` being the unqualified case rather than an absent one. Parsing it — at all three places a return type is read — is worth 20 corpus names on its own. Twenty-one corpus names had used the form the whole time; they sat in the declined bucket, where a single aggregate count kept them from reading as a gap.

The corpus ships as `tests/msvc_reference_corpus.txt` so the demangler is measured against a reference rather than against its own output, and a regression that starts emitting third spellings fails the build rather than going unnoticed.

## What it understands

Qualified and nested names, name and argument back-references, the basic and extended type set, return types with their own qualifier, pointers and references with their own and their pointee's qualifiers, `__restrict`, arrays, function pointers, **declarator composition over all of those**, variadic parameter lists, class/struct/union/enum types, templates, data symbols with their storage class, constructors, destructors, operators, and the `vftable`/`vbtable` family.

A type is built as a tree and spelled around a declarator rather than rendered eagerly, because C nests the two — the name belongs inside its own type. That is what makes `int (__stdcall *j)(signed char, unsigned char)` and `int (* __cdecl f(void))[2]` come out right, and it is why a function pointer used as a data declaration, a return type, or the pointee of another pointer are ordinary cases rather than refusals.

Everything the corpus carries now reads: back-references as the mangler numbers them, local scopes and lambdas, member pointers to both data and functions, integer and function-type template arguments, unnamed namespaces, the RTTI family, conversion operators, string literals, thunks, user-defined literal operators, and the storage forms a data symbol may take.

## How the grammar was established

From Microsoft's documented decorated-name format, with `llvm-undname` used to settle every case where the documentation is silent or ambiguous — pointer qualifier composition in particular, where `PBQAD` and `PAQAD` both spell `char *const *`, was derived by probing the reference rather than reasoned about.

## Refusing what the grammar does not allow

A corpus of well-formed names only proves the demangler reads what it should. It says nothing about what it does with the rest, and a symbol table holds whatever bytes were written into it. So the corpus was also used as fuzzer seed: 15044 candidates built by truncating, splicing and mutating those 609 names, each fed to both implementations, asserting that wherever this demangler answers it agrees with the reference.

It is now seven seeds and 110495 derived names per run, and it is what caught most of what follows. Three seeds had looked clean; widening the set immediately found more, so a zero on a narrow seed set is not evidence.

The first run found 256 shapes expanded confidently while the reference refused them. Five fixes took it to 4, every one of them refusing more rather than expanding anything new:

- **A parameter list must be closed by its throw specification** — fixed lists end `@Z`, variadic ones `ZZ`. The terminator was consumed only when the parser was not already at end of input, so any name truncated just past its parameters got a plausible answer: `?a2@@YAHX` read as `int __cdecl a2(void)`, and `?f@@YAXHZ` as a variadic call whose last byte was never there.
- **`_D` through `_I` are not type codes.** No Microsoft-compatible mangler emits them — `__int8`, `__int16` and `__int32` are spelled with the plain `char`, `short` and `int` codes — so a name carrying one is not MSVC-decorated.
- **A template name is an identifier**, not a digit or an operator code.
- **A special name takes a signature or a storage class by which code it is, never both.** A `vftable` with a parameter list and an `operator!` with a vftable's storage class are equally malformed.
- **A reference cannot carry a qualifier from the type enclosing it** — C++ has no such form, and it had been rendering as `char const &const volatile *`.

Wrong answers across those seven seeds now stand at **2**, and both are the divergences kept on purpose, where the reference is the weaker side:

- `_L` and `_M` (`__int128`) are spelled from clang's Microsoft **mangler**, which emits them, because `llvm-undname` has no reader for what its own compiler produces.
- `??_X` and `??_Y` are spelled from Microsoft's table as the placement delete closures. The reference recognises them structurally but emits an empty name — `public: int __cdecl Base::(int)` — which is not a usable identity.

## Bounded against hostile input

Both resources a crafted name can spend are capped. Depth stops below the point where CPython's own recursion limit could be what ends a parse — otherwise the answer would depend on how deep in SMDA the provider happened to be called, which I verified is identical across three caller stack depths. The rendered result is capped too: argument back-references can reuse an earlier rendering repeatedly, which grew a 132-byte name into 496 MB of output (2.5 GB resident, retained by the cache) before the bound; it now declines in 0.02 ms.

## Robustness

Every prefix of every corpus name is fed through the demangler in the tests; none may raise. Malformed shapes — empty fragments, out-of-range back-references, unknown calling conventions, trailing bytes, missing terminators — each have a case asserting they come back untouched.

## Validation

```
python -m pytest tests/ -q                  1298 passed, 1 skipped, 2452 subtests
diff-cover coverage.xml --fail-under=100    100%
```

Throughput on the corpus: 155k names/s cold, 27.6M/s through the cache, worst single name 161 µs. A decorated name is demangled once per symbol at label-provider time, so this is nowhere near the analysis hot path, but it is bounded on hostile input rather than merely fast on ordinary input.

The new module is at 100% statement and branch coverage.

## Follow-up

The PDB provider is untouched here. On master it still gets MSVC names expanded by `pdbparse`; once #245 replaces that with purepdb, which does not demangle, this is what fills the gap — and that is #257, which stacks on #245 and this branch and does exactly that. It is not in this PR because it needs both underneath it.


r0ny123:
Pushed 099ac77, which hardens the demangler against names the grammar does not actually allow, and corrects one published number.

The corpus proves the demangler reads well-formed names correctly, but says nothing about what it does with the rest — and a symbol table holds whatever bytes were written into it. So I used the corpus as fuzzer seed instead: 15044 candidates built by truncating, splicing and mutating the 609 reference names, each fed to both implementations, asserting that wherever this demangler answers it agrees with `llvm-undname`.

256 shapes came back expanded confidently while the reference refused them. The largest class was a missing terminator: a parameter list is closed by its throw specification, but that byte was only required when the parser was not already at end of input, so `?a2@@YAHX` — a name truncated mid-encoding — was answered as `int __cdecl a2(void)`. Four smaller rules were missing too: an extended integer table carrying six codes no Microsoft-compatible mangler emits, a template name allowed to begin with a digit, a special name allowed to take both a signature and a storage class, and a reference allowed to carry a qualifier from the type enclosing it (which rendered as `char const &const volatile *`).

That is 256 → 4, and every change refuses more rather than expanding anything new: the corpus is unchanged at 363 of 609 exact with 0 given a different spelling.

The correction: the real-PDB figure is 11 of 14, not the 10 stated earlier. I re-measured the committed tree before and after this change and both give 11, so it was a measurement error on my side rather than a regression — the remainder is 3 local-scope names, not 4. The body now reflects that.

Two divergences from the reference are deliberate and now documented in the body, because copying it would make the output worse in both: `_L`/`_M` (`__int128`), which clang's Microsoft mangler emits but `llvm-undname` cannot read back, and `??_X`/`??_Y`, which the reference recognises but spells with an empty name (`public: int __cdecl Base::(int)`).

r0ny123:
Pushed 407537c, which adds the last missing piece of the type grammar — found by finally building a binary with the toolchain this is meant to read.

No PE available to me carries MSVC-decorated names (sqlite3's 277 exports are all C), so the MSVC arm of the PE provider had only synthetic symbols behind it while the Itanium arm had a real fixture. `tests/msvc_cxx_pe_xored` closes that: a small translation unit built for `x86_64-pc-windows-msvc` by clang-cl 22.1.7, exporting free functions, a namespace, a class returned by value and an `extern "C"` name.

Its very first real export was one the demangler refused — `?combine@geometry@@YA?AUMatrix@@AEBU2@NI@Z` — and the reason generalised. A return type is the one position that carries a qualifier of its own, and that qualifier is not decoration: every function returning a class by value is spelled with it, `?A` being the unqualified case rather than an absent one. A return type is parsed at three places (a plain function, a function pointer, a `$$A6` function type), so all three now share one reader instead of calling `type()` directly.

Corpus goes 363 → 383 of 609 exact, still 0 given a different spelling, and all four decorated exports of the new fixture come back as signatures.

Worth stating plainly, since it is the part I got wrong: 21 corpus names had used this form the whole time. They sat in the 246-name declined bucket, where a single aggregate count kept a common C++ construct from reading as a gap — a 0-wrong corpus and 100% branch coverage were both green throughout. The differential fuzzing in the previous commit could not have found it either, because it only asks whether a wrong answer is given, never whether a right one is missing.

r0ny123:
Pushed one more guard, found while auditing the demangling paths end to end.

The NUL check was written because *"an expansion holding a control character would travel into the report as a symbol name"* — and then tested only for NUL. Every other control byte went straight through: `?ctrl\x01@@YAXXZ` expanded to `void __cdecl ctrl\x01(void)`, putting a control character in a reported symbol name and from there in the serialized report. The parser reads them as ordinary identifier bytes, so it has no reason to reject one on its own; testing the input is what keeps the answer clean.

A decorated name is read from a NUL-terminated string of source-legal characters and cannot hold any of these, so refusing them costs no real name — the reference corpus is unchanged at 381 expanded.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1211 passed / 1 skipped / 1597 subtests, diff-cover 397/397.

r0ny123:
Back-references now resolve the way the mangler wrote them, which was the largest thing this demangler could not read.

The rules come from clang's own Microsoft **mangler** rather than the demangler that reads it: a name is recorded only when the table does not already hold it and while it holds fewer than ten; a template instantiation is written in a fresh scope that opens *before* the template's own name, so that name takes index 0 and its arguments are numbered from 1; the rendered template belongs to the enclosing scope — except the symbol's own name, which is not recorded at all, so `??$f@H@N@@YAXV0@@Z` resolves 0 to `N`.

Two shapes also stopped being answered: `B` was read as a reference introducer, which would make it a volatile-qualified reference that C++ cannot write and no mangler emits; and an argument back-reference was read as a type anywhere a type belongs, when it stands for a whole argument — `?h@@YAXPAHPA0@Z` was getting an answer for a name that cannot exist.

Measured against llvm-undname 22.1.8:

| | before | after |
|---|---|---|
| reference corpus spelled identically | 383 / 609 | **415 / 609** |
| corpus given a different spelling | 0 | **0** |
| fuzz wrong answers (seeds 20260817 / 7 / 99991) | 4 / 5 / 6 | **0 / 3 / 2** |
| fuzz names read | 2795 / 2774 / 2719 | **3047 / 3033 / 2985** |

Differential fuzzing is 15785 names per seed, derived by mutating and splicing the corpus. The five remaining wrong answers are shapes this demangler already answered before this change, in three families it still models incorrectly — `??_B` with the vftable storage form, `$$C` in a parameter list, and the `__ptr64` modifier in front of a function type. I tried refusing all three and it cost nine corpus names and added wrong answers elsewhere, so they are left as they were rather than guessed at; each has a reproducing name recorded.

One thing this surfaced: the corpus pair `?g@@YAXUS@@PA0@Z` asserted a demangled spelling that llvm-undname actually refuses, contradicting the file's own stated contract. It moves to the declined list.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1214 passed / 1 skipped / 1610 subtests, diff-cover 376/376.

r0ny123:
Widened the differential fuzzing from three seeds to seven — 110495 derived names — which found four more rules the grammar states and this did not.

- **The `6` storage form belongs to the vftable family.** Every data-special operator was reading it, so `??_Bx@@6B@` got an answer. vcall, typeof and the local static guard take a storage class this parser does not model, and now decline rather than borrow the vftable's. (Confirmed against the reference: `??_7`, `??_8` and `??_S` take it; `??_9`, `??_A`, `??_B` do not.)
- **No mangler writes `__ptr64` in front of a function type** — `P6A` and `R6A` are names, `PE6A` and `RE6A` are not.
- **`$$C` qualifies an array element or a template argument**, not a parameter of its own and not a pointee. Worth noting for review: my first attempt required a template scope and cost eight real names, because the array element is the other place it belongs.
- **A data symbol's trailing qualifier belongs to what its outermost pointer points at**, and was dropped whenever the type was not a plain one. `?s@@3PADB` is `char const *s`; `?s@@3PAPADB` is `char *const *s`. This one was a wrong *spelling* rather than an impossible name, so it would have reached a report.

Spacing follows from the same probing: a sigil abuts a type ending in neither an alphanumeric character nor `>`, while a named declarator is spaced off whatever precedes it — `struct S_*` and `enum <unnamed-type-*`, but `struct S *` and `enum E$ __cdecl fooE(void)`.

| | before today | now |
|---|---|---|
| corpus spelled identically | 383 / 609 | **415 / 609** |
| corpus spelled differently | 0 | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | 31 | **6** |

Of the six left: three are one spacing corner on mutant lambda *data* names (`??R<lambd@@3ABHB`), and two are the divergences this demangler keeps deliberately — `__int128`, which clang's mangler emits and the reference cannot read, and the vector-deleting-destructor family, which the reference spells with an empty operator name. Every rule above is pinned with the name that proved it.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1216 passed / 1 skipped / 1636 subtests, diff-cover 392/392.

r0ny123:
Unnamed namespaces are now read — every `namespace { }` in an MSVC-built binary is spelled `?A` with a discriminator, and a name qualified by one declined outright.

Two details decide it, both settled against the reference rather than assumed:

- The discriminator distinguishes two unnamed namespaces inside one binary and the reference does not spell it, so both render as ```anonymous namespace'``` — which is what the source looks like too.
- **It is the discriminator, not that spelling, that a later back-reference resolves to**: `?f@?A0x1@@YAXV1@@Z` names its parameter `class 0x1`. Recording the rendering instead shifts every later index and spells a different name. The mutated LLVM symbols in the fuzz corpus caught exactly that.

The fragment is read only where a namespace can appear: a leading `??A` is `operator[]`, and claiming it as a namespace turned `??AFoo@@QAGXXZ` into `Foo::`anonymous namespace'(void)`. That one was my own bug, found by the fuzzer within a minute of introducing it.

| | now |
|---|---|
| corpus spelled identically | **417 / 609** (from 383 at the start of this PR) |
| corpus spelled differently | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | **6** |

The six: three are one spacing corner on mutant lambda *data* names (`??R<lambd@@3ABHB` — a shape no compiler emits; two attempts to fix it each regressed the corpus, so it stays declined-as-is), and three are the deliberate divergences where the reference is the weaker side — `__int128`, which clang's own mangler emits and llvm-undname cannot read, and the vector-deleting-destructor family, which it spells with an empty operator name.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1218 passed / 1 skipped / 1646 subtests, diff-cover 100%.

r0ny123:
Two more constructs ordinary C++ produces, which between them accounted for most of what the corpus still declined.

**Pointer to member function** — how any callback into a class is spelled. The class qualifies the declarator rather than the type (`void (__thiscall S::*)(void)`), the member's own cv follows the parameter list where a member function keeps it, and the pointer's own qualifiers stay on the pointer (`Q8` is `S::*const`).

**Integer template arguments** — what `std::array<int, 3>` and every fixed-size container carry. A single digit stands for itself plus one; anything larger is nibbles `A`–`P` ended by `@`, with a leading `?` for negative. The accumulator is **64 bits and wraps**, which the corpus itself shows: the eighteen nibbles of `$0HPPPPPPPPPPPPPPPPPP@` are 18446744073709551615, and a magnitude wrapping to zero under a minus sign is spelled `-0`.

One implementation note worth flagging for review: parenthesising the declarator needed an *anchored* test, not a substring one. A rendered parameter may itself contain `::*`, and treating that as a member pointer put brackets around `std::forward`. The corpus caught it.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **457 / 609 (75%)** |
| corpus spelled differently | 0 | **0** |
| fuzz names read (per seed) | ~2795 | **~3250** |

Remaining fuzz divergences across seven seeds are 7, in four families: the `??R<lambd` spacing corner (a data symbol whose name starts with `<`, which no compiler emits — two attempts to fix it each regressed the corpus), the two deliberate divergences where the reference is the weaker side (`__int128`, which clang's mangler emits and llvm-undname cannot read; the vector-deleting-destructor family, which it spells with an empty operator name), and one spliced-mutant shape where `?M` appears where a name belongs.

What still declines is dominated by `$I`/`$U` member-pointer-with-thunk encodings and local-scope/lambda fragments; the latter need nested-symbol parsing, which is a real refactor of `parse()` rather than a rule a probe can settle.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1220 passed / 1 skipped / 1665 subtests, diff-cover 100%.

r0ny123:
Names scoped inside a function now read — lambdas, function-local statics and local classes, the largest thing left that an ordinary build emits. I had shelved this earlier because a first attempt introduced wrong answers; two rules settle it, and both were found by measuring rather than reasoning:

- **The enclosing name continues the outer back-reference table**, it does not open its own. `?N@?1??SN@?$NS@H@0@QEAAHXZ@4HA` resolves its `0` to the outer `N`; reading it in a fresh table spelled that `SN::NS<int>::SN`.
- **The enclosing name is a symbol**, so its own leading template is not recorded either — the same exception the outer name gets. Getting this wrong cost three wrong answers in the fuzz corpus and nothing on the reference corpus, which is exactly why the seed set matters.

The fragment stops where the enclosing name stops: the `@` after it terminates the qualified name it belongs to, not the fragment. The scope number is one less than the one spelled.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **484 / 609 (79%)** |
| corpus spelled differently | 0 | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | 31 | **7** |

The seven are unchanged by this commit — three the lambda-data spacing corner, two the `??_X` divergence, one `__int128`, one spliced mutant. Adding local scopes did not add any.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1222 passed / 1 skipped / 1673 subtests, diff-cover 100%.

r0ny123:
Three more forms an ordinary build emits, closing the largest of what still declined.

- **`__unaligned`** qualifies what a pointer points at, spelled after that pointee's own const and volatile (`int const __unaligned *`), and it *travels* with them — a pointer pointing at an unaligned pointer keeps it (`int __unaligned *__unaligned *`), which needed the qualifier merge to carry more than const/volatile.
- **User-defined literals**: `??__K_deg@@YAHO@Z` is `operator ""_deg`, the suffix being the identifier after the code. Any other double-underscore code still declines.
- **The zero scope**: a scope inside a function can be written `@` instead of a digit, and that one is spelled zero — a digit is spelled one higher than written, so the two forms meet at nothing.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **491 / 609 (81%)** |
| corpus spelled differently | 0 | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | 31 | **7** |

The seven are unchanged across the last four features: three the lambda-data spacing corner, two `??_X`, one `__int128`, one spliced mutant. What still declines is now dominated by `$I`/`$U` member-pointer-with-thunk encodings and `$$A8`/`$$Z` template forms — shapes a compiler emits rarely if ever.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1224 passed / 1 skipped / 1685 subtests, diff-cover 100%.

r0ny123:
One more correctness fix, and this one is worth flagging because **neither instrument caught it** — the reference corpus and the fuzz corpus both stayed exactly where they were.

`const MyClass instance` reached the report as `class MyClass inst`. The trailing qualifier of a data symbol was applied only when the declared type was a plain one, so every class, struct and enum silently lost it. Arrays had it worse: the qualifier was appended to the array node's own text, which holds dimensions rather than a type name, so `?arr@@3QAY01HB` came out `[2] const *const arr` — a spelling of nothing. That one arrived with my earlier pointer fix, which is why this replaces the special cases with a single rule.

The rule the reference follows is that the qualifier belongs to **what the symbol declares**: a pointer passes it one level in (`?s@@3PADB` → `char const *s`, while `?s@@3PAPADB` keeps it on the inner pointer), and an array passes it to its element the way C spells one (`?arr@@3QAY01HB` → `int const (*const arr)[2]`).

Since neither corpus carries these shapes, the nine spellings it settles are pinned directly as tests.

Also tried and **rejected** in this round: `$1`/`$E` address-of template arguments, worth ~5 corpus names. The nested symbol's back-reference behaviour matches neither obvious reading — sharing the enclosing table costs 4 corpus wrongs and ~35 per fuzz seed, isolating it costs 6 and ~60. Reverted rather than shipped on a guess; the measurements are recorded so a later attempt starts from them.

State: corpus **491 / 609 exact (81%)**, 0 spelled differently, 7 fuzz wrong answers across 110495 names — three of them the lambda-data spacing corner, three the deliberate divergences, one a spliced mutant.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1225 passed / 1 skipped / 1694 subtests, diff-cover 100%.

r0ny123:
Two more forms, and between them most of what still declined.

**A function type written as a template argument may carry what only a member function carries.** `$$A8` is that form — it names no class (the reference refuses one), so it reads as an ordinary function type with the qualifier appended. The qualifier is more than cv: a reference qualifier and `__restrict` are written *in front* of it and spelled *after* it, so `GB` is ` const &` and `IA` is ` __restrict`. Each is written at most once, so `HH` is not a name — the fuzzer found that within a seed of adding it.

**A scope inside a function carries the same number a template argument does**, and I had read it as a single digit. Nibbles are just as legal: `?L@` is the eleventh scope, and reading only digits declined every name past the ninth. `A` stays out of that set because `?A` is the unnamed namespace.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **509 / 609 (84%)** |
| corpus spelled differently | 0 | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | 31 | **7** |

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1230 passed / 1 skipped / 1703 subtests, diff-cover 100%.

r0ny123:
Pointers to data member now read. The code standing where a pointee's qualifier would be says both that the pointer points into a class and what the member is qualified by, so `PRfoo@@D` is `char const foo::*`; as a data symbol the same pointer repeats that qualifier and names its class again by back-reference (`?m@@3PRfoo@@DR1@`).

Two shapes stay **refused rather than guessed**:

- C++ has no reference to member, so `AT...` is not a name however much it parses like one.
- When the member type opens with a qualified pointer, the reference drops qualifiers this would keep — `PQfoo@@SAPEAX` is `void **foo::*` there, not `void *const volatile *foo::*` — and nothing on the producer side settles which is right, so it declines.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **516 / 609 (85%)** |
| corpus spelled differently | 0 | **0** |

Two more fuzz divergences appear with this, both the declarator-spacing corner already recorded, now reachable through a member pointer written as a data symbol — so 9 across seven seeds rather than 7, in the same four families.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1234 passed / 1 skipped / 1712 subtests, diff-cover 100%.

r0ny123:
Three storage forms a data symbol may take, and between them most of what still declined.

- **`__ptr64` stands in front of the qualifier**, where something is pointed at: `?s@@3PEAHEA` is a name and `?s@@3HEA` is not, since nothing there is pointed at to carry it.
- **A pointer into a class spells its storage the long way even so** — the member's qualifier and its class again by back-reference — so `?m@@3PEFRfoo@@DER1@` carries both that and the `__unaligned` in front of it, and the short form the other types use is not a name for it.
- **A name may carry no signature at all**, in which case the linkage is what is spelled: `?extern_c_func@@9` is `extern "C"`.

The fuzz corpus also caught `__unaligned` being *dropped* when it stood in front of a pointer into a class — three wrong spellings, now fixed.

| | at PR open | now |
|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | **524 / 609 (86%)** |
| corpus spelled differently | 0 | **0** |

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1236 passed / 1 skipped / 1721 subtests, diff-cover 100%.

r0ny123:
Every name in the reference corpus now reads: **609 of 609 spelled exactly as `llvm-undname` spells it, 0 given a different spelling.**

The last stretch was twelve grammar rules, each settled by probing the reference rather than reasoned about, because the corpus carries no name that exercises any of them:

- An attribute-spelled calling convention carries a space of its own in front of a declarator, where `__cdecl` and `__vectorcall` carry only the separator — `int (__attribute__((__swiftcall__))  *j)(int)` — while a convention spelled with nothing leaves no gap behind it at all.
- A member function's modifiers are written `__ptr64`, `__restrict`, `__unaligned`, then a reference qualifier, each at most once, and are spelled back in that same order. `QEIAA` is a name and `QIEAA` is not.
- A vcall is the slot it dispatches through: it carries neither a qualifier nor a convention of its own, and is never spelled with storage.
- `$$B` stands where an argument stands and nowhere a type may nest — not as a parameter, not as an array's element, not nested inside another argument — and never over a function type.
- An integer is an argument, so it is not an array's element either, and an array of no dimensions is not a type.
- No modifier stands in front of a function type or a member function pointer: `P8B@@` is a name, `PE8B@@` is not.
- How far a base class table reaches, and what it is flagged with, are unsigned — the third and fourth numbers of `??_R1` may not be negative.
- What a dynamic initialiser runs for is a name, and a digit there stands for an earlier name that does not exist.
- A conversion operator reads what it converts to from its return slot, so it is spelled with a signature and never with storage.

## The rendering bug behind several of them

Whether a declarator was a nested *function* was decided by looking for `(` in the rendered text. That mis-fires on any declarator whose **name** contains one of its own — an `operator()` name, or a local scope embedding a whole rendered signature — putting a space where the reference puts none. The producing site now says so explicitly instead of the consumer guessing from the text.

Sweeping for the same class found it a second time in the same file: an unanchored `"::*" in declarator` test, six lines above an anchored expression asking exactly the same question and carrying a comment about why the anchor is needed. A local-scope name whose enclosing signature takes a member-pointer parameter came out with a double space. Both sites now use the anchored form, and two names pin it.

## Measurements

| | at PR open | before this round | now |
|---|---|---|---|
| corpus spelled identically | 383 / 609 (63%) | 524 / 609 (86%) | **609 / 609 (100%)** |
| corpus spelled differently | 0 | 0 | **0** |
| fuzz wrong answers, 7 seeds / 110495 names | 31 | 57 | **2** |

The middle column is worth reading twice. Fuzz wrong answers went **up** from 31 to 57 while the corpus was going from 86% to 100%, because each new construct is a construct the demangler now answers — roughly a thousand more mutants read per seed — and every new answer is a new chance to diverge. A rising wrong count against a rising recall is not a regression, and treating it as one would have argued for stopping exactly where the remaining work was.

The 2 that remain are the divergences kept deliberately, both cases where the reference is the weaker side: `__int128`, which clang's own Microsoft mangler emits and `llvm-undname` cannot read back, and the vector-deleting-destructor family, which it spells with an empty operator name.

Every one of the 39 names added as tests here was cross-checked against `llvm-undname` in both directions — the expected spellings match it, and the refusals are refusals there too.

## Throughput

155k names/s cold, 27.6M/s through the cache, worst single name 161 µs. Demangling happens once per symbol at label-provider time, so this is far from the analysis hot path; the number that matters is the bound on hostile input, not the average.

Re-validated: ruff clean, `ty` exit 0, `pytest tests/` 1298 passed / 1 skipped / 2452 subtests, diff-cover 100%.

danielplohmann:
Superseded by #288 — rebased onto current master (PeSymbolProvider re-integrated over #287's Itanium demangling and #265's lief_name).

PR #256, opened 2026-08-17
fix(labels): keep ELF import names as the binary spells them

r0ny123:
## What was wrong

`ElfSymbolProvider.parseImports` demangled every name it returned, on the premise — stated when the demangling was moved there — that this method *"only feeds the label view"*. That premise does not hold.

The method feeds `BinaryInfo.getImportedFunctions`, and from there `xmetadata["imported_functions"]`, which has two consumers that need the name the binary actually carries:

- `BinarySynthesizer._getImportMap` passes it into import-table reconstruction in `PeSynthesizer`, `ElfSynthesizer` and `MachoSynthesizer`, where it becomes the *literal* name of an import entry. A synthesized ELF was carrying `operator new(unsigned long)` in a slot that holds an identifier.
- `SmdaReport._mergeImportedFunctions` fills the same dict from `disassembly.import_slots`, whose names come undecorated from the API resolver. So one dict held two spellings of the same import, keyed the same way, with the demangled one winning the merge.

PE and Mach-O never demangled here. That reads like the gap this change fills in, and is in fact the correct behaviour: an import name is an identifier — something to match against a DLL export table and to rebuild a binary with — not prose to read. ELF was the outlier.

## What this changes

`parseImports` returns `parse_elf_relocation_imports` unchanged, so all three formats now agree, and the docstring records why this slot must stay encoded rather than why it was demangled.

Nothing human-facing regresses. No display path reads this dict: the only other consumer, `X86Backend`, tests membership by address and never looks at the name. Function labels, exports and symbols are untouched and still demangled — `parseExports` and `_formatSymbolName` are unchanged, and a test pins that exports still demangle while imports no longer do.

## Tests

The test that pinned the old behaviour now pins the new one, and deliberately keeps its mocked `demangle_itanium_symbol` in place: the mock would return `bar()` if the code still consulted it, so asserting `_Z3barv` proves the demangler is not reached. `test_binary_info_elf_imported_functions_stay_encoded` is new and covers the path synthesis actually reads, at the `BinaryInfo` boundary rather than on the provider.

A second commit pins the same contract on the formats this change does not touch, and on the consumer that motivates it: a decorated MSVC import through `PeSymbolProvider.parseImports` and `BinaryInfo.getImportedFunctions`; a C++ Mach-O import through the provider's import map, `getApi` and the `BinaryInfo` view at once, which on Mach-O are one object; and a report carrying a decorated name in each of the three formats round-tripped through `synthesizeBinary`, asserting that exact name lands in the rebuilt import table. Nothing covered that path in either direction before.

The docstring on `parse_elf_relocation_imports` is corrected too — it still said demangling happens one level up.

## Validation

`ruff check .` and `ruff format --check .` clean; `ty` exits 0 with 229 diagnostics, unchanged from the base commit; `pytest tests/` 1185 passed, 1 skipped, 1480 subtests; `diff-cover --compare-branch=upstream/master --fail-under=100` reports 1/1 changed line covered.

### DIFF
--- a/src/smda/common/labelprovider/ElfSymbolProvider.py
+++ b/src/smda/common/labelprovider/ElfSymbolProvider.py
@@ -147,15 +147,15 @@ class ElfSymbolProvider(AbstractLabelProvider):
         return function_symbols
 
     def parseImports(self, lief_binary):
-        """Return relocation-backed imports with demangled, human-readable names.
+        """Return relocation-backed imports as the binary spells them.
 
-        Only this label view is demangled. ElfApiResolver consumes the undecorated
-        map so API references keep matching PE and Mach-O import names.
+        This is not a label view. It reaches xmetadata["imported_functions"], which
+        BinarySynthesizer writes back as the literal names of a reconstructed import table,
+        and which SmdaReport._mergeImportedFunctions fills from the undecorated API-resolver
+        view as well - so a demangled spelling here is both an unusable import name and a
+        second spelling of names already in the same dict.
         """
-        return {
-            address: (lib, demangle_itanium_symbol(name))
-            for address, (lib, name) in parse_elf_relocation_imports(lief_binary).items()
-        }
+        return parse_elf_relocation_imports(lief_binary)
 
     def collectSymbols(self, lief_binary):
         if not isinstance(lief_binary, lief.ELF.Binary):
--- a/src/smda/common/labelprovider/import_parsers.py
+++ b/src/smda/common/labelprovider/import_parsers.py
@@ -59,10 +59,10 @@ def parse_pe_delay_imports(lief_binary, base_addr=None):
 def parse_elf_relocation_imports(lief_binary):
     """Build import map keyed by relocation slot address: addr -> (lib, name).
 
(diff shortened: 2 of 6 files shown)


r0ny123:
Added coverage for the two formats this change does not touch, and for the consumer that motivates it.

A decorated MSVC import now keeps its spelling through `PeSymbolProvider.parseImports` and `BinaryInfo.getImportedFunctions`; a C++ Mach-O import keeps its spelling in the provider's import map, `getApi` and the `BinaryInfo` view at once (on Mach-O those are one object, so they could only drift together); and a report carrying a decorated name in each of the three formats round-trips through `synthesizeBinary` with that exact name present in the rebuilt import table. That last one is what makes the encoded spelling load-bearing rather than a preference — it was the gap called out above as uncovered in either direction.

Also corrected the docstring on `parse_elf_relocation_imports`, which still said demangling happens one level up.

Re-validated: ruff clean, `ty` exit 0 at 229 diagnostics (unchanged), `pytest tests/` 1185 passed / 1 skipped / 1480 subtests, diff-cover 1/1.

danielplohmann:
Superseded by #291 — rebased onto current master.

PR #257, opened 2026-08-17
feat(labels): demangle the MSVC names a PDB carries

r0ny123:
> **Stacked — do not merge before its dependencies.** This branch contains #245 and #255 (and through the latter, #249, #250, #254) so that it can compile at all. The subject of this PR is the **final commit only**, `feat(labels): demangle the MSVC names a PDB carries` — four files. Once the dependencies land, the diff here reduces to that commit on its own. It is a draft until then.

## Why this exists

#245 rebuilt the PDB provider on purepdb and left MSVC-decorated names decorated, saying so explicitly: *"there is no maintained, permissively-licensed MSVC demangler to substitute."* #255 then wrote one, and wired it into `PeSymbolProvider` only. This connects the two.

## Where it actually bites

Narrower than it sounds, and worth stating because it dictates the fixture. purepdb prefers a procedure record's name, which the compiler already writes undecorated — a fully `/Z7`-built image needs none of this. The decorated form reaches the name slot when a function has a **public record and no procedure record**: a translation unit built without debug info, a static library shipped without its PDB, `/DEBUG:FASTLINK`.

Analyzing the new fixture with `disassembleFile(exe, pdb_path=pdb)`:

| offset | before | after |
|---|---|---|
| `0x180001000` | `geometry::combine` | `geometry::combine` (proc record, untouched) |
| `0x180001064` | `?measure@text@@YAHPEBDI@Z` | `int __cdecl text::measure(char const *, unsigned int)` |
| `0x18000107f` | `?measure_export@@YAHPEBD@Z` | `int __cdecl measure_export(char const *)` |

## How it dispatches

On the leading `?`, ahead of the Rust evidence gate — the same key `PeSymbolProvider` uses. The two schemes cannot collide, since no Rust mangling begins with `?`. A name the demangler declines comes back byte-identical, and a fragment label built from a declined parent stays decorated with it. That set has shrunk as #255 progressed: local-scope names, lambdas and RTTI descriptors all read now, so what the tests here pin as declined is a function-local name with no type and a name truncated past its parameter list. Rust names, the undecorated-name path and the `getApi` contract are unchanged.

## Fixture

`tests/msvc_cxx_pdb_pe_xored` + `tests/msvc_cxx_pdb_xored`: a clang-cl / lld-link `x86_64-pc-windows-msvc` DLL built from two translation units, one compiled with `/Z7` and one without, so the PDB carries both name shapes at once. Built with `/GS- /GR- /O1` and linked `/DLL /NOENTRY /NODEFAULTLIB /DEBUG`, so it links against no CRT and no import library and involves no Microsoft-licensed material; the only stub needed is `extern "C" int _fltused = 0;`. Xored like the other binary fixtures. 80 KB for the pair, paired so the PDB can be checked against the image it describes.

Coverage added: the demangled public and the untouched procedure name off the real PDB, an assertion that no reported name stays decorated, report integration through `Disassembler`, and synthetic cases for the fragment labels and the four declined shapes.

## Also worth knowing

A whole-tree sweep of the class this belongs to — *a recovered name reaching a report slot without the demangler for its scheme* — found the function-label views consistent after this change, and one genuine defect in the opposite direction on the import side, filed separately as #256.

## Validation

`ruff check .` and `ruff format --check .` clean; `make typecheck` exits 0 with no new diagnostics; `pytest tests/` 1298 passed, 1 skipped, 2452 subtests; `diff-cover --fail-under=100` reports 100% of changed lines covered.


r0ny123:
Two additions from auditing the demangling paths end to end.

**A PDB name now has to be free of control characters to be reported.** A PDB string table holds whatever bytes were written into it, and every name travelled straight into the report — `bell\x07name` was stored and reported as-is, and as a fragment parent it would have carried the byte into every `$+0x…` label derived from it. The test is for control characters rather than for non-ASCII, because a demangled Rust identifier is legitimately non-ASCII and the printable-ASCII rule `PeSymbolProvider` applies to raw mangled names would discard real ones here. Empty names are skipped with them.

**The MSVC control-character guard** this branch picks up from #255, where the NUL check turned out to cover only NUL.

Also measured, while looking for recall gaps: on `ntdll.pdb` (XP, a real Microsoft-toolchain PDB) 141 decorated names appear, of which 137 are `??_C@…` string literals and 4 are functions — all 4 demangle. The string literals are correctly declined: llvm-undname spells every one of them `` `string' ``, so expanding them would give 137 symbols the same name and destroy the identity a decorated name still carries. They are data publics with no code size, and `addSymbolCandidate` already rejects them by code area, so they seed nothing.

Re-validated: ruff clean, `make typecheck` exit 0, `pytest tests/` 1259 passed / 1 skipped / 1597 subtests, diff-cover 470/470.

r0ny123:
**Correcting something I asserted above.**

That comment says the `??_C@…` string literals in `ntdll.pdb` are "correctly declined", on the grounds that `llvm-undname` spells every one of them `` `string' `` and expanding them would give 137 symbols one shared name. **That is wrong.** The reference decodes a string literal to its *contents* — `??_C@_02DKCKIIND@hi?$AA@` is `"hi"`, not `` `string' ``. So the argument I gave for declining them was backwards: expanding them makes those names **more** distinct, not less.

I had the reasoning inverted for several turns and wrote it down before re-checking it against the reference. The demangler now reads string literals properly, in the stack this branch sits on: the length counts the terminator, the eight characters after it are a hash of the bytes rather than part of them, and the bytes themselves are written plainly or as an escape — a digit standing for one of ten punctuation characters, or `$` and two nibbles for any byte at all.

The part of that paragraph that still holds: they are data publics with no code size, and `addSymbolCandidate`'s code-area filter already rejects them, so they seed no function candidates either way. The recall claim — 4 of 4 decorated **function** names on `ntdll.pdb` — is unaffected.

**What changed on this branch itself.** With the demangler underneath now reading the whole reference corpus (609 of 609, 0 spelled differently — see #255), an RTTI type descriptor reaching this provider demangles rather than coming back decorated. `??_R0?AVexception@std@@@8` reads as ``class std::exception `RTTI Type Descriptor'``, and a PDB carries one of those per polymorphic class. The name moves from the list asserting it comes back untouched to one asserting the spelling.

Re-validated on the merged branch: ruff clean, `ty` exit 0, `pytest tests/` 1298 passed / 1 skipped / 2452 subtests, diff-cover 100%.

danielplohmann:
Superseded by #293 — dependencies landed as #288/#289, rebased on top.

PR #258, opened 2026-08-17
perf(labels): demangle a Mach-O's Swift names in one call

r0ny123:
## What was wrong

Swift demangling shells out to the host toolchain, and it did so **once per symbol**. A Mach-O that carries Swift carries it by the hundred, so process spawns dominated symbol recovery — 26 ms each, and 104 distinct Swift names in the `jokerspy` corpus sample alone. Three of the twelve corpus samples are affected; `jokerspy` is 181 Swift symbols, `osx.frostyferret` 94, `osx.hloader` 33.

## What this changes

`MachoSymbolProvider.update` hands every symbol-table name to `primeSwiftSymbols` before the per-symbol lookups begin, and one process answers the whole set.

| sample | before | after |
|---|---|---|
| `objective-see/jokerspy` | 2.28s | 0.03s |
| `malpedia/osx.frostyferret` | 1.74s | 0.02s |
| `malpedia/osx.hloader` | 0.26s | 0.02s |
| **whole 12-sample corpus** | **4.29s** | **0.08s** |

The recovered `{address: name}` map is **identical sample for sample** — 2875 named symbols across the corpus, none spelled differently, none lost. This is a scheduling change, not a behaviour change.

## Details that matter

Names go in over **stdin rather than argv**, because a symbol table can hold more names than a command line can carry.

The reply is **positional**, so a short or padded one is discarded rather than realigned — taking it would label symbols with each other's names. A name containing a newline cannot be expressed in a line-per-name protocol, so it is excluded from the batch and left to the single-name path.

A host with **no Swift toolchain** — which is most Linux CI — now pays one failed lookup for the whole binary instead of one per symbol. The existing single-name path is unchanged for callers that reach the demangler directly.

## Validation

`ruff check .` and `ruff format --check .` clean; `ty` exits 0 with 229 diagnostics, unchanged from the base commit; `pytest tests/` 1191 passed, 1 skipped, 1480 subtests; `diff-cover --compare-branch=upstream/master --fail-under=100` reports 45/45 changed lines covered.

New tests cover the one-process guarantee, the discarded misaligned reply, a non-zero exit, an unrunnable demangler not being retried per symbol, the newline exclusion, a host with no toolchain, and both symbol-table sources feeding the batch.

### DIFF
--- a/src/smda/common/labelprovider/MachoDemangler.py
+++ b/src/smda/common/labelprovider/MachoDemangler.py
@@ -13,6 +13,66 @@ _CXX_PREFIXES = ("__Z", "_Z")
 _SWIFT_PREFIXES = ("_$s", "$s", "_$S", "$S")
 
 
+_SWIFT_CACHE = {}
+
+
+def primeSwiftSymbols(names):
+    """Resolve every Swift name in one call, before the per-symbol lookups begin.
+
+    The Swift demangler is a subprocess, and a Mach-O carrying Swift carries it by the
+    hundred: 104 distinct names in the jokerspy corpus sample, at 26 ms of process spawn
+    each. Fed the whole set through stdin it answers in 20 ms, and a host with no Swift
+    toolchain costs one failed lookup instead of one per symbol.
+
+    stdin rather than argv, because a symbol table can hold more names than a command
+    line can carry.
+    """
+    pending = sorted(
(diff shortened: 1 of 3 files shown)


danielplohmann:
Superseded by #292 — rebased onto current master.

PR #259, opened 2026-08-17
fix(core): bound a synthesized section by the image limit

r0ny123:
## What was wrong

A report's `code_sections` carry the extents synthesis lays out, and the span between them is what gets allocated. Nothing bounded it. CI's fuzzer found a **611-byte report** whose single section claimed `0x13DAF6E5B0` bytes — synthesis laid it out at that size and was killed at the 4 GB rss limit, having reached **10 GB**.

```json
"code_sections":[["",4,85314600944]]
```

## Why it slipped through

The bound already exists one level up, and on both sides of it: `BinarySynthesizer._resolveFunctionOffsets` refuses a *function-offset* span wider than `MAX_IMAGE_SIZE`, and `PeFileLoader`/`ElfFileLoader` refuse an image larger than it. A section extent read straight from the report reached the layout without passing either.

## What this changes

Such a section is dropped, the way an inverted or empty one already is, with a warning naming the span.

Both formats that lay sections out this way needed it: ELF is where the fuzzer landed, and Mach-O builds from the same field with the same shape. PE does not — it derives its sections from the planted regions instead, so it is untouched.

Peak memory for the reproducer goes from **over 7 GB to 93 MB**.

## Tests

The minimized reproducer is pinned as `tests/fuzz_regressions/synthesis_oversized_section_span_xored` and replays through the existing harness. A unit test covers both formats directly, since the pinned artifact only selects ELF.

## Validation

`ruff check .` and `ruff format --check .` clean; `ty` exits 0 with no new diagnostics; `pytest tests/` 1182 passed, 1 skipped, 1483 subtests; `diff-cover --compare-branch=upstream/master --fail-under=100` reports 8/8 changed lines covered.

Note on the fuzzing job, updated: when this was opened the crashing input sat in the shared corpus cache, so `Fuzz (fuzz_synthesis)` failed on every branch that restored it, including branches unrelated to synthesis. That is no longer the case — the input is no longer in the cache being restored, and the job passes on this branch and on the other open ones alike. The bug it found is real either way, and the reproducer is pinned here so it is replayed deliberately rather than depending on what a cache happens to hold.


danielplohmann:

PR #266, opened 2026-08-20
fix(core): let the gap scan give up when the analysis budget is spent

r0ny123:
## What

Both gap scans (`intel/FunctionCandidateManager.nextGapCandidate` and its AArch64 counterpart) walk forward from the gap pointer until they find a candidate or run out of image, and neither consulted the analysis timeout while doing it. The loop advances one byte (intel) or one instruction word (AArch64) per iteration over anything it cannot use, so a large claimed data run inside the code area is walked in full even after the budget for the whole run is gone. The timeout is enforced between functions, so the scan itself was the one unbounded stretch left in candidate discovery.

Both loops now poll every 4096 iterations — the same interval and shape `_splitMergedFunctionsAtPdataEnds` already uses — and return "no further candidate" once the budget is spent.

## Cost when nothing is wrong

One counter increment and one modulo per iteration; the callback is not reached at all on a scan that finishes normally.

## Scope note

This bounds a latent overrun, not a measured one: the analysis overrun that prompted the audit turned out to be caused by a wrong bitness probe on mapped 64-bit images (separate change), which inflates candidate counts rather than lengthening a single scan. Measured on the bundled fixtures, no recovered function set changes.

## Tests

`tests/testGapScanTimeout.py`, four cases: for each backend a long scan that completes normally (positive control, so the timeout case cannot pass vacuously) and the same scan with a budget that is already spent, asserting the scan returns no candidate, marks `analysis_timeout`, and calls the callback exactly once.

## Validation

- `make test` — 1184 passed, 2 skipped
- `diff-cover` — 100% on 8 changed lines

### DIFF
--- a/src/smda/aarch64/FunctionCandidateManager.py
+++ b/src/smda/aarch64/FunctionCandidateManager.py
@@ -764,7 +764,11 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
         def in_exec(addr):
             return any(start <= addr < end for start, end in exec_ranges)
 
+        scanned = 0
         while True:
+            scanned += 1
+            if scanned % 4096 == 0 and self._candidateTimeoutTripped():
+                return None
             if base + size < self.gap_pointer:
                 return None
             # align to the instruction stride
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -195,7 +195,11 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
             window_bytes = self.disassembly.getRawBytes(offset, max(256, length))
             return window_bytes[:length]
 
+        scanned = 0
         while True:
+            scanned += 1
+            if scanned % 4096 == 0 and self._candidateTimeoutTripped():
+                return None
             unskipped_pad = False
             if self.disassembly.binary_info.base_addr + self.disassembly.binary_info.binary_size < self.gap_pointer:
                 LOGGER.debug("nextGapCandidate() gap_ptr: 0x%08x - finishing", self.gap_pointer)
--- a/tests/testGapScanTimeout.py
+++ b/tests/testGapScanTimeout.py
@@ -1,0 +1,85 @@
+import unittest
+
+from smda.aarch64.definitions import INSTRUCTION_SIZE
+from smda.aarch64.FunctionCandidateManager import FunctionCandidateManager as AArch64CandidateManager
+from smda.common.BinaryInfo import BinaryInfo
+from smda.DisassemblyResult import DisassemblyResult
+from smda.intel.FunctionCandidateManager import FunctionCandidateManager as IntelCandidateManager
+from smda.SmdaConfig import SmdaConfig
+
+BASE = 0x400000
+# Long enough that a single scan crosses the timeout-check interval twice.
+SKIPPED_BYTES = 4096 * 8
+AARCH64_NOP = b"\x1f\x20\x03\xd5"
+
+
+def _disassembly(image, bitness, architecture):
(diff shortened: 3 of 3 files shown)


r0ny123:
On the benchmark's "PR is slower, -15.53%" verdict: I do not think that is this change, and here is the evidence rather than an assertion.

**What the change costs per iteration.** One counter increment and one modulo. The callback is only reached on the 4096th iteration, and only then when a scan is that long — a normal scan never reaches it.

**Local A/B, same machine, interleaved.** Each fixture is run on both trees back to back, three repetitions, per-fixture minimum taken, so machine drift lands on both sides equally:

| | master | this branch | delta |
| --- | --- | --- | --- |
| Sum over 22 fixtures | 7.664s | 7.521s | **-1.86%** |
| `rust_pe_gnu` (5.0s of that total) | 5.037s | 4.705s | -6.57% |

Per-fixture spread is ±20% on the millisecond-scale fixtures, which is what a shared machine looks like; the aggregate is inside the noise and, if anything, negative.

**Why the CI number is hard to read.** base and PR are timed on separate runners, as the report itself says. The same harness ran twice against #265, where the second run added only a test file and left the analysed code identical, and reported **+28.62%** and then **+10.92%** — an 18-point swing on an unchanged comparison, against a stated noise band of ±5.9% and ±4.9%. Cross-runner variance on this workload appears to be well outside the band the report computes from repeated runs on one side.

The parts of the report that are not timing-based both pass and are the ones that matter here: **correctness 0/153 files differ**, determinism PASS. This change is expected to alter nothing about the output, and it does not.

r0ny123:
This branch adds two integer operations per iteration of the gap scan, which is the hottest loop in candidate discovery, so it is the one change in this series where a cost is plausible rather than impossible. The CI job reported it at -16.79%. Measured locally instead, since that job's verdict is not usable (see #268 for the A/A pair that shows why).

**Method.** All variants run in one process-spawn loop on one machine, order rotated each round, timed with `time.process_time()` rather than wall clock. 22 bundled fixtures, `TIMEOUT=0`, `WITH_STRINGS=False`, `CALCULATE_HASHING=False`. Two null controls run alongside: `masterB` is master measured a second time under a different name (a true A/A), and #272 changes only comments and prose so its runtime code is byte-identical to master. Whatever those two report is the floor.

**Median paired per-fixture ratio vs master:**

| run | rounds | masterB (A/A) | #272 (null) | **#266** |
| --- | --- | --- | --- | --- |
| 1 | 7 | +0.91% | -1.20% | **+3.48%** |
| 2 | 14 | +0.47% | +0.26% | **+3.08%** |
| 3 | 14 | +1.24% | +1.09% | **+1.44%** |

Runs 1 and 2 put this branch about 2.5 points above both controls, which looked like a real cost. So I tested the obvious cause directly: replaced `scanned % 4096` with a countdown (`until_poll -= 1; if not until_poll:`), which removes the modulo from every iteration, and measured it as its own variant in run 3.

It came out at **+1.41%** against this branch's **+1.44%** — no difference. If the modulo were the cost, removing it would have shown. It did not, so the hypothesis is refuted and I am not shipping that change; a rewrite that measures identical is churn, not an optimization.

In run 3 this branch also sits inside the null-control band, so the honest conclusion is that the cost is **below this harness's resolution** (roughly ±1.5-3% on the paired median, judging by how far the two null controls move between runs), not that it is exactly zero. Totals across the three runs were +1.32%, +2.19% and -0.03%, straddling zero.

Function counts were identical to master on all 22 fixtures in every run, in every variant.

For context, the same harness priced the rest of the series the same way, all within the null-control band: #265 +0.84%, #267 +3.03%, #268 +2.02%, #269 -1.55%, #270 +0.53%, #271 -2.66%, #273 +0.70%.

No change to this branch.

PR #268, opened 2026-08-20
fix(intel): read a mapped image's bitness from its REX.W usage

r0ny123:
## What was wrong

`BitnessAnalyzer.determineBitness` collected the first bytes at every direct-call target and scored them against the 32- and 64-bit function-start byte tables. Those tables overlap heavily — both list push/mov/sub prologues — so the vote is decided by which table happens to hold more of the common entries, not by anything that actually differs between the two modes.

This is a memory-dump problem specifically. A file carries its machine type in a header; a dump carries nothing, so the probe is the only answer available. And reading a 64-bit image as 32-bit does not fail: it decodes every REX prefix as `dec`/`inc`, splits real instructions, and returns a full report whose blocks and edges are wrong.

## The discriminator

`0x48` is REX.W in 64-bit mode and `dec eax` in 32-bit mode. In 64-bit code it is followed by one of a small set of opcodes almost every time (the pointer arithmetic, moves and compares that dominate an amd64 text section); in 32-bit code the byte after it is unrelated, so the share lands near the 29/256 a uniform byte would give.

Sampling is restricted to the loader's code areas when it named any, because data dilutes the share toward the uniform one — measured, the clang-built C binary scores 0.993 over its code and 0.744 over the whole image. A raw dump names no code areas and is judged whole; a named area too small to sample falls back to the whole image rather than straight to the byte tables.

Measured over PE/ELF/Mach-O images from Go, gcc, clang, mingw-Rust and 32-bit malware, plus the bundled fixtures:

| sampled over | 32-bit max | 64-bit min |
| --- | --- | --- |
| code areas | **0.351** | **0.937** |
| whole image | 0.245 | 0.744 |

The 0.5 threshold sits inside both gaps, which is what lets one threshold serve both sampling modes. Four other statistics were measured first and rejected: frequency-weighted tables (6/14 correct), the REX ratio at call targets (9/13), decode length (4/13), sane termination (10/13).

Below 64 observed `0x48` bytes the share is noise and the old table vote still decides; sampling stops at 65536, far past the point where the share stabilises.

## End to end

Analysing the bundled x64 ELF as a mapped image, with no bitness supplied:

| | bitness | functions |
| --- | --- | --- |
| master | 32 | 171 |
| this branch | 64 | 152 |

The higher count was the wrong one — those are fragments of instructions split at REX prefixes.

Re-checked across the whole measurement corpus after the code-area change: 21 of 21 images with a known bitness classify correctly.

The engine also announces an inferred bitness as a warning rather than a debug line: a caller that could have supplied it has no other sign that it was guessed.

## Tests

`tests/testBitnessProbe.py` — 18 cases: the share statistic itself (pairs, `dec eax` followers, sample floor, sampling cap, trailing prefix byte, mixed share), code-area sampling (a diluted image the whole-buffer share reads as 32-bit, area clamping, empty areas, the too-few-samples fallback, an image with no areas), the bundled fixtures mapped the way a dump reaches SMDA, end-to-end recovery from a mapped 64-bit image, and the inferred-bitness warning in both directions.

## Validation

- `make test` — green
- `diff-cover` — 100% on changed lines

### DIFF
--- a/src/smda/common/RecursiveDisassembler.py
+++ b/src/smda/common/RecursiveDisassembler.py
@@ -434,8 +434,10 @@ class RecursiveDisassembler:
         self.disassembly.analysis_start_ts = datetime.datetime.now(datetime.timezone.utc)
         if binary_info.bitness not in [32, 64]:
             binary_info.bitness = self.backend.probeBitness(self.disassembly)
-            LOGGER.debug(
-                "Automatically Recognized Bitness as: %d",
+            LOGGER.warning(
+                "No bitness was supplied; inferred %d-bit from the buffer's contents. "
+                "Pass the bitness explicitly when it is known - decoding in the wrong mode "
+                "yields a plausible-looking report whose blocks and edges are all wrong.",
                 binary_info.bitness,
             )
         else:
--- a/src/smda/intel/BitnessAnalyzer.py
+++ b/src/smda/intel/BitnessAnalyzer.py
@@ -20,7 +20,10 @@ REX_W_PREFIX = 0x48
 # every 0x48 introduces one of the opcodes above; in 32-bit code the byte that follows
 # is unrelated, so the share lands near the 29/256 a uniform byte would give. Measured
 # across PE/ELF/Mach-O images from Go, gcc, clang, mingw-Rust and 32-bit malware:
-# 32-bit stays at or below 0.36, 64-bit at or above 0.74.
+# over the loader's code areas, 32-bit stays at or below 0.36 and 64-bit at or above
+# 0.93; over a whole image, where data dilutes the count toward the uniform share, the
+# floor for 64-bit drops to 0.74. The threshold sits inside both gaps because a raw
+# memory dump names no code areas and has to be judged whole.
 REX_W_SHARE_THRESHOLD = 0.5
 # Below this many observations the share is noise; fall back to the first-byte tables.
 REX_W_MIN_SAMPLES = 64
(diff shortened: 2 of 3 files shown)


r0ny123:
This branch just produced an accidental A/A test of the benchmark job, which is worth recording because it bounds what the numbers can be used for.

The same tree was benchmarked twice tonight. The only thing that changed between the runs was the commit author string — the branch was force-pushed to correct it, and `git diff 26d1982 a9ec464` is empty, so both runs analysed identical bytes with identical code.

| run | verdict | median paired speedup | 95% CI | noise band | Wilcoxon p |
| --- | --- | --- | --- | --- | --- |
| 20:16 (`26d1982`) | inconclusive | **-3.12%** | [-4.31%, -2.32%] | ±4.1% | 0.0000 |
| 03:20 (`a9ec464`) | **PR is faster** | **+27.88%** | [+27.27%, +28.47%] | ±3.1% | 0.0000 |

A 31-point swing on unchanged code, and the second run reports it as a significant *improvement* with a confidence interval barely a point wide and p = 0.0000.

The mechanism is the one the report already discloses — base and PR run on separate runners — but the consequence is worth being explicit about: the confidence interval and the p-value are computed from within-side repeat variance, which measures how consistent each runner is with itself. Cross-runner offset never enters the estimate, so a large constant offset between the two machines produces a tight interval around the wrong number. Repeating runs narrows the interval without moving it toward zero, which is exactly what makes the output look authoritative when it is not.

That matches what the other PRs in this series show. #272 changes only `README.md` and comment lines — no executable change at all — and was reported at -59.80%. #271 deletes a function with no call sites and was reported at -19.74%. Across the series the verdicts have ranged from -59.80% to +27.88% with no diff capable of producing any of it.

Two things worth separating, because the job gets one of them right:

- **Correctness and determinism are trustworthy and stayed clean throughout.** 0 of 153 files differ, 153/153 function sets match in every pairwise comparison, both sides deterministic. Those are not timing-based and they did their job.
- **The timing verdict is not usable as a gate in its current form.** Making it so would mean timing both sides on the same runner, or reporting a paired difference against an A/A baseline measured the same way, so that runner offset cancels.

No change to this branch on the strength of it. Flagging it because a green "PR is faster" is as misleading here as a red one, and the next person to read one of these will not have the A/A pair sitting next to it.

PR #270, opened 2026-08-20
fix(intel): scan a direct jump table whose bound was not recovered

r0ny123:
## What was wrong

Of the three jump-table extractors, the direct one — the 32-bit `mov <reg>, dword ptr [<reg>*4 + table]` / `jmp <reg>` shape — was missing two things its siblings already had, plus a third they use to keep the gap scan out of the table.

1. **An unrecovered bound abandoned the table.** `_findJumpTableSize` reports "not found" as `0`, not `None`, and `if jumptable_size and ...` read that as *zero entries*. This is the same misreading `_resolveExplicitTable` was fixed for; here it means a switch whose `cmp` the backtracker cannot reach contributes no targets at all.
2. **Nothing bounded the entries.** Every dword in the scanned range became a jump target; only the caller's later in-image test kept the bogus ones out of the block queue. Both siblings stop at the first entry outside the image.
3. **The table bytes were not claimed as data**, so the gap scan can seed function candidates inside them — the reason `_extractRelativeTableOffsets` claims its own.

## Effect

On a 32-bit switch with no recoverable bound:

| | functions | blocks in the dispatcher |
| --- | --- | --- |
| master | 4 (dispatcher + each switch body as its own function) | 1 |
| this branch | **1** | **4** |

The switch bodies are unreferenced without the table, so the gap scan recovers each as a function of its own — a wrong CFG that looks like a plausible one.

All 22 bundled fixtures are byte-identical (function, block and instruction counts, and the recovered start sets). That is not because the change is inert: the direct extractor is reached 37 times across those fixtures and returns nothing every time, because none of them carries this table shape with a resolvable base. Hence the constructed case rather than a fixture.

## Scope note

The sibling `continue`s in `_extractRelativeTableOffsets` are deliberately untouched — they are the subject of the fix already sitting on `fix/intel-jumptable-entry-bound`, and duplicating it here would only create a conflict.

## Tests

`tests/testJumpTableAnalyzer.py` — six unit cases for the extractor (unrecovered bound, out-of-image stop, a real bound still capping the scan, short read, data claims, and `getJumpTargets` forwarding the state), plus one end-to-end case building the 32-bit switch above and asserting one function with four blocks and the three consumed entries claimed as data. Five of the seven fail on master.

## Validation

- `make test` — 1187 passed, 2 skipped
- `diff-cover` — 100% on 9 changed lines

### DIFF
--- a/src/smda/intel/JumpTableAnalyzer.py
+++ b/src/smda/intel/JumpTableAnalyzer.py
@@ -119,15 +119,25 @@ class JumpTableAnalyzer:
                 break
         return bonus_offset
 
-    def _extractDirectTableOffsets(self, jumptable_size, off_jumptable):
+    def _extractDirectTableOffsets(self, jumptable_size, off_jumptable, state=None, jump_instruction_address=None):
+        # an unrecovered bound arrives as 0, which read as "zero entries" and abandoned the
+        # table; the per-entry image bound below is what actually ends the scan, exactly as
+        # in _resolveExplicitTable and _extractRelativeTableOffsets
+        jumptable_size = jumptable_size if jumptable_size else 0xFF
         jump_targets = set()
-        if jumptable_size and off_jumptable and self.disassembly.isAddrWithinMemoryImage(off_jumptable):
+        if off_jumptable and self.disassembly.isAddrWithinMemoryImage(off_jumptable):
             for index in range(jumptable_size):
                 raw_entry_bytes = self.disassembly.getBytes(off_jumptable + index * 4, 4)
                 if raw_entry_bytes is None or len(raw_entry_bytes) < 4:
-                    continue
+                    break
                 entry = struct.unpack("<I", raw_entry_bytes)[0]
+                if not self.disassembly.isAddrWithinMemoryImage(entry):
+                    break
                 jump_targets.add(entry)
+                if state is not None:
+                    # claim the entry as data, or the gap scan seeds function candidates
+                    # inside the table bytes
+                    state.addDataRef(jump_instruction_address, off_jumptable + index * 4, size=4)
         return sorted(jump_targets)
 
     def _extractRelativeTableOffsets(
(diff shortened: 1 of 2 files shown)


r0ny123:
## The other half of this area, measured but deliberately not in this PR

`_findJumpTableSize` takes the first `cmp <reg>, <imm>` it finds walking back from the jump, without checking that the register is the one indexing the table — the file's own comment says so. That is a separate defect from the three this PR fixes, and it is worth recording what it actually costs before anyone rebuilds the resolver for it.

**How often it fires.** Instrumenting every indirect-jump dispatch across the bundled fixtures and classifying each recovered bound by whether the compared register is in the def-use closure of the jump's operand:

| fixture | bound tied to the index | untied | untied *and* producing targets | no bound found |
| --- | --- | --- | --- | --- |
| `rust_pe_gnu` | 177 | 27 | **14** | 113 |
| `mirai_x64` | 4 | 0 | 0 | 3 |
| `mirai_i386` | 3 | 0 | 0 | 2 |
| `bashlite` | 1 | 0 | 0 | 4 |
| `cutwail`, `komplex`, `pe_export_label_test` | 0 | 0 | 0 | 0 |

So it is real, and confined to one fixture here: 14 of 210 bounded dispatches.

**What tying it would change.** Replacing the bound with one that only accepts a `cmp` whose register reaches the jump through a backward def-use chain, and diffing the recovered function set:

| | functions | added | lost |
| --- | --- | --- | --- |
| `rust_pe_gnu` | 2354 → **2359** | 6 | 1 |
| every other intel fixture | unchanged | 0 | 0 |

**Why it is not in this PR.** A 0.25% move on one fixture, with no ground truth in the repository to say whether the six added starts are real or the one lost start was, is exactly the kind of change that should be decided on a labelled corpus rather than on a count going up. The same reasoning applies here as to the alignment-floor comment a few files away.

The three fixes that *are* in this PR are the ones that need no such judgement: an abandoned table, an unbounded entry scan and unclaimed table bytes are wrong regardless of corpus.

Harness for both measurements is available if useful — instrument `getJumpTargets`, classify the chosen `cmp`, then monkey-patch `_findJumpTableSize` and diff the reports.

PR #271, opened 2026-08-20
refactor(core): drop the unused indirect-switch walker

r0ny123:
## What

`RecursiveDisassembler.resolveIndirectSwitch()` walks the bytes that follow a jump table, reading them as case indices — the two-level MSVC switch, where a byte array indexes a smaller table of distinct targets.

It has had **no caller since it was introduced in 2020**, so it has never run in any analysis.

## Why delete rather than wire it up

Because there is nothing here to validate a wiring against. The shape it looks for was measured across the bundled fixtures instead of assumed: of the eight explicit jump tables the corpus resolves, **none** is followed by a byte run that could be an index array. (Three have some trailing bytes below the entry count, but none as long as the count, and zero padding passes that test trivially.)

A real implementation of the two-level form needs a fixture that exhibits the pattern and a measurement showing what it recovers — which is the whole of the work, with or without this text in the tree. An untested helper sitting in the architecture-neutral engine is a liability rather than a head start.

Its only exercise was a test of its own behaviour, which goes with it.

## Validation

- `make test` — 1179 passed, 2 skipped (one fewer than master: the test of the deleted method)
- `diff-cover` — no added lines to cover

### DIFF
--- a/src/smda/common/RecursiveDisassembler.py
+++ b/src/smda/common/RecursiveDisassembler.py
@@ -166,31 +166,6 @@ class RecursiveDisassembler:
             return -value if referenced_addr.group("sign") == "-" else value
         return 0
 
-    def resolveIndirectSwitch(self, addr_switch_array, size):
-        indirect_switch_bytes = []
-        current_offset = addr_switch_array + size * 4
-        if self.disassembly.isAddrWithinMemoryImage(current_offset):
-            LOGGER.debug(
-                "0x%08x analyzing potentially indirect switch table (size: 0x%08x).",
-                current_offset,
-                size,
-            )
-            current_byte = self.disassembly.getByte(current_offset)
-            if isinstance(current_byte, str):
-                current_byte = ord(current_byte)
-            while (
-                current_byte is not None
-                and current_byte < size
-                and current_offset not in self.fc_manager.getFunctionStartCandidates()
-            ):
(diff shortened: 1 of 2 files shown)


r0ny123:
The "PR is slower / -19.74%" verdict is not attributable to this diff, and I am not changing anything in response to it.

This branch is 39 deleted lines and zero added ones: `RecursiveDisassembler.resolveIndirectSwitch` and the single unit test that called it. On the base commit, `git grep resolveIndirectSwitch` over `src/` and `tests/` returns exactly two hits — the definition and that test. There is no production call site, so no executed instruction differs between base and PR. A removed function that was never invoked cannot make the analyzer 19.7% slower in either direction.

The report's own data agrees on behaviour: 0 of 153 files differ, and the function set matches 153/153 in all nine pairwise run comparisons.

What the number does reflect is the harness comparing two sides that ran on separate CI runners, which the report notes itself. The per-side determinism figures show a 5.2% median timing CV on base alone, and the pairwise matrix spreads from -9.47% to -20.96% depending purely on which raw runs are paired — a 11.5-point spread for one fixed pair of trees. The same harness has produced double-digit deltas on this repository for diffs with no runtime effect.

If a genuine performance question comes up on this branch later, the way to answer it is an interleaved A/B on one machine rather than a cross-runner comparison, which is how the timing claim on #266 was resolved.

PR #272, opened 2026-08-20
docs(core): record what RESOLVE_TAILCALLS and TIMEOUT actually mean

r0ny123:
## What was missing

`RESOLVE_TAILCALLS` ships **off**, and neither the README nor the config said so or said what turning it on is worth.

With it off, `TailcallAnalyzer` accumulates nothing and promotes nothing; the only surviving behaviour is the always-on rule that a jump into an already-recovered function ends the current one sanely. That is a defensible cost trade, but the README's CFG section reads as though tailcall handling is simply present, and a reader has no way to tell which half they are getting.

`TIMEOUT` has a related gap: a run that hits it returns `status == "timeout"` and a function set that is a **lower bound**, and a caller comparing counts across samples has no other signal that one of them was cut short.

## Measured, not described

| | functions | analysis time |
| --- | --- | --- |
| `libstdc++.so.6` | 8473 → **8810** (+337) | +40% to +130% |
| Go ELF memory image | 1936 → 1944 (+8) | +21% |
| Go Mach-O memory image | 2123 → 2131 (+8) | +26% |

The time range for `libstdc++` is deliberately a range: it is wall-clock, and the two measurements that produced it differed by how much the machine had to spare. Quoting a single number there would be quoting the machine.

## Changes

Config comment next to the flag, and a README paragraph next to the existing performance-lever section — the two places a reader already looks for this.

## Validation

- `make test` — 1180 passed, 2 skipped
- `diff-cover` — no executable lines changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,16 +107,23 @@ nesting 4.6% and SCC 3.2% of all calls, and disabling all three removes 17.8%. T
 a downstream consumer does not read.
 
 `RESOLVE_TAILCALLS` runs the other way round: it is **off** by default and buys recovery for time.
-It promotes the target of a jump that leaves a function into a function of its own, in a pass after
-gap analysis, so what it is worth depends on how much a binary tail-calls. On `libstdc++.so.6` it
-adds 337 functions (8473 to 8810) for 40-130% more analysis time, the spread being how much the
+Only the x86/x64 and AArch64 recursive disassemblers read it — the CIL and Dalvik backends run their
+own analysis pipelines and ignore it, so enabling it costs and buys nothing there. Where it does
+apply, it promotes the target of a jump that leaves a function into a function of its own, in a pass
+after gap analysis, so what it is worth depends on how much a binary tail-calls. On `libstdc++.so.6`
+it adds 337 functions (8473 to 8810) for 40-130% more analysis time, the spread being how much the
 measuring machine had to spare; on Go ELF and Mach-O memory images it adds 8 functions each, for
 20-26%. A jump into an already-recovered function is treated as a tailcall either way — only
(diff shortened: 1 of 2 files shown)


r0ny123:
Recording this one because it is a useful calibration point for the benchmark job rather than a finding about this branch.

This PR changes two files: `README.md`, and comment lines in `SmdaConfig.py`. There is no executable change of any kind — `git diff master...HEAD -- '*.py'` touches only lines beginning with `#`. Python discards comments at compile time, so base and PR run byte-identical bytecode.

The job reported **"PR is slower", median paired speedup -59.80%, 95% CI [-64.02%, -57.31%], Wilcoxon p = 0.0000, n = 153**, with a noise band of ±6.5%.

A statistically significant 60% regression from a diff that changes no code means the test is measuring the runners, not the branch. The pairwise matrix in the same report shows base medians of 0.5015s / 0.5172s / 0.4735s against PR medians of 0.8125s / 0.7802s / 0.7505s — the two sides are simply running on machines of different speed, and repeated sampling on each side tightens the confidence interval around that difference rather than around any property of the code.

The report's own caveat covers this ("base and PR are timed on **separate** CI runners"), but the headline verdict is computed from the paired difference and the noise band is estimated from within-side variance only, so cross-runner offset never enters it. That is why a null diff can land far outside the band. Two related data points: #271 is a 39-line pure deletion of a function with no call sites and was reported at -19.74%; #273 leaves both new options off by default and was reported at -42.09%.

Correctness and determinism are unaffected by this and remain meaningful — 0 of 153 files differ, 153/153 function sets match in every pairwise comparison, both sides deterministic.

No change to this branch.

PR #273, opened 2026-08-20
feat(common): read linker function tables as candidates on both instruction sets

r0ny123:
Compilers and linkers already record where functions begin, and SMDA was using almost none of it. This picks up two of those tables, both behind flags that default to off.

## ELF `.eh_frame`, on x86/x64 as well as AArch64

`EhFrameDecoder` already decoded frame description entries into `(start, length)` pairs, but only `aarch64/FunctionCandidateManager.py` ever called it. An x86 or x64 ELF got nothing from a table its own compiler had emitted.

The decode, the candidate promotion and the Mach-O slice handling move to the arch-neutral `common/FunctionCandidateManager.py`, and both backends call them. The AArch64 subclass keeps only what is genuinely architectural — its 4-byte instruction alignment, now expressed as a `CANDIDATE_ALIGNMENT` class attribute that defaults to 1.

## `.eh_frame_hdr`, so a dumped ELF can use it too

`.eh_frame` is found through the section header table, which a memory dump does not carry. Program headers do survive, and `PT_GNU_EH_FRAME` points at `.eh_frame_hdr`, whose binary search table lists the same function starts as 4-byte offsets relative to its own address. Reading that table gives a dumped ELF starts it would otherwise lose entirely. The table is looked for both at its recorded virtual address and at that address rebased on the image, since a dump can carry either.

Those header fields are read little-endian, so the parser checks `EI_DATA` and refuses a big-endian image rather than decoding it into byte-swapped offsets that address unrelated bytes. That check is not hypothetical: a big-endian OpenRISC binary was reading its program header offset as `0x34000000` before it went in.

## Mach-O `LC_FUNCTION_STARTS`

Nothing read this at all. The linker writes it, stripping keeps it, and Mach-O segment file offsets and VM offsets coincide — so unlike a PE COFF symbol table, it also reads correctly out of a mapped image. That is what makes it worth trusting on the memory-dump path.

lief accumulates the load command's deltas into offsets from the `__TEXT` segment's virtual address rather than into virtual addresses, so each entry is rebased on that segment. That is also where a fat slice's own base comes from. This was confirmed against lief's parser rather than assumed.

## Why both ship off

| source | measurement | why that is not enough to enable it |
| --- | --- | --- |
| Mach-O `LC_FUNCTION_STARTS` | across the 14 bundled Mach-O fixtures, linker-declared starts recovered rise 1884 → 2083 of 2155 (87.4% → 96.7%), no sample loses a function, and the function count rises by less than recall does — so wrong starts are being replaced, not merely added | the metric consumes the same table the pass reads. It can only show agreement with the linker, not that the boundaries are right. `evaluate_runtime.py` also gates on *any* function-set change across the benchmark corpus |
| ELF `.eh_frame` | moves an AArch64 fixture from 278 recovered functions to 279 | an unwind range is not a function by definition; `.eh_frame` also covers ranges that are not independent functions |

Both want ground-truth function boundaries before they become defaults. Turning either on is a one-line config change for anyone who wants to evaluate them.

## Tests

`tests/elf_ehframe_hdr_x64_xored` is new — the only bundled ELF with a populated `.eh_frame_hdr`. Every existing ELF fixture is built without unwind tables or keeps only a 4-byte terminator section, which made the first version of these tests compare an empty set against an empty set and pass for the wrong reason. Each test that reads a table now carries a positive control asserting the table is non-empty before asserting anything about it.

Built with `gcc -O1 -static -nostdlib -fasynchronous-unwind-tables -Wl,--eh-frame-hdr` from five small functions, then stripped (gcc 13.3.0). `-nostdlib` alone does not link the search table in.

- `tests/testEhFrameCandidates.py` — 45 tests
- `tests/testMachoFunctionStartCandidates.py` — 15 tests

## Validation

- full suite: 1220 passed, 2 skipped, exit 0
- `ruff check .`, `ruff format --check .`, type check: all clean
- `diff-cover`: 100% on 163 changed lines
- all 22 bundled fixtures byte-identical with both flags at their defaults


r0ny123:
A note on the second commit, `2ecfe08`: `.eh_frame` and `.eh_frame_hdr` are unioned rather than tried in order, so a binary carrying both contributes every FDE start from each. Taking the first source that parsed dropped the other's entries silently. The whole tree was swept for the same take-the-first-source-that-worked shape; this was the only site.

Validation on the final tree:

| gate | result |
| --- | --- |
| full suite | 1224 passed, 2 skipped, exit 0 |
| `diff-cover --fail-under=100` | 100% on 164 lines |
| type check | exit 0, no new diagnostics |

PR #275, opened 2026-08-21
fix(intel): recover switch dispatches that carry a notrack prefix

r0ny123:
> **Heads up: this sits on top of #265–#274.** Those ten are already open against `master`, and they're most of the file list GitHub is showing you here. The actual new work is everything from `c435c27` onwards. Merge the ten first — this is written to land on them.

Short version: modern gcc hides every dense switch in a non-PIE x86-64 binary behind a prefix SMDA couldn't see. And separately, those ten branches couldn't be judged for accuracy at all, because there was no labelled function-boundary data in the repo to judge them against. This PR does both — builds the ground truth, then uses it. A review round on top of that turned up eight more defects, which are fixed here too.

## The bug this started from

CET's `notrack` prefix marks an indirect branch whose target doesn't need to be an IBT landing pad. It's encoded as `0x3e` — which is also the DS segment-override byte — so capstone hangs it off the *operand* instead of the mnemonic:

```
0x4012fe   notrack jmp   qword ptr ds:[rax*8 + 0x402018]
```

`_analyzeJmpInstruction` spotted far jumps (`seg:offset`) by looking for a colon in the operand. A segment-qualified memory operand has one of those too, so the whole dispatch got filed as a far jump and thrown away: table never read, case bodies never linked to their function, and then the gap scan cheerfully promoted every one of those bodies to a function of its own. `-fcf-protection` is the default on Debian, Ubuntu and Fedora, so this is just what ordinary binaries look like now. Measured: 48 sites across 12 of 50 locally built binaries, and on one of them **a single switch produced 43 bogus functions**.

Seeing the dispatch turns out to be only half the job. The bound gets compared against the memory cell the index is loaded from (`cmp byte ptr [rdi + rcx], 0x19`), and the bound regex is deliberately register-only, so it can't match that — meaning the table ran on to its `0xff` fallback and ate whatever functions were stored behind it. The bound scan now follows the dispatch's index backwards through its copies instead.

And the classifier isn't the whole story either. The engine stores the operand exactly as capstone rendered it, so normalizing at the point where a branch is *classified* doesn't reach the code that later matches anchored patterns against a backtracked instruction. Put the prefix on the table **load** of a 32-bit switch rather than on the branch and the original bug comes straight back — all six case bodies carved into their own functions again.

## How any of this got judged

50 binaries built locally from dispatch-heavy C, across gcc 13.3.0 and clang 18.1.3, `-O0` through `-Os`, PIE / no-PIE / static / freestanding, x86-64 and x86-32. Ground truth is each binary's own `.symtab`, and SMDA analyses a **stripped copy**, so it never gets to peek at the table it's being scored against. **11,384 labelled functions.**

`body_splits` counts recovered addresses that land strictly *inside* a labelled function body. Those are unambiguous errors — a real function got broken apart. This matters because a plain function count can't tell a recovered function from a shattered one, which is exactly why the jump-table bound question couldn't be settled before.

| tree | recall | body splits |
| --- | --- | --- |
| `master` | 93.781% | 1095 |
| the ten stacked branches | 93.781% | 1062 |
| + seeing the dispatch, on its own | 93.710% | 722 |
| + the rest of the branch | 93.781% | 722 |
| this branch, with the landing-pad fix | **93.869%** | **712** |

So against the ten: **−350 body splits (−33%), recall up from 93.781% to 93.869%, nothing regressed across all 50 binaries.** Note that shipping just the classifier half would have been a net regression — that's why the two halves are one change.

## The eight things the review round found

Each of these produces a *wrong* answer rather than a missing one. All fixed, all with tests that fail on the pre-fix tree, each paired with a positive control that passes on both sides.

**A perfectly good x86 dump got refused.** The unsupported-instruction-set probe counted its signature across the whole buffer, and a dump is mostly data. A real stripped x86-64 shared object happened to carry 21 aligned copies of the openrisc return word — those four bytes spell `DH` in a UTF-16LE index table — and came back as an error. Sticking 64 bytes of UTF-16 text onto a bundled x86 memory dump did the same thing to it. Now the count has to land inside one sliding 64 KB span, each hit has to sit in code rather than text, and the two signatures that couldn't be separated at *any* threshold got lengthened instead. Across **60,940 files of every type on one filesystem: 117 used to be called some foreign architecture, now two are — and both are genuinely PowerPC object files.** All ten bundled foreign samples still get recognized.

**One instruction could run for 17 minutes.** The jump-table entry scan treated an attacker-controlled bound as an iteration count, so `0xffffffff` asked for a four-billion-entry walk over a 32 KB image. **1049.7 s → 0.002 s**, same targets recovered. The register-form path was reachable the same way at 3.3 s and is now 0.002 s too.

**A truncated function came back as two.** Every pass re-asked the timeout callback instead of reading the verdict already recorded, so a budget scoped to one pass would answer "not spent" to the next one, and the gap scan would promote the leftovers of the function it had just cut short. Dalvik had the same problem from the other end: polled 67 times, told "not spent" 66 of them, and reported `status: ok` for a run whose method loop it had already given up on — 64 functions presented as a complete analysis.

**An indirect call resolved to a stale address.** The backward walk modelled `mov` and `lea` and silently skipped everything else, so `mov eax, 0x402000` / `add eax, 4` / `call eax` resolved to `0x402000`.

**...and the guard that fixed that read a branch as a clobber.** Stopping at any unmodelled mnemonic is the right call for an instruction that writes a register it never names, and the wrong call for a jump. The walk recurses into predecessor blocks when the start block doesn't define the register — and a predecessor block *ends* with a branch, which is the first instruction its reversed walk looks at. So every recursion died on step one: **612 of 978 walks** across the bundled fixtures, on `jne` (197), `je` (108), `jmp` (96), `jb` (11) and `jae` (7). This isn't just a missing API name — a resolved target gets booked as a function candidate, so on the 153-file corpus the guard cost six real functions, three of which came back once branches were allowed through. The guard also now lets a value past a `call` whose ABI says it preserves the tracked register, and past a vector move that writes no general register at all. That part is sound and measurably free, but honestly the corpus says it recovers nothing extra — it's kept because a guard that thinks `call` clobbers `ebx` is wrong on its face, not because it bought anything. `loop` stays excluded (it decrements rcx), and so do `movd`/`movq`/`movsd`, since capstone spells each of those for a form that *does* write a general register.

**And one more clobber the walk read straight past.** The same backward walk exempts `mov` and `lea` from its clobber check, on the reasoning that the patterns below it model them. They model five forms. Any `mov` whose source is a register-relative memory operand — `mov eax, dword ptr [esi]` above all — writes the tracked register in a form none of them match, so the walk sailed past it and resolved the call against whatever had written the register earlier. That's a wrong target rather than a missing one, and it books the wrong address as a function candidate. `mov eax, dword ptr [esi]` is the ordinary vtable dispatch load, so this isn't a corner: counted on four 32-bit PE corpus samples, the walk passes an unmodelled write of the very register it's tracking 264, 302, 303 and 2 times per sample, against 499–574 writes it does model. Reproduced end to end — for `mov eax, 0x401000 ; mov eax, dword ptr [esi] ; call eax` the walk seeded `0x401000`, the dead assignment, as a function candidate, and now seeds nothing. Both ways this could have been overtightened are excluded and tested: a `mov`/`lea` whose destination is memory writes no register and doesn't stop the walk, and an unmodelled write of some register other than the tracked one just drops that register's recorded value and carries on.

**A far branch got read as a near one.** capstone renders `FF /5` and `FF /3` with a bare `ptr `, where the near forms name their width — and once a segment override is normalized away, that width is the only thing telling them apart. Eight bogus targets were being absorbed into the dispatching function.

**A function fragment got carved into a function.** A PE ARM64 exception record whose `UnwindData` carries flag 2 is a packed fragment continuing someone else's body, so its begin address is a mid-body word. The path that reads a *parsed* exception directory has always skipped those; the headerless carve only read the begin address, so a fragment starting on `stp x29, x30, [sp, #-16]!` sailed through as an independent function and split the one it was continuing. The check goes at admission rather than in the record validator, because the carve finds its table by counting consecutive valid records — rejecting flag 2 in the validator would just end the table at its first fragment.

**A damaged section header ate the entire analysis budget.** The AArch64 stored-pointer scan strode over whatever `sh_size` claimed — about a billion iterations of reads that can only ever return `None` — and handed back an empty function set for an image that normally analyses in 65 ms. Found by a test that planted a table at the tail of a real fixture, clobbered its section header table, and then ran for over ten minutes.

## Everything else in here

- **The analysis budget is an actual bound now.** `TIMEOUT` was polled between candidates and between passes but never *within* one, and the tailcall pass got invoked with no callback at all. Function analysis now polls every 256 basic blocks (across the bundled fixtures that's 5648 functions, median 1 block, p99 93, and only six ever reach 256); the exception-record walks and the tailcall pass poll too.
- **PE ARM64 exception records get carved out of a headerless image**, the way x64 has since 4.4.7. On a mapped image whose entries are two-word veneers, recovery goes from **0 functions to 24**. False-positive rate measured over 2961 files: the scan found a qualifying run in 51 of them, all x86, and admitted a candidate from exactly none.
- **Both raw-buffer probes are bounded**, including against a one-byte shift that used to defeat the bound entirely: a misaligned 100 MB buffer goes **6.19 s → 1.28 s**, and the AArch64 probe **2.91 s → 0.15 s**.
- **The three opt-in candidate sources carry real measurements** now instead of "off until validated". `RESOLVE_TAILCALLS` gets refuted rather than deferred: 70 more real starts, for 235 more broken functions. No defaults change — each flip moves a frozen baseline, so that's a separate decision.

## Baselines this moves on purpose

Three of the 23 bundled fixtures shift, all of it from the index tie, none of it from the segment fix:

| fixture | functions | instructions |
| --- | --- | --- |
| `mirai_x64` | 151 → 150 | 12,564 → **13,187** |
| `mirai_i386` | 151 → 150 | 14,758 → **15,318** |
| `rust_pe_gnu` | 2354 → 2354 | 154,207 → **154,227** |

Both mirai binaries dispatch on `cmp dword ptr [rbp + 0x10], 0xa` / `ja` / `mov eax, dword ptr [rbp + 0x10]`. The old scan recovered a bound of **1** and read a single table entry; the tied scan recovers **11**. The case bodies become blocks of the function that jumps to them, and nothing in either binary calls the address that stops being a function.

## The corpus gate, actually read instead of waved through

`evaluate_runtime.py` fails on any function-set difference at all, which means it goes red for precisely the changes that improve function recovery and green for the ones that change nothing. This branch makes it red. Here's the read, taken from the run's own report artifacts and reproduced locally against the corpus:

- **47 of the removals are false positives.** 43 sit in one darkgate padding run, every one with no inbound reference, 19 of them branching into the *interior* of a function that survives, and one decoding as `push esi / jno / dec eax / xor bh, [ecx]` — padding read from the wrong offset. The other four are one per danabot sample, decoding zero padding.
- **3 of the removals are addresses `master` only found because of a bug.** The call at `0xa49259` takes its target from `mov edx, dword ptr [0xb3a024]` — except `pop edx` at `0xa49246` overwrites `edx` first. `master` read a stale register and happened to land on a real Delphi method pointer, so it got a real function out of provably wrong dataflow. This branch refuses, correctly. Getting those three back properly means treating stored method-table pointers as candidates, which is a feature, not a repair.
- **1 removal was inherited from the stack underneath, and is fixed here.** The sixth file, `elf.akira`, lost `0x552ee0` — a function with five call sites. Bisected to `fa1bfcd`, the merge of `fix/intel-direct-jumptable-bounds` (#270), which made the direct-table read speculative when the bound isn't recovered. Root cause: a mis-derived table base of `0x20` — inside the ELF header — passes `isAddrWithinMemoryImage`, its first entry reads back `0x0`, and since the image is mapped at base 0, that passes too. Address 0 gets queued as a jump target, decodes as `add byte ptr [rax], al`, and the enclosing function is discarded whole. Fixed in this PR, along with the four siblings a tree-wide sweep of the class turned up.

**So yes, the gate is red — and behind it are 47 addresses this branch is right to drop and three it's right to stop inventing.** The gate can't tell those apart, which is the problem the classifier commit goes after rather than argues around.

## The address-0 admission class

`isAddrWithinMemoryImage()` answers "is this address inside the mapped range", not "is this a plausible target". In an image mapped at base 0 it happily accepts `0` — and two independent producers in this tree hand it a `0` that means *nothing found*: `struct.unpack` over a run of zero bytes, and `getReferencedAddr`, which returns 0 for an operand with no hex literal in it. That second producer is what makes this a class rather than a one-liner; a sweep anchored on the first shape can't see it. Every `isAddrWithinMemoryImage` call site got checked, five were admitting the sentinel, all five are fixed:

| site | how 0 gets there | what it costs |
| --- | --- | --- |
| direct table, per entry | zero bytes unpacked | address 0 queued for decoding |
| absolute table, per entry | zero bytes unpacked | address 0 queued for decoding |
| absolute table, **base** | `jmp qword ptr [rax*8]` renders no literal | a "table" read at address 0 |
| push-ret destination | `push <reg>` renders no literal | address 0 queued for decoding |
| indirect call slot | an unrelocated slot reads back 0 | the call booked against address 0 |

Both newly-guarded branch sites were reproduced end to end in a base-0 buffer: `push eax; ret` recovered **zero functions from the entire image**, and a dispatch through a zero table base absorbed four unrelated bodies into fabricated control flow.

**Corpus effect: one file out of 155.** On `elf.akira`, 13295 → 13299 functions:

- `0x552ee0` **recovered** — 5 direct callers, 58 instructions, only 8 of which were decoded at all before.
- `0x6960d0` **recovered** — 7 direct callers, 276 instructions, previously entered at two false starts *inside its own prologue*.
- `0x69c584` **lost** — and this one looked like a real function, a gcc entry with a stack-protector canary.

That loss was **not** the guard, and chasing it down turned out to be worth the trouble, because the first two explanations I had for it were both wrong. Tracing `analyzeFunction` on both trees:

```
BASE  0x69c584  cand  became_fn=True   blocks=155  owner=0x69c584   (at 3972 functions)
      0x69c95c  cand  became_fn=True   blocks=4    owner=0x69c95c

FIX   0x69c580  cand  became_fn=True   blocks=155  owner=0x69c580   (at 1098 functions)
      0x69c584  cand  became_fn=False  blocks=0    owner=0x69c580   (at 3976 functions)
      0x69c95c  cand  became_fn=True   blocks=129  owner=0x69c95c
      0x69c580  gap   became_fn=False  blocks=31   owner=None       (at 13292 functions)
```

I first read that as an ordering race, then as a candidate that claims a region and gets dropped without releasing it. Both readings had the addresses backwards. `0x69c580` is the **real entry** — it carries an `endbr64` — and `0x69c584` is simply its second instruction, the `push rbp; mov rbp, rsp` four bytes in. That's why the two are so easy to mix up: every landing-pad-prefixed function has a textbook prologue at entry+4, so entry and entry+4 are indistinguishable by prologue shape.

What actually happens is that two passes discard an already-recovered function — the gap-revert when a later candidate's analysis reaches bytes the function claimed, and the tailcall pass when an external jump lands inside a function. Neither one asks whether the entry it's throwing away was ever a guess in the first place. An `endbr64` on x86-64, or a `bti` on AArch64, is there because the compiler had to mark an address an indirect branch may legally target — that's the *image declaring the start*, not a scanner inferring it. Demoting it in favour of an address found inside its own body picks the weaker claim.

**That's fixed here, in `9d0c0b1`.** A candidate now answers `isLandingPadEntry()`: `False` on the arch-neutral base, `endbr64` on 64-bit intel, `is_bti_landing_pad` on AArch64 (which the backend already carried for its gap scan). AArch64 needed it because `AArch64Disassembler` subclasses the same recursive engine and hit the same revert with no protection at all. The predicate reads the candidate's **own** entry bytes and never the colliding offset, so a landing pad sitting in the interior of some other function can't grant it immunity. The tailcall revert itself is left alone — it rests on a real external jump edge rather than a scan's guess, so exempting it wholesale would disable legitimate splitting; only the re-creation is forced.

Against the labelled ground truth: recall **93.781% → 93.869%** (10676 → 10686 of 11384), body splits **722 → 712**, unlabelled 332 unchanged. Ten more real entries, ten fewer real functions broken apart, and no growth in addresses outside every labelled range — so it's finding entries rather than inventing them. With `RESOLVE_TAILCALLS` on, which is the configuration where the tailcall site is reachable at all: 94.396% → 94.431%, body splits 957 → 947. No bundled fixture moves. On the 155-file corpus one file differs, where all 11 added addresses are landing pads followed by a real prologue and all 14 dropped are interior code.

Worth being upfront about the risk here: `endbr64` is **not** exclusively a function entry. Under `-fcf-protection` clang emits one at every switch-case target, while gcc's notrack dispatch exempts its own. So that got measured rather than assumed — 35 of the 50 labelled binaries carry landing pads, 4690 of them in the clang builds, and body splits *fell* there. A case target wrongly held immune would have pushed that count up, since a case body lies inside a labelled function.

Net on `elf.akira` from the address-0 guards themselves: **+2 real, −1**, with the −1 now closed by the commit above.

**What the guards cost:** nothing measurable, and here's the corpus to back that. The three sites beyond the original entry guard change no function set across 155 samples. Instrumented, the sentinel reaches the image bound 49 times across 25 samples and is admitted 16 times — all of them in the one base-0 image, the rest being PE mapped above zero where the image bound already rejects it. The bundled fixtures can't see this at all: 211,734 instructions and 3,534 returns, not one of them preceded by a push. The call-slot site is starker still — 1,697 of 2,910 readable slots dereference to zero and 1,138 more land outside the image, against 63 of 63 outside the image and zero zeros on the fixtures. None of the 23 fixtures moves on status, function count, instruction count, API-call count or function offsets.

## The benchmark harness, so the next change is easier to judge

**The gate now tells you which way each changed address points.** It has no labelled boundaries, so all it can see is *that* the function set moved — but both sides' reports already exist, and structure is enough to separate the cases. Five readings, in order: an address still decoded on the other side inside another function is a boundary that moved; an address something references is a function whatever its body looks like; a body ending in a return, in an extent the other side doesn't decode differently, is a routine; a body that *opens* the way a compiler opens a function is a function even when the first three decline; everything else is padding. Scored against every address that changed in the run it was built from, hand-classified off the disassembly first: **54 of 54 agree.**

One of those five readings then turned out to be too loose, and it's worth recording because it produced a wrong headline number rather than a subtle one. The reading that says a dropped address "decodes like a whole routine" accepted **any** return, and the set it tested against carried the far and interrupt forms — `retf`, `retfq`, `iret`, `iretd`, `iretq`. Their opcodes are `0xca`, `0xcb` and `0xcf`, perfectly ordinary bytes in text and data, so a two-instruction body decoded out of a string lands on one routinely, while a compiled routine returns to its caller with a *near* return. Of the 15 addresses the run reported as "likely real functions lost", **12 were string bytes decoded as code** — two clusters that are the same embedded blob seen in two different danabot samples:

```
0xb3f8d0  sub byte ptr [ebx + eax*8 + 0x6f], cl ; add cl, al ; retf 0x666f
0xb3f8d9  in al, dx ; retf 0x346f
0xb3f8f1  int1 ; retf 0x986f
```

Every one of them terminates in a far return whose immediate ends in `0x6f` — the letter "o". The other 3 end in a near `ret`, open with the Delphi register convention's `push ebx ; push esi`, and are 16 instructions long. The reading takes only a near return now.

The fifth reading wasn't in the original four, and it got found the way these gaps do — by pointing the instrument at fresh data. It labelled `0x69c584` above `likely_false_positive`: nothing references it, its body ends in a `call` rather than a `ret`, and the other side decodes 1,758 addresses across its extent, so all three positive rules declined and it fell through to the fallback, which says *padding*. A function reached only indirectly can have all three of those properties and still be a function. The cheapest reading nobody had used was how the body *starts* — three shapes cover what a compiler emits: a CET landing pad, the frame-pointer pair, and the run of callee-saved pushes a register-convention entry opens with. Of the 47 hand-labelled false positives, 11 begin with a `push`, and every one of those either branches immediately afterwards or pushed an immediate rather than a register, so the shapes separate cleanly. Re-scored: **all 50 dropped labels intact**, and the canary entry now reads `likely_loss`. A gap, not a trade-off. Report schema version goes to 4 for the added signal. It's reported and never gated — four readings validated on one run is evidence for a reader, not grounds to fail a build — and an unreadable report costs you the reading and nothing else.

**Two corpus samples were never analysed by either side.** The harness was asking whether an ELF sample's *path* contained "x86" or "x64". A malpedia family folder only spells that out when it happens to be named that way, and `elf.akira` isn't — so two genuinely x86-64 samples have been invisible to this gate the whole time. It reads `e_machine` out of the header now, honouring `EI_DATA` instead of assuming a byte order. Over the corpus: 156 dispatched before, 158 after, the two additions exactly that pair, nothing dropped. One of them differs on this branch, and it's the only sample in the corpus that shows an address being **gained** — a direction the previously-visible files never exercised, and the one that surfaced the address-0 class above.

## Validation

- `ruff check`, `ruff format --check`, `make typecheck` — clean. `ty` exits 0 with 235 diagnostics, all pre-existing warnings, no errors.
- `make test` — green, exit 0. **1511 collected: 1510 passed, 1 skipped, 1575 subtests, 0 failures.**
- Diff coverage — **100%**, 0 uncovered (`diff-cover --compare-branch=upstream/master --fail-under=100`).
- Performance, interleaved on one machine with two A/A controls putting the noise at roughly ±2.5%: this branch reads +1.55% median paired, inside that band, while decoding ~5% more instructions on two fixtures.

## Two things checked rather than assumed

The ARM64 packed `FunctionLength` unit is load-bearing for the record validator, so it got round-tripped through lief's own parser: raw field 2/5/9 → 8/20/36 bytes, exactly 4×. It's 4-byte words.

The `notrack`-as-`ds:` rendering was confirmed by actually disassembling the encoding with the pinned capstone, rather than taken from documentation.

## Known limits, stated plainly

The xtensa and sh4 signatures each derive from a single artifact — their short forms don't separate from ordinary data at any threshold, so each one carries an extra word of the sequence its own sample emits. An image not using that idiom gets missed, which is the direction this code is deliberately biased. And `disassembleBuffer`'s architecture guess reads bytes the caller supplies, so it can be steered — pass `architecture=` when you know it, and it's never overruled.


danielplohmann:
Superseded by #282: the same branch rebased onto master after #265–#274 landed as squash merges (the duplicated stack commits made the original unmergeable). Content is identical — only the ten duplicate commits were dropped. Please base future stack children on #282's merge.

PR #276, opened 2026-08-21
perf(labels): parse a Go binary's pclntab once per image, not once per provider

r0ny123:
> **Heads up: this sits on top of #275** (at commit `cbb1dd3` — #275 has since picked up a further commit that this branch does not carry), which sits on top of #265–#274. GitHub can't point a fork PR at another fork's branch, so the file list and commit list below include all of #275 — **the only new work here is one commit, `239360a`** (5 lines of `src/`, plus tests). Merge #275 first.

Three hot-path candidates got profiled, implemented and measured. **One of them ships. The other two were rejected on measurement**, and the evidence for rejecting them is written up below and in the commit body — because the profile makes both of them look like obvious wins, and they aren't.

## What ships

More than one `GoSymbolProvider` gets constructed for a single analysis — `LanguageAnalyzer` builds one, `RecursiveDisassembler` registers another — and each calls `update()` with the same `binary_info`. So a Go binary's pclntab was being decoded **twice**. The two parses can't possibly disagree (same offset, same buffer), and the second result just gets thrown away in favour of an identical map.

The parsed symbols are now cached right next to the offset that `getPcLntabOffset` already caches on `binary_info`. The cache lives on the **image** rather than on the provider, because the image is what the two instances actually share — a per-instance memo can't see the other one's work, which is what the first attempt at this got wrong. Each reader gets its own copy of the dict, so one provider can't scribble on another's map.

| | before | after |
| --- | --- | --- |
| `_parse_pclntab` calls | 2 | **1** |
| wall clock, median of 3 interleaved pairs | 3.822 s | **3.620 s** |

All three interleaved pairings favour the cached build. Reports come out **byte-identical** — function set, instruction count, and all 1762 resolved Go symbol names — which is the property that makes a timing claim mean anything in the first place.

Non-Go binaries aren't affected: measured over 40 corpus samples across two interleaved passes, 46.80 s / 46.35 s before against 46.32 s / 46.41 s after. Though to be straight about it, that's not really evidence of neutrality so much as evidence the code never runs there — **none of those 40 samples is a Go binary**, which is exactly why the Go-specific measurement above had to be built rather than read off the corpus.

## What got rejected, and why the profile lied

**The per-byte `code_map` / `ins2fn` fill** is the single hottest function in a run. Every formulation that preserves dict semantics turns out *slower* than the loop that's already there. Over 200k synthetic instruction spans with a realistic x86 length distribution:

| formulation | time |
| --- | --- |
| the current per-byte loop | **186 ms** |
| `dict.update` over a chained `zip` | 243 ms |
| per-instruction `zip` | 295 ms |
| `dict.fromkeys` for the constant map | 263 ms |
| hoisting both `__setitem__` bound methods | 225 ms |

The dict stores are the cost, not the loop overhead. Getting a win here means abandoning the per-byte representation altogether — an interval list plus `bisect`, which has only four `ins2fn` read sites but a real behavioural surface — rather than writing a faster fill. That's a different change with a different risk profile, so it's not in here.

**The candidate-heap comparator.** Every heap comparison goes through `_MaxHeapItem.__lt__` into `FunctionCandidate.__lt__` — two Python calls plus two score reads, about 6.5% of a run by profile. Swapping the wrapper for `(-score, addr, item)` tuples so `heapq` compares in C removed **93,859 Python calls** from a two-sample profile.

And it's **4.6% slower** end to end: 46.24 s / 46.10 s baseline against 48.41 s / 48.18 s, across 15 corpus samples timed twice, interleaved, with the two passes agreeing to within 0.3%.

Reason: a candidate's score changes in place, so `update()` has to rebuild every key before re-heapifying — an O(n) Python pass — where the old code handed the unchanged list straight to `heapq.heapify` in C. The updates outnumber the comparisons the tuples were supposed to make cheaper. **A call-count drop is not a timing result**, and a profile attributes cost to where it's spent, not to what a change will cost somewhere else.

## Validation

- `make test` — green, exit 0. **1496 collected: 1495 passed, 1 skipped, 1568 subtests, 0 failures.** Three of those tests are the ones added here.
- `ruff check`, `ruff format --check`, `make typecheck` — clean. `ty` exits 0, no errors.
- Diff coverage vs the base branch — **100%**, 5 changed `src/` lines, 0 uncovered.
- The three regression tests: a second provider must not re-parse; the cached map has to be copied rather than shared; and a positive control that a *different* image parses again, so the cache can't be passing by simply never being consulted. The first fails on the pre-change tree; both controls pass either side.

## Why this is its own PR

#275 deliberately changes the recovered function set, and the corpus gate is the thing that proves it. A performance change's entire claim is that the output is identical. Folding the two together would wreck the benchmark's ability to tell "the correctness work moved this address" apart from "the performance work broke it" — so this is stacked rather than merged in.


danielplohmann:
Superseded by #283 — same change rebased onto master after #282 landed.

PR #278, opened 2026-08-22
fix(intel): record a data reference for an absolute memory operand

r0ny123:
> **Heads up: this stacks on #275.** GitHub can't point a fork PR at another fork's branch, so the file and commit lists below include all of #275 — **the new work here is two commits, `4f7517d` and `c9dea15`**. Merge #275 first.

This is the enabling half of a feature that got abandoned once already. The short version of that story: reading Delphi method tables to recover the virtual methods nothing else names is a real win, but the rule couldn't be made defensible, because the one filter that would separate a genuine dispatch table from a run of garbage — "does code actually reference this table?" — didn't exist. This PR makes it exist. The retry that uses it is stacked on top.

## The gap

The intel backend called `addDataRef` from `JumpTableAnalyzer` **and nowhere else**. An address a function loads outright — a dispatch table, a global, a stored method pointer — left no trace anywhere in the report. The AArch64 backend has recorded these from its own operands since it landed.

On one corpus dump the gap is total: **1,052 data-reference targets across the whole image, and not one of them in the `0xb3xxxx` region where the sample keeps its tables** — including the table that `mov edx, dword ptr [0xb3a024]` loads by name, three instructions before a call that dispatches through it.

## The change

Every operand naming an absolute address inside the mapped image now books a reference. A base or index register, or a rip-relative displacement, leaves the effective address unknown here, so those forms deliberately don't match — the regex is anchored and only accepts a bare literal inside the brackets.

**The AArch64 guard couldn't just be copied over**, and this is the load-bearing detail. It skips an address the container declares as code, which is right for a file and *inert* for a dump: `isInCodeAreas` returns `True` for the **whole image** when no code areas were parsed, so consulting it there refuses every single reference. So the veto only applies where the container actually declared something. That's the difference between this pass working on a memory dump and doing nothing whatsoever, and a memory dump is the input this library exists for.

`c9dea15` fixes a second, narrower mistake in the first commit that came out of review. The pass originally skipped every branch, on the reasoning that a branch operand names a code target the branch analyzers already record. True of a *direct* branch — which names its target as a bare immediate and so could never have matched the bracketed form this pass reads anyway. The filter's only actual effect was to exclude `call dword ptr [0xADDR]` and `jmp dword ptr [0xADDR]`, where the address is the **pointer slot** the branch reads through. That's data, whatever the branch analyzers go on to do with the value they find in it.

## Measured

**Corpus, 155 samples: five files change, +9 functions and −23.** Every one of the 19 removals on the two samples I looked at has **no inbound reference**, sits within 64 bytes of an address the program itself reads or writes as data, and decodes as bytes no user-mode code contains:

```
in al, dx      in al, 0xca      int1      out dx, al      aam 0xdb
```

Privileged port I/O and a debug trap, in fragments of two to twenty-three instructions, repeating byte for byte across two builds of the same family. Two of them are written to by name — `mov dword ptr [0xb3f8d0], eax` — which is what a global looks like and what code never does. These are data regions that were being decoded as code.

| corpus | result |
| --- | --- |
| malpedia, 155 samples | 5 files differ, +9 / **−23** (false positives on data); `c9dea15` adds +1 / −0 |
| bundled fixtures | one deliberate move, explained below; every other field on all 23 unchanged |
| labelled 50-binary ground truth | recall 93.869%, body splits 712, unlabelled 332 — **identical**, and identical again after `c9dea15` |

**That ground-truth row is not evidence the change is free, and the firing counter says why:** the pass is called **1,265,470 times** on the labelled corpus and books **17** references, because those samples are files whose declared code areas the veto respects. An identical before/after there is the shape of a measurement about a corpus that doesn't exercise the code — not the shape of a change that costs nothing. The dumps are what exercise it: on the two 32-bit dumps measured, `c9dea15` takes data references from 664 → 1585 and 786 → 1800, naming 258 more distinct addresses each.

## The one fixture that moves

`asprox_0x008D0000` was recovering `0x8de000` as an eight-instruction function opening `lodsb ; wait ; fidiv word ptr [edi + 0x51]`. Its single inbound reference is `call dword ptr [0x8de000]` at `0x8d31f4` — so that "function" is the pointer slot itself, decoded as code, which is exactly the promotion the review comment predicted. Booking the slot as data drops it: `testIntegration` moves from 106 functions and 15714 instructions to 105 and 15706, and both assertions now carry that reason inline.

## A weakness this exposes in #275's changed-address classifier

The classifier #275 introduces labels **17 of the 19 removals `likely_loss`**, because each one ends in a `ret` and doesn't overlap differently-decoded bytes. A two-instruction fragment sitting on a global the program writes to is not a lost function. The rule wants a length floor, or a check against the data references this commit makes available. Left alone here — it's a separate change and needs its own scoring against the labelled set.

## Validation

- `make test` — **1523 collected, 1570 subtests, 0 failures**, exit 0.
- Diff coverage vs the base — **100%**, 0 uncovered lines.
- **17 tests**, each arm with a positive control: register-relative, scaled-index and rip-relative operands must **not** match; the code-area veto must narrow the result rather than empty it; an image declaring no code areas must still record; a call through an absolute slot must record the slot, while a direct call — the control — must record nothing and still reach its callee.

The `binary_info` read is guarded because tests build partial disassembly doubles: the first full run failed nine existing tests that construct a `SimpleNamespace` with no `code_areas` on it at all.

## Not taken here

The same review asked for the operand width to be passed to `addDataRef`, so that a `dword ptr` operand marks four bytes of `data_map` rather than one. The data-model point is right and the patch is four lines — but measuring it first showed it moves **10 corpus samples, +20 / −5**, by breaking apart long garbage functions and letting the gap scan re-carve the region, recovering real Delphi entries in some places and manufacturing shorter fragments in others. That's a recovery heuristic, not a bookkeeping fix, so it wants its own change and its own scoring.

## Next

With these references present, "the table is referenced by code" becomes a real filter instead of an absent one. That retry is #279, stacked on this.


danielplohmann:
Superseded by #284 — cherry-picked onto master after #282/#283 landed.

PR #279, opened 2026-08-22
feat(intel): recover Delphi method-table entries nothing else names

r0ny123:
> **Heads up: this stacks on #278**, which stacks on #275. GitHub can't point a fork PR at another fork's branch, so the file and commit lists below include both of those — **the new work here is `d7a67ee` and `6c4a82d`** (the third commit is just the merge that picks up #278's head). Merge #275, then #278, then this.

## What this is

A late candidate source that reads Delphi method tables, so the entries a program only ever calls *through* the table stop being invisible.

I tried this once before and abandoned it, and it's worth saying why, because the reason is the whole point of #278. A Delphi virtual method is reached only through the object's method table. An entry no call site names is reachable by nothing the recursive passes do: no reference seeds it, and gap analysis only finds it when it happens to fall between two functions that were recovered. The obvious rule — "a run of aligned dwords where most entries are already known functions" — works exactly as measured at a ≥80% threshold: it captures all three target addresses with zero body splits. The problem is what *else* it captures. Of the 1,035 addresses it seeds across three samples, the overwhelming majority point at bytes that don't read as code at all:

```
  45  xor byte ptr [edi], bl ; lodsb al, byte ptr [esi]
  45  nop ; inc edi
  34  mov ah, 0x9c ; scasd eax, dword ptr es:[edi]
  31  <undecodable>
```

That's roughly 345 unvalidated candidates per sample. Three filters that should have separated them each failed for a concrete reason: the `COMMON_PROLOGUES` entry-shape table is MSVC/C-shaped and has no Delphi entries (`55` scores 51, `558bec` scores 50, but Delphi's register convention opens `push ebx; push esi` and `53` scores **6**, so zero of the 1,035 clear the score ≥ 30 bar); `data_map` is empty on raw memory dumps, so "is the target already known to be data" has nothing to consult; and the third filter — "does code reference this table?" — **didn't exist**, because the intel backend booked no data reference for an absolute memory operand.

#278 is that missing signal. This is the retry, and with the reference filter actually available the rule separates cleanly.

## Why these functions are lost without it

Three such entries show up on the malpedia corpus. Upstream `master` recovers all three, and a stack probe says how — each is booked by `IndirectCallAnalyzer.resolveRegisterCalls` after the backward walk has recursed into predecessor blocks:

```
== FunctionCandidateManager.addCandidate
    IndirectCallAnalyzer.py:306 resolveRegisterCalls
    IndirectCallAnalyzer.py:257 processBlock
    IndirectCallAnalyzer.py:258 <genexpr>
    IndirectCallAnalyzer.py:257 processBlock
```

That route reads a register the walk never saw written — a wrong answer rather than a missing one — and the clobber guard on #275 stopped trusting it. The entries are real; the way they were being reached wasn't. Restoring that route would just restore the guess.

## The rule

A run of aligned dwords is read as a method table only when two independent facts agree:

1. **it validates itself** — at least 80% of the run's entries are already recovered functions;
2. **code points at it** — some entry address in the run is a data-reference target.

Neither is enough on its own, per the histogram above.

The scan runs after gap analysis *and* its drain, through a new `locateLateCandidates` hook, because both facts are products of the passes before it. At the existing deferred hook, one of these samples has only 4,423 of its eventual 6,829 functions, and the pass admits nothing at all.

`6c4a82d` adds a second drain **after** the late pass, which matters for a reason that isn't obvious: a late function can itself resolve a register-indirect call and register its callee, and since the tailcall pass is off by default, nothing else would ever drain that candidate. It's pinned by a regression test rather than by a corpus number.

## Evidence

| corpus | result |
| --- | --- |
| malpedia, 155 samples | **3 files change: +3 functions, −0**, at exactly `0xa65150`, `0xb07370`, `0x6f7360` — the three addresses above |
| 23 bundled fixtures | unchanged — the suite's own exact function and instruction assertions for the bundled dumps are green |
| 50 locally built labelled binaries | 0 rows differ — recall 93.869%, body splits 712, unlabelled 332 |

That ground-truth row is reported with an instrument rather than an assumption: a firing counter shows the pass is entered on all 50 binaries and detects Delphi on none of them, so that corpus can't judge this change in either direction.

One more thing worth flagging about the evidence table — the "bundled fixtures unchanged" row originally cited a standalone harness of mine, which analyses every fixture with `disassembleUnmappedBuffer`. But the bundled dumps get analysed as *mapped* images at their real base address, so for exactly the fixtures that matter it was measuring a different analysis than the suite does, and it would not have caught a change to them. The row cites the suite's own assertions now, which are the real gate.

The scan's inner loop was rewritten after the first corpus measurement — streaming `struct.iter_unpack` over 64 KiB chunks instead of a per-offset `unpack_from` accumulating a list of tuples. Rather than argue the rewrite is equivalent, it was re-measured: the two implementations produce **identical function sets on all 155 samples**.

## Cost

The scan only runs on a 32-bit image where the Delphi VMT pass fired; everywhere else it returns on its first test. Where it does run, measured end to end against the analysis it's part of:

| sample | image | pass | share of run |
| --- | --- | --- | --- |
| danabot dump | 1.19 MB | 58 ms | 0.5% |
| danabot dump | 1.45 MB | 79 ms | 0.6% |
| danabot unpacked | 2.72 MB | 107 ms | 0.4% |
| darkgate unpacked | 0.42 MB | 17 ms | 0.3% |

Nothing about a run is carried along as a list. A crafted image can make a single run as long as the image itself, so every test either streams over the run with `struct.iter_unpack` or is decided from its length, and the budget is polled once per 64 KiB rather than once per entry — including inside a single long run, which `6c4a82d` routes through one generator so no traversal can run for seconds without consulting it.

## Validation

- `make test` — **1549 collected, 1548 passed, 1 skipped, 1570 subtests, 0 failures**, exit 0.
- Diff coverage — **100%**, 0 uncovered.
- `tests/testPointerTableCandidates.py` covers each arm, and every test expecting an admission is paired with a control that removes exactly one of the two facts and expects none. **25 of the 26 fail on the pre-change tree** (checked by running this file against #278's head); the one that passes is a control, which is what a control is for.

## What this deliberately does not do

- **The "points into code" bound is still `min(recovered) … max(recovered)`.** Narrowing it to a non-circular executable range was the other idea on the table, and it wouldn't help on the samples that matter: they're memory dumps with no section table, so an executable-range bound is empty there. The span is only a cheap pre-filter now — the two facts above do the separating, and the measured outcome is exact.
- **The pass is gated on Delphi**, not on the shape alone. The shape isn't Delphi-specific, but the measurement is: widening the gate is a separate change that needs its own corpus number.


danielplohmann:
Superseded by #286 — cherry-picked onto master after #284 (its data-ref dependency) landed.

PR #281, opened 2026-08-23
fix(intel): recover the function starts four boundary defects were hiding

r0ny123:
> **Heads up: this stacks on #275**, alongside #276 and #278 (all three branch off it independently). GitHub can't point a fork PR at another fork's branch, so the file and commit lists below include all of #275 — **the new work here is the six commits from `3423baa` onwards.** Merge #275 first.
>
> One note on the commit trailers: every commit carries a `Refs: #125` line pointing at an issue in my fork that tracks where SMDA's function recovery sits against Ghidra. There's no equivalent issue here, and `#125` in this repo is an unrelated closed dependabot PR — so please read those trailers as fork-local bookkeeping rather than a reference to anything upstream. The context they point at is summarised below instead.

Four independent defects that made SMDA report the wrong function boundaries: one merged whole functions together, one dropped a function's start entirely, one left every case body of a switch to be promoted into a function of its own, and one named a function one byte into its own prologue. Two further commits answer review findings raised on the branch.

For context on why these four and not others — I measured SMDA's recovery against Ghidra 12.1.3 run headless over identical inputs, and the resulting gap splits into three unrelated problems rather than one. Roughly two thirds of it is `.cold` fragment parity, which I'm deliberately **not** chasing: matching it would break MCRIT hashes to chase a metric. What's left is genuinely undiscovered functions, and body splits (712 against Ghidra's 47). This PR is those two.

## Measured

50 labelled binaries, each scored against **its own symbol table**, with the analysis run on the **stripped** copy — so neither side reads what it's being scored against. `body_splits` counts recovered addresses landing strictly inside a labelled function body, which is an unambiguous error.

| | recall | body splits | unlabelled |
|---|---|---|---|
| base (#275's branch) | 93.869% | 712 | 332 |
| + register-base jump tables | 93.869% | **316** | 332 |
| + alignment-floor exemption | 93.974% | 314 | 332 |
| + padded landing-pad cut | 94.501% | 314 | 332 |
| + prologue entry seed | **95.098%** | **306** | 332 |

Recall 93.869% → 95.098% (+140 labelled starts), body splits 712 → 306 (**−57%**), unlabelled unchanged, and **no file regressed**. All 23 bundled fixtures come out identical in function and block counts. The two review-fix commits move none of these numbers.

## The commits

**1. `fix(intel)`: switch dispatches whose jump table base is held in a register.** Both existing arms find a table by pulling an address out of the operand text. A rip-relative `lea` into a register leaves no address to pull, so the dispatch resolved to nothing and every case body became a function of its own. The base is now chased back to its `lea` through a walk that stops at the first instruction not known to leave the register alone.

Gated on non-PIE ELF, and that gate matters: a shared object relocates every absolute address it stores, so an absolute table there isn't a switch — it's a function-pointer table, and reading one as a switch merged four functions into a 14 KB routine in libc and two multi-megabyte ones in libcrypto. An ELF whose header type can't be read answers "position independent" too, because refusing to read a table costs a switch its case bodies, while wrongly admitting one merges whole functions together.

**2. `fix(core)`: a direct call outranks the inferred alignment floor.** The floor is inferred from where call-referenced candidates sit, and it tolerates a twentieth of them being unaligned — and then the consumer turned around and refused every one of them. `abort` in a static glibc sits at an odd address, is reached by 37 direct calls, and was never analysed at all: 419 bytes belonging to no function. Applying the floor to that whole population discards exactly what the inference already allowed for. Symbols stay subject to the floor — exempting those was measured to cost true positives.

**3. `fix(intel)`: cut a function at the padded landing pad it runs into.** The existing cut only fires where padding follows a call, so a function padded up to the next entry decoded straight through it. glibc's multiarch entries are exactly that shape — `__strcasecmp_avx2` swallowed all 2546 bytes of `__strcasecmp_l_avx2`. A CET landing pad is only emitted where an indirect branch can arrive, which is the entry-shape evidence that alignment alone doesn't give you.

**4. `fix(intel)`: seed the entry of a prologue that opens with `push rbp`.** `push r15; push r14` is a run of callee-saved pushes, not a distinguishing opener, so it matches one byte into a clang prologue that keeps no frame pointer. `main` in every clang build above `-O0` was being recovered at its second instruction — so its real start counted as missing *and* the recovered address counted as a body split. One error, reported twice.

**5. `fix(intel)`: apply an implicit register write only to the form that makes it.** Review finding. The one-operand `imul` writes rdx:rax; its two- and three-operand forms write nothing but the register they name. Answering for the pair regardless of form stopped a walk chasing a base held in rax or rdx at an `imul rcx, rsi` that can't touch either. The backend's syscall walk already drew this distinction, so the two disagreed with each other.

**6. `fix(intel)`: keep the push-pair seed out of a function the image already names.** Review finding, and commit 4's residue. The seed steps back one byte onto a `0x55` read as `push rbp` — but `push 0x55` is spelled `6a 55`, and any encoding ending in that byte looks identical. On one of those the step lands inside a real entry, producing a second function one byte into it. Fixed by asking the image rather than the bytes: symbols, call references and exception records are all located before the prologue scan runs, so a declared entry opening in the bytes just before puts that byte inside its first instruction, and neither address gets seeded.

## Anti-regression sweep

Sweeping the root causes tree-wide turned up two real siblings, both fixed in commit 1 with regression tests:

- **The relative-table reader** admitted a parsed `int32` delta as a jump target on the mapped-image bound alone, exactly like the absolute readers did. It fires on **221 of 5045** entries across five corpus binaries.
- **The bound recovery carried an index tie across a direct `call`.** A call returns its value in rax, but its operand is a bare address, so the destination parse read no key and the tie survived — reporting a compare from *before* the call as the bound. A recovered bound is trusted as exact, so an overlarge one runs the entry scan past the end of its own table. Reproduced at 1000-vs-4 against controls; a live tie crosses a call on **82 of 336** bound recoveries.

Both mnemonic tables were checked against capstone by decoding candidate encodings and reading what it actually reports written, rather than by eye. That turned up `sal` (the shift group's `/6` opcode, a mnemonic in its own right) and `rdtscp` missing; two deliberate imprecisions are documented now rather than left to be rediscovered.

**These sweep fixes move none of the numbers above, on any file, and change no fixture** — that's the whole of what can be said for them empirically. They're in because each is a wrong answer reachable by construction, and the firing counts show the paths are ordinary rather than exotic.

## Known limits, recorded rather than papered over

- Commit 6's window is eight bytes, picked by measurement: over 404 matches across four images, 347 step onto a byte the disassembler independently recovers as an instruction start and none step into an instruction's interior; going to a full 15 costs one of those 347. A declared entry opening 9–15 bytes back would still be missed. Three narrower fixes were built and rejected first — requiring the preceding byte to end a function cost **four real functions** on the bundled Rust PE (entries there sit behind `eb`/`e9` tail jumps and `0f 0b`, so the bytes that may precede an entry aren't enumerable); seeding both addresses gave 3545 functions against 3413 on that fixture; syncing a decode forward from up to 48 bytes back lost 46–64 of the 347 real entries.
- The landing-pad cut doesn't fire behind legacy filler idioms (`mov edi, edi` and friends), which capstone doesn't spell `nop`. Lossy only, never wrong — and unmeasured, so not widened here.
- An MSVC `push rbp; push rbx` prologue could in principle seed two bytes late. 0 instances across 23 fixtures against a 406/377 positive control, and there was no MSVC x64 artifact available to validate a fix against.
- The indirect-call walk still treats `push` as preserving `rsp`. It can't produce a wrong answer today, and narrowing it would tighten a walk that has previously cost real functions on a PE corpus not reachable from here.

## Validation

- `make test` — green, exit 0: **1578 collected, 1577 passed, 1 skipped, 1593 subtests, 0 failures.**
- `diff-cover --compare-branch=upstream/master --fail-under=100` — **785 lines, 0 missing, 100%**, across the whole stack rather than just this branch's share.
- 23 bundled fixtures — **0 differ** (positive control: 12 fixtures, 7261 functions).
- Each of the first four commits passes its targeted suite independently, so the series stays bisectable.


danielplohmann:
Superseded by #285 — cherry-picked onto master after #282–#284 landed.

Issue #317, opened 2026-09-07
Collected review follow-ups from the open-PR triage pass


danielplohmann:
Running list of small follow-ups noticed while triaging the open PR queue. Each is something a merged PR left slightly short of right — not a defect worth blocking that PR on, and not worth a PR of its own either.

**This one is ours to do, not up for grabs.** Every item here comes out of reviewing someone else's contribution, so it is on us to finish rather than to hand back. Please do not open a PR against these; they will be closed out from here in a small number of scope-separated changes once the review pass is done.

Grouped by the `type(scope)` each fix belongs under, because they cannot travel in one PR — the title validator takes a single type and scope, and a combined "misc cleanups" change would be exactly the wider-than-its-title problem we are trying to avoid elsewhere.

---

### `docs` — README overstates which fields identify a backend

From #308 (316c2bc). The new "What an offset in a report means" section tells a consumer the backend is named "in two places, `report.architecture` and `metadata.language` in `toDict()`".

`metadata.language` is a **source-language score map**, not a backend identifier:

```
cil      -> {'.net': 1.0}
dalvik   -> {'dalvik': 1.0}
intel    -> {'c/asm': 0.1, 'delphi': 0.0, '.net': 0.0, 'visualbasic': 0.0, 'go': 0.0, 'rust': 0.0, 'c++': 0.0}
aarch64  -> {'c/asm': 0.1, 'delphi': 0.0, '.net': 0.0, ...}
```

It is a single decisive entry on the two managed backends and a distribution on the native ones — and that distribution carries a `.net` score. So a consumer following this advice and branching on `.net` appearing in `metadata.language` can read a *native* report as managed, then treat its virtual addresses as file offsets. That is the precise error the section exists to prevent.

`report.architecture` is authoritative and correct on all four backends. The `CilDisassembler` docstring added by the same PR is already careful here — it says "the language score map" — so only the README half needs tightening: name `architecture` as the field to branch on, and describe `metadata.language` as corroborating rather than equivalent.

### `fix(ci)` / `refactor(common)` — three annotations were removed rather than satisfied

From #302 (7f892b9). Addressing ty 0.0.74's `unsound-return-statement` by deleting the return annotation silences the check but loses the documentation, and in two of the three cases the root cause is an untyped attribute one level up:

| Site | What happened | Root fix |
|---|---|---|
[...]

PR #319, opened 2026-09-07
chore: cut v4.5.1

danielplohmann:
Cuts v4.5.1. **No longer held for #299 and #300** — those are re-scoped to v4.6.0, so this is ready.

Everything merged so far is in and the branch carries all of `master` (`6240b74`), so it adds exactly three lines: the entry, `VERSION`, and `__version__`.

## What the entry already covers

All fourteen changes since v4.5.0: #302, #303, #304, #305, #306, #307, #308, #310, #316, #318, and today's #309, #311, #312, #320, #321, plus the Dependabot `ty` bump #302 had to answer for.

Grouped as **Recovery (intel)** / **Recovery (AArch64)** / **Defaults changed** / **New (default off)** / **Docs** / **Housekeeping**, closing with the movement note the v4.4.7 entry established.

## The release-level figures, measured on this tree

`rust_pe_gnu_xored`, against the 2,186 `.text` function symbols its mingw build retains:

| | v4.5.0 | this branch | delta |
|---|---|---|---|
| starts | 2355 | 2201 | |
| TP | 2021 | 2047 | **+26** |
| FP | 334 | **154** | **−180** |
| recall | 92.452% | 93.641% | +1.19pp |
| precision | 85.817% | **93.003%** | **+7.19pp** |

The 11-sample ARM64 Mach-O corpus, against `LC_FUNCTION_STARTS` (2,056 truth functions):

| | v4.5.0 | this branch | delta |
|---|---|---|---|
| TP | 1787 | 1818 | **+31** |
| FP | 1113 | 1107 | **−6** |
| recall | 86.916% | 88.424% | +1.51pp |

No sample loses recall in either table.

## Why #299 and #300 moved to v4.6.0

They were going to ride along, on this morning's plan. Two things changed it:

- **#300's measurements are against a baseline that no longer exists.** It reports the 72-cell built AArch64 ELF corpus at PPV 76.676 → 80.554; #310 measured the *same* corpus at 87.596 → 89.186 and is merged as `7d01480`. Its post-change figure sits below the pre-change figure of work already in master, so the table has to be re-run before any of it can be read — and only @r0ny123 has those corpora.
- **#300 is a `feat(core)` carrying ten engine changes.** That is a minor-bump shape, not a patch addition to a release that is otherwise fixes plus one default flip.

#299 contains #300, so it follows. Neither is blocked on a decision from here — they need a rebase and a re-measurement, both tracked on those PRs.

## Changelog formatting

@r0ny123 asked on this PR whether the entry could be formatted as a list. Done in `d678566`: the release summary is the top-level bullet, one sub-bullet per section, and a third level inside the three sections that cover more than one change.

Only this entry changes shape — the other 192 keep the one-line format. Converting them is a large rewrite that loses nuance in translation, and #323 proposes keeping them verbatim under an `Older releases` heading instead. The wider keep-a-changelog adoption is tracked there, deliberately scheduled *after* this release so that a format change and release content are not reviewed in the same pass.

The text is unchanged, and that was verified mechanically rather than by eye: stripping list markers and collapsing whitespace reproduces the original entry exactly. Two opening letters were capitalised, where splitting a section header onto its own line left a bullet starting mid-sentence.

## Still manual after merging

There is no release automation in this repo — no workflow publishes on a tag. Merging this lands the entry and the version bump; tagging `v4.5.1` and publishing remain manual.

## Carried caveats

Two figures are marked as the contributor's own because nothing here can reproduce them, and #322 tracks closing that gap:

- **#310's `bti j` 803/0 split** — no bundled fixture holds a single `bti j` word.
- **#312's ARM64 PE interior-gap result** — no bundled fixture is an ARM64 PE. Scanning all 104 fixture files: 8 × i386, 4 × AMD64, 1 × `0xFD1D`, **0 × `0xAA64`**.

Both mechanisms were verified directly, so this is about magnitudes rather than correctness.

Also recorded: @r0ny123 [withdrew #304's built-Rust precision figure](https://github.com/danielplohmann/smda/issues/317#issuecomment-5580795241) after reproducing this side's measurement cell for cell. It never reached the changelog — only the pooled 191-sample numbers did, which stand.

And a note for whoever reads a benchmark verdict into this entry: the corpus gate's **correctness** half is solid (it cleared #320 and #321 at 0 of 155 files differing over 175,946 functions), but its **timing** half called two behaviour-neutral typing changes −3.70% and +14.78%. Cross-runner variance, tracked in #322. Do not quote a speedup from it.

## Validation

`pre-commit run` clean on all three files. Full suite on this tree: 1958 passed, 2 skipped, 2593 subtests. `smda.__version__ == SmdaConfig.VERSION == "4.5.1"`, which `testIdaExporterReuse` asserts. `ESCAPER_DOWNWARD_COMPATIBILITY` stays at `4.4.5` and `INTEL_PIC_HASH_ESCAPE_VERSION` at `4.3.5` — no escaper output changed, so nothing needs reprocessing.


danielplohmann:
Back to draft. #309, #311 and #312 landed this morning and their sections are in, along with #320 and #321 — but #299 and #300 are being taken in a later session and the release waits for them, so this stops short of the merge for the same reason it did yesterday.

The branch carries all of `master` (`6240b74`) and adds three lines: the entry, `VERSION`, `__version__`.

One thing worth flagging for whoever picks this up rather than leaving it to be discovered: **the closing movement section goes stale when #300 merges.** It changes function-boundary decisions, so both scored tables in the body — `rust_pe_gnu_xored` against its COFF symbols, and the ARM64 Mach-O corpus against `LC_FUNCTION_STARTS` — need re-measuring against the final tree, not amending. #299 claims no output change, which the corpus benchmark's correctness half can confirm directly.

r0ny123:
@danielplohmann Could this changelog be formatted as a list to make it easier to read? Also, could we use https://github.com/olivierlacan/keep-a-changelog across the mcrit ecosystem?

danielplohmann:
@r0ny123 both of your questions, answered separately:

**Formatting as a list** — done in `d678566`. The release summary is the top-level bullet, one sub-bullet per section, and a third level inside the three sections that cover more than one change (Recovery (intel), Recovery (AArch64), Housekeeping). At ~2,700 words on one line this entry was the one that provoked the question, so it seemed right to fix it here rather than make you wait for the wider change.

Only this entry changed shape; the other 192 keep the one-line format for now. The text is untouched — verified by stripping the list markers and collapsing whitespace, which reproduces the original string exactly, rather than by reading it back. Two opening letters were capitalised where splitting a section header onto its own line left a bullet starting mid-sentence.

**keep-a-changelog across the ecosystem** — yes, and the decision is already made on our side for SMDA and MCRIT both, so the two stay consistent. #323 has the details and asks for your input on the parts that are genuinely open. Four of them would benefit most from your view:

- version-order vs date-order, given three releases in this file are patches on an older line published after a newer minor
- whether to keep the per-backend split (`Recovery (intel)` / `Recovery (AArch64)`) as a bullet prefix under the spec's change-type headings, since mapping literally collapses six sections into mostly `Fixed`
- a custom `Compatibility` section for the "recovery output moves" notice, which the spec has no slot for
- **the bullet convention that preserves the depth**, which is the one I would most like agreed at the same time as the format. The structure will not keep the measurement and the caveat attached by itself, and if we do not write the convention down the substance quietly drains out of these entries. You write most of the source material, so it has to be one you would actually want to follow.

One thing #323 records that surprised me while checking: the spec's header template asserts semver, and SMDA does not follow it — v4.3.4 shipped a whole new default-off pass as a patch, v3.0.2 added two config options as a patch, and v4.0.0 was an addition rather than a break. Worth stating the real scope instead, which in this project is `ESCAPER_DOWNWARD_COMPATIBILITY` and the `*_PIC_HASH_ESCAPE_VERSION` markers — those are what MCRIT actually reads.

Unrelated to both: #299 and #300 are now aimed at **v4.6.0** rather than this release. Details on those PRs; the short version is that #300's corpus table predates #304/#307/#309/#310/#311 and needs re-running before it can be read, and ten engine changes under `feat(core)` is a minor-bump shape anyway. No rush on either — better measured than fast.

PR #320, opened 2026-09-08
refactor(common): type the attributes the dropped annotations rest on

danielplohmann:
First of the two items left open in #317, both of which needed work upstream of the annotation rather than the annotation itself.

## What #302 actually hit

`getInstructionsForBlock` lost its `-> List[SmdaInstruction]` to satisfy ty 0.0.74's `unsound-return-statement`. The declaration was not the problem: `SmdaFunction.blocks` was `Dict[int, List[Any]]`, so the return could not have been anything better. Typing `blocks` accurately — both population sites build `SmdaInstruction` objects — restores the return annotation for free.

## The `Optional[str]` join, and why it refuses rather than defaults

Typing `blocks` surfaces exactly what #317 predicted: `getPicHashSequence` and `getOpcHashSequence` join over `instruction.bytes`, which is `Optional[str]`.

Seven other sites in these files write `bytes or ""`. Those are formatting and length arithmetic, where a blank is harmless. A hash is not that: a blank there would silently hash a *different* instruction sequence and the result would still look like a valid PIC/OPC hash to everything that consumes one, MCRIT included. So this refuses instead, via a `NoReturn` helper — which also means the ternary narrows to `str` without adding a call per instruction on the hot path.

## Root cause, one level down

`SmdaInstruction.mnemonic` and `.operands` were untyped for the same reason. Annotating them makes three Dalvik `.startswith` reads visible; those take the `or ""` the file already uses, which is behaviour-preserving for a count — a None mnemonic starts with nothing either way.

`SmdaReport.base_addr`, `.bitness`, `._derefs_cache` and `SmdaFunction.smda_report` are annotated here because #317's *second* item, narrowing `extract_strings`, cannot be typed without them. That change follows in its own `utility` scope rather than riding along in this one.

## Verification

Behaviour-neutral, measured rather than asserted:

| check | result |
|---|---|
| PIC/OPC hashes over cutwail, komplex, mirai_x64 (394 functions) | bit-identical |
| guarded-path timing | flat (0.1759s → 0.1767s on mirai_x64) |
| full suite | 1947 passed, 2 skipped, 2593 subtests |
| `ty` errors | 0 |

The only diagnostic movement is two categories becoming *more* precise (`None | Unknown` → `SmdaReport | None`), plus **one new warning** at `StringExtractor.py:201`, where `.operands` is now known to be optional. The follow-up PR clears it and returns the count to the master baseline exactly.

Refs #317.

### DIFF
--- a/src/smda/common/SmdaFunction.py
+++ b/src/smda/common/SmdaFunction.py
@@ -5,7 +5,7 @@ import logging
 import re
 import struct
 from operator import itemgetter
-from typing import Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, NoReturn, Optional
 
 from smda.aarch64.AArch64InstructionEscaper import AArch64InstructionEscaper
 from smda.aarch64.definitions import CALL_INS as AARCH64_CALL_INS
@@ -23,6 +23,9 @@ from smda.dalvik.DalvikInstructionEscaper import DalvikInstructionEscaper
 from smda.intel.IntelInstructionEscaper import IntelInstructionEscaper
 
 from .SmdaInstruction import SmdaInstruction
+
+if TYPE_CHECKING:
+    from smda.common.SmdaReport import SmdaReport
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


danielplohmann:
The Malpedia benchmark ran on this one (same-repo branch, so the secret is available) and gives the corpus-level confirmation the local fixtures cannot:

| | |
|---|---|
| Correctness | **PASS — 0 of 155 common files differ** |
| Functions | 175,946, identical on both sides |
| Determinism | PASS, 3 runs per side |

That is the assertion this change needed: a typing pass that touches the PIC/OPC hash path and three Dalvik mnemonic reads should move nothing, and across 155 samples it moves nothing.

One number worth stating rather than burying: the paired timing reads **−3.70% (95% CI [−4.46%, −2.01%], Wilcoxon p = 0.0000)**, and the harness marks it *inconclusive — within cross-runner noise (±4.7%)*. I am treating it as noise rather than as a regression, for three reasons: base and PR are timed on separate runners, the individual run-pair matrix spans −17.21% to **+1.20%**, and the only added runtime work is one `is not None` per instruction on the `escaper is None` hash path, measured locally at +0.5% on that loop alone (0.1759s → 0.1767s over mirai_x64). The Dalvik `or ""` reads do not execute on this corpus at all. If it were real, it would have to come from somewhere this diff does not touch.

PR #321, opened 2026-09-08
refactor(utility): narrow extract_strings to the tuple it actually yields

### LINKED ISSUE #317: Collected review follow-ups from the open-PR triage pass

danielplohmann:
Second and last of the items left open in #317. Depends on #320, which is now in as `7a55dce`.

## Why this one could not be a one-line change

#302 weakened `extract_strings` to `Iterator[Tuple[str, Any, Any, str]]`. The two `Any`s are an instruction offset and a data address, and all three yield sites produce `(str, Optional[int], int, str)` — so the signature looks like a one-line correction.

It is not, and #317 recorded why: `read_string`, `read_go_string` and `derefs` were untyped, so every yield inferred as `tuple[Unknown, ...]`. Typing those three is the actual work, and it reaches one level further than #317 expected — they read `buffer`, `base_addr` and two caches off the report, so the annotations only hold once `SmdaReport.base_addr`, `.bitness` and `._derefs_cache` are declared. That is what #320 was for.

## Three optionals that were latent are now explicit

- **`base_addr` joins the existing `buffer is None` guard** in `read_bytes`, `detect_ascii_len`, `detect_unicode_len` and `read_string`. A report with no base address cannot turn a VA into an offset at all, so it is the same unusable state the buffer check already refuses, and each site keeps the answer it already gave for an unusable report.
- **`extract_strings` returns early when the function carries no report**, instead of reaching an `AttributeError` inside `read_string`.
- **the Go stack-string path takes `(... .operands or "")` before splitting**, falling through to the String-struct branch exactly as a non-matching operand list already did. This also clears the one warning #320 introduced.

`read_bytes`, `detect_ascii_len` and `detect_unicode_len` are annotated too — same call chain, and the yields cannot be proven without them.

## Verification

Behaviour-neutral, measured:

| check | result |
|---|---|
| extracted strings over cutwail, komplex, mirai_x64, njrat, rust_pe_gnu | **3,412 strings, identical count and content on every fixture** |
| full suite | 1947 passed, 2 skipped, 2593 subtests |
| `ty` errors | 0 |
| `ty` diagnostics | back to the master baseline exactly |

The only diagnostic movement across #320 and this PR together is two categories becoming more precise (`None | Unknown` → `SmdaReport | None`).

Closes #317.

### DIFF
--- a/src/smda/utility/StringExtractor.py
+++ b/src/smda/utility/StringExtractor.py
@@ -1,10 +1,11 @@
 import re
 import string
 import struct
-from typing import Any, Iterator, Tuple
+from typing import Iterator, List, Optional, Tuple
 
 from smda.common.ExceptionHandling import reraise_non_operational_exception
 from smda.common.SmdaFunction import SmdaFunction
+from smda.common.SmdaReport import SmdaReport
 
 _IS_PRINTABLE_CHAR_CODE = tuple(chr(char) in string.printable for char in range(256))
 _ASCII_RE = re.compile(b"[\x09-\x0d\x20-\x7e]*")
@@ -14,15 +15,18 @@ _UNICODE_RE = re.compile(b"(?:[\x09-\x0d\x20-\x7e]\x00)*")
 # https://github.com/mandiant/capa/blob/v4.0.0/capa/features/extractors/smda/insn.py
 
 
-def read_bytes(smda_report, va, num_bytes=None):
+def read_bytes(smda_report: SmdaReport, va: int, num_bytes: Optional[int] = None) -> bytes:
     """
     read up to MAX_BYTES_FEATURE_SIZE from the given address.
     """
 
-    rva = va - smda_report.base_addr
     buffer = smda_report.buffer
-    if buffer is None:
+    base_addr = smda_report.base_addr
+    # base_addr joins the existing buffer check rather than getting its own: a report with no
+    # base address cannot turn a VA into an offset at all, which is the same unusable state
+    if buffer is None or base_addr is None:
         raise ValueError("buffer is empty")
+    rva = va - base_addr
     buffer_end = len(buffer)
     max_bytes = num_bytes if num_bytes is not None else 0x100
     if rva + max_bytes > buffer_end:
@@ -31,7 +35,7 @@ def read_bytes(smda_report, va, num_bytes=None):
         return buffer[rva : rva + max_bytes]
 
 
-def derefs(smda_report, p):
+def derefs(smda_report: SmdaReport, p: int) -> Iterator[int]:
     """
     recursively follow the given pointer, yielding the valid memory addresses along the way.
     useful when you may have a pointer to string, or pointer to pointer to string, etc.
@@ -45,7 +49,7 @@ def derefs(smda_report, p):
         yield from cache[p]
         return
 
-    chain = []
+    chain: List[int] = []
     current = p
     depth = 0
     word_size = 8 if smda_report.bitness == 64 else 4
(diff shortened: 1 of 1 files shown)


danielplohmann:
Corpus benchmark agrees with the local fixtures: **0 of 155 files differ**, 175,946 functions, determinism PASS on both sides. Extracted strings were already identical over five bundled fixtures (3,412 strings); this says the same thing at corpus scale.

**A note on the timing half, because these two PRs together make the point better than either does alone.** Both are behaviour-neutral typing changes with no meaningful runtime component, and the harness's paired verdicts were:

| PR | median paired speedup | verdict |
|---|---|---|
| #320 | **−3.70%** (95% CI [−4.46%, −2.01%], Wilcoxon p = 0.0000) | inconclusive — within noise (±4.7%) |
| #321 | **+14.78%** (95% CI [+14.07%, +15.99%]) | **PR is faster** |

Neither is real. `extract_strings` runs once per function over data refs; it cannot make a corpus 14.78% faster, and #320 cannot make it 3.70% slower. Both figures are tight confidence intervals around cross-runner hardware variance, and #320's even carried `p = 0.0000`.

Worth carrying forward when a *genuine* perf claim comes through this gate: a narrow CI and a small p-value here are not evidence that a difference is real, because base and PR are timed on separate runners and the between-runner spread swamps both. Added to #322, where the question of what this benchmark can and cannot settle is already open.

Issue #323, opened 2026-09-08
Adopt keep-a-changelog: agreeing the fine tuning before the rewrite


danielplohmann:
@r0ny123 suggested adopting [keep-a-changelog](https://github.com/olivierlacan/keep-a-changelog), and **the decision is yes** — for SMDA and for MCRIT, so the two stay consistent. What is open is the fine tuning, and a few of those choices are load-bearing rather than cosmetic. Opening this for your input on the six below before anything is written.

## Why, in one paragraph

Not the formatting. **The `Unreleased` section.** Right now #319 exists as a long-lived draft PR purely because there is nowhere to accumulate entries, and you [declined to touch `CHANGELOG.md` in #311](https://github.com/danielplohmann/smda/pull/311) for exactly that reason — "its entries are per released version and there is no unreleased section, so adding one means choosing a version number, and that is yours." The consequence is that every changelog line gets written by one person at release time, reconstructed from merge commits days later. That is the "commit log diff" anti-pattern the spec warns about, done by hand. With an `Unreleased` section each PR writes its own line at merge time, while whoever made the change still remembers the caveat.

---

## 1. History: don't convert it

192 entries going back to 2018-07-01. Converting them is large, low-value, and translation loses nuance.

Proposal: start the new format at `[Unreleased]`, and put everything from v4.5.1 down under a final `## Older releases` heading, **verbatim**. The spec's "an entry for every version" is still satisfied — they are all still there, just in the old one-line shape.

While we are in there, three genuine filing errors that the strict format would surface anyway:

| | |
|---|---|
| `v1.0.0` (2018-07-01) is filed at line 186, **above** the v1.9.16–v1.9.11 block | plainly misfiled; belongs at the bottom |
| `v4.4.0` (2026-07-26) listed above `v4.3.11` (2026-07-28) | legitimate — a patch on an older line released after the newer minor |
| `v3.1.0` (2026-05-26) above `v3.0.2` (2026-06-01) | same |

Only the first is a mistake. But the last two force a decision: **order by version or by date?** The spec says latest first without saying which. I lean version order with the real date kept, since that is what a reader of a maintenance release expects.

## 2. Do NOT claim semver — this is the one I would get wrong by default

The keep-a-changelog header template asserts adherence to Semantic Versioning. **SMDA does not follow it**, and saying so would be a false claim rather than an aspiration:

[...]

PR #327, opened 2026-09-10
feat(common): refuse a candidate the image declares interior, not only a gap one

### LINKED ISSUE #324: AArch64 ELF: 96% of function-start false positives fall inside a range .eh_frame already declares
On the 72-cell built AArch64 ELF corpus, **5,899 of master's 6,129 function-start false positives (96.2%) fall strictly inside a function range the image's own `.eh_frame` already declares**. They are not unnamed functions — they are the engine splitting a routine the unwind data delimits. **Zero of the 59,365 true positives lies strictly inside a declared FDE**, so refusing them costs no recall on this corpus.

Measured against `6240b74`, both arms on the same machine, exact-address match, truth is the unstripped link's `STT_FUNC` plus PLT:

| arm | TP | FP | FN | PPV | TPR | F1 |
|---|---|---|---|---|---|---|
| master `6240b74` | 59,365 | 6,129 | 961 | 90.642 | 98.407 | 94.365 |
| + refuse a candidate strictly interior to a declared FDE | 59,365 | **230** | 961 | **99.614** | 98.407 | **99.007** |

Micro (pooled) figures. The same tree's macro means are PPV 91.497 / TPR 97.705, which is the master baseline #300 reports, reproduced here independently.

## Why this is not already covered

#300 adds `USE_ELF_FDE_INTERIOR_GAPS`, which refuses on exactly this evidence, and it is worth −1,846 false positives there. But it is reached only from the gap scan — it tests `self.gap_pointer`, so candidates from the prologue scan, from branch targets, and from the tailcall paths never meet it. Measured on that branch, the interior population that survives it is:

| tree / arm | FP | of which strictly interior to a declared FDE |
|---|---|---|
| master `6240b74` | 6,129 | 5,899 |
| #300 branch, default config | 2,484 | 2,245 |
| #300 branch + `USE_ELF_EH_FRAME_CANDIDATES` | 2,399 | 2,159 |

So the rule as written catches under half of what its own evidence source could refuse. This issue is about widening where it is consulted, not about a new heuristic — and it is complementary to #300 rather than a competing version of it.

## What is left afterwards

The 230 survivors on master are a short list, and none of them needs `.eh_frame`:

| class | n | note |
|---|---|---|
| data decoded as code | 106 | not in an executable section at all — no section, `.gcc_except_table`, `.rodata`, `.dynsym`, `.rela.dyn` |
| PLT entries | 60 | 58 of them are PLT0, the resolver trampoline, which is never a function |
| erratum-843419 stub headers | 39 | see below |
| other `.text`, outside any FDE | 25 | |

The third class is worth naming because it is linker furniture rather than anything the compiler emitted. When `ld` applies the Cortex-A53 erratum 843419 workaround it emits a stub section, pads it to a multiple of 0x1000, and writes `b .+0x1000 ; nop` at its head so fall-through skips the veneers. That 8-byte header is what gets booked as a function. Confirmed structurally across all 43 sites in the corpus: 43/43 are followed by a veneer at header+8 and then zero padding to the page boundary, and 43/43 have zero incoming references. One worked example:

```
0xbdffc  adrp x0, #0x3df000     ADRP at page offset 0xffc — the erratum trigger
0xbe000  ldr  x0, [x0, #0xfd0]
0xbe004  b    #0xbe938          the load, replaced by a branch to the veneer
0xbe008  str  x1, [sp, #0x188]  resume point
...
0xbe930  b    #0xbf930          stub-section header  <-- booked as a function start
0xbe934  nop
0xbe938  ldr  x1, [x0]          the relocated load
0xbe93c  b    #0xbe008          branch back
```

In 37 of the 43 the stub is dead — the veneer was allocated but the erratum site was never patched, so the original instruction is still in place. The symbol table names the veneer in exactly the 6 live cases and names the header in none of the 43, so the ground truth is right to omit them.

## Caveats

The rule is ELF-only, so PE and Mach-O corpora have to come out bit-identical as the control, the same shape of control #300 uses for its own `.eh_frame` rules. It also inherits whatever the FDE decoder gets wrong: an over-long decoded range would swallow a real function start. The 0-of-59,365 result is evidence that does not happen on this corpus, and the decoder was cross-checked against an independent DWARF implementation (below), but it is the failure mode to watch.

The corpus is C/C++ built by one gcc generation. It says nothing about hand-written assembly with no unwind data, which is the population where an interior test has nothing to consult.

## How this was measured

Three legs that do not share an implementation:

- SMDA's own `EhFrameDecoder`, the parser the existing flags use.
- `llvm-objdump --dwarf=frames`, an independent DWARF implementation. It agrees with `EhFrameDecoder` on **all 55,901 FDE records across all 72 binaries, with zero cells differing**, and reproduces the table above exactly.
- IDA Pro 9.3 autoanalysis over the 72 stripped binaries, as a function-start oracle independent of both. It places the great majority of the interior population inside a function body rather than at a start, and finds an incoming call reference for none of it.

Corpus, truth and per-binary predicted sets are the same ones referenced from #300.

r0ny123:
Closes #324.

`USE_ELF_FDE_INTERIOR_GAPS` refuses a candidate that opens strictly inside a range the image's own `.eh_frame` declares. It is reached only from the gap scan — it tests `self.gap_pointer` — so candidates from the prologue scan, from branch targets and from the tailcall paths never meet it. On the 72-cell AArch64 ELF corpus, 2,245 of the 2,484 false positives still standing after #300 are interior to a declared range, and **not one of them is a gap candidate**: 1,213 come from the initial seeding scans and 1,001 carry a call reference.

This asks the same question at the point analysis is about to begin on a candidate, whatever seeded it. Nothing new is read from the image and no heuristic is added; the evidence and the rule are the ones already shipped.

## The guards

Both of the gap rule's conditions apply unchanged. A PLT is exempt, because the whole table sits under one FDE and the range test would read every stub after the first as interior to it. And the range's own start has to be a recovered function, because an FDE can begin in the alignment padding ahead of its function, which leaves the real entry a few bytes in interior to nothing.

**A third is new, and the corpus is what found it.** A declared range can reach past everything its function's control flow actually arrives at. Refusing an address out there discards bytes nothing else claims — along with any reference only those bytes carry. Without this guard the AArch64 corpus loses three true positives, and all three are the same shape: a real function whose only incoming reference sits in the unreached tail of another function's declared range. In `brotli_gcc-arm64_O2-static` the entry point's FDE is `[0x401e40, 0x401e7c)`, its control flow stops before `0x401e78`, and the `bl` at that address is the only thing naming the real function at `0x400600`.

Requiring the owner's own recovered extent to surround the address costs 163 of the refusals and returns all three. It is also most of why Rust comes out unchanged.

## Measured

Against compiler symbol tables, `master` `e9dfe5f` → this branch, both trees run back to back on the same machine. Re-measured after the rebase onto #325; the figures are unchanged, which is expected since that PR's one behaviour change is PE-only and this rule is ELF-only. **No corpus loses a true positive.**

| corpus | n | PPV | TPR | ΔFP | ΔTP |
|---|---|---|---|---|---|
| Built C/C++ AArch64 ELF | 72 | 95.994 → **97.063** | 98.657 | **−683** | 0 |
| Built C/C++ ELF (gcc, clang) | 140 | 98.903 → **98.969** | unchanged | **−77** | 0 |
| Built C/C++ MinGW PE | 120 | bit-identical | bit-identical | 0 | 0 |
| Built Rust (gnu targets) | 24 | bit-identical | bit-identical | 0 | 0 |
| Built Go (pclntab truth) | 23 | bit-identical | bit-identical | 0 | 0 |
| ARM64 Mach-O (`LC_FUNCTION_STARTS`) | 11 | bit-identical | bit-identical | 0 | 0 |
| Malpedia dumps (`.fnmap` truth) | 57 | bit-identical | bit-identical | 0 | 0 |

Micro (pooled) figures; TP and FN are identical in every row. The four bit-identical rows are the control, not filler: 120 PE cells, 23 Go, 11 Mach-O and 57 memory dumps carry no `.eh_frame`, so any movement there would have meant the rule firing where it has no business firing.

**Rust is bit-identical for a reason, not because the rule is inert there.** Its images decode their ranges — 740 FDEs on the first cell — and 25 of that cell's 26 false positives *are* interior to one. The guards decline all of them: over its first four cells, 78 of 94 on the owner's recovered extent and the other 12 on the owner not being recovered.

The rule never refuses an address the image names. Across the four densest AArch64 cells it refuses 4,070 addresses and every one is inferred-only — not a symbol, not an exception record, not language metadata, not a gap or tailcall candidate.

Analysis is slightly faster, because a refused candidate is one nothing then analyses: `zlib_gcc-arm64_O2-static` 1.26s → 1.19s, `lua_gcc-arm64_O2-static` 2.04s → 1.92s (median of 3).

## Tests

Nine new cases in `tests/testFdeInteriorGaps.py`. Eight drive `declaredInteriorOwner` directly, one condition each, and the ninth covers the wiring: it claims one address the fixture does recover and asserts analysis then declines it. Each was checked to fail when the guard it describes is removed, and only that one — dropping the extent guard fails two, the PLT exemption one, the recovered-owner condition one, and removing the call site fails the wiring case.

The bundled fixtures reach the gap scan's own refusal or the collision check before this arm sees anything, which is why the wiring case claims an address rather than asserting over what the fixture happens to drop. An earlier version of it asserted over the recovered set and passed with the feature removed; that is why it is written this way.

This file also had `unittest.main()` sitting above a test class, so running it directly never defined `FdeInteriorGapRuleAArch64Test`. Moved to the end.

- `python -m pytest tests/` → **2,069 passed, 1 skipped, 2,593 subtests**
- `make typecheck` → exit 0, 0 error-level diagnostics
- Diff coverage against this PR's base → **100%** (16 lines, 0 missing)

No bundled fixture baseline moves.

## One sibling, deliberately left out

Three rules read declared evidence at the gap pointer only: the `.pdata` exception range, `USE_LSDA_LANDING_PADS`, and the one widened here. The landing-pad rule has nothing left to reach — of 302 false positives still standing on four AArch64 cells, none is a declared landing pad. The `.pdata` rule does: 6 of 74 remaining false positives on three MinGW PE cells are inside a declared range.

That one is left for its own change. It moves PE, which is exactly the control that makes this one safe to read, and it needs its own measurement over the PE corpora rather than riding on these numbers. Happy to open an issue for it, or fold it in here if you would rather see them together.

## Corrected after review

**The docstring's claim about what the rule can reach was wrong.** It said that inside the owner's recovered extent "the owner already accounts for the address, so nothing is lost". The rule is consulted immediately after the collision check, which tests membership in `code_map` — every byte of every recovered instruction — so every address that reaches it is one the owner's analysis never covered, and `function_borders` is an envelope built from the instruction extremes rather than a coverage map.

It now names the population (alignment holes between blocks, an unreached tail, a data island), names the residual exposure (a real entry carrying no FDE of its own, sitting in one of those holes inside a neighbour's declared range), and says that what bounds it is the measurement rather than the structure. The magnitude stays out of the source: an absolute false-positive count is a property of every other rule in the engine on the day it was taken, while the invariant — no true positive lost — is a property of the mechanism.

**A test-isolation fix found in the same pass.** `analysedFixture()` patches `declaredInteriorOwner` onto the class and restores it in a `finally`, but the sample write and the disassembler construction sat outside the `try`; either raising would leave the claim installed for every later test in the process. Everything after the patch is inside the `try` now.

Full suite on the current tip: **2,119 passed, 1 skipped, 2,596 subtests**, `ruff` clean, `make typecheck` exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ past roughly six lines it belongs in the PR the entry links.
   not only a gap-scan candidate -- the gap pointer reaches only what the gap scan walks to, so the prologue,
   reference and symbol scans seeded the rest unchecked. A procedure linkage table is exempt, the declared
   range's own start has to be a recovered function, and that owner's recovered extent has to surround the
-  address. *Measured on `a043145` against compiler symbol tables:* 72 AArch64 ELF cells 95.994 -> 97.063 PPV
+  address. *Measured on `857081f` against compiler symbol tables:* 72 AArch64 ELF cells 95.994 -> 97.063 PPV
   (-683 false positives) and 140 built C/C++ ELF cells 98.903 -> 98.969 (-77), both at identical true positives
   and false negatives; the PE, Go, ARM64 Mach-O and 57 malpedia cells are bit-identical, which is the control
   that it reaches only images carrying an `.eh_frame`. *Not reproducible from the bundled fixtures, which
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,16 @@ past roughly six lines it belongs in the PR the entry links.
 
 ### Changed
 
+- **(common)** Refuse a candidate from any source that the image's `.eh_frame` declares interior to a function,
+  not only a gap-scan candidate -- the gap pointer reaches only what the gap scan walks to, so the prologue,
+  reference and symbol scans seeded the rest unchecked. A procedure linkage table is exempt, the declared
+  range's own start has to be a recovered function, and that owner's recovered extent has to surround the
+  address. *Measured on `a043145` against compiler symbol tables:* 72 AArch64 ELF cells 95.994 -> 97.063 PPV
+  (-683 false positives) and 140 built C/C++ ELF cells 98.903 -> 98.969 (-77), both at identical true positives
+  and false negatives; the PE, Go, ARM64 Mach-O and 57 malpedia cells are bit-identical, which is the control
+  that it reaches only images carrying an `.eh_frame`. *Not reproducible from the bundled fixtures, which
+  produce 0 refusals.* (#327)
+
 ### Deprecated
 
 ### Removed
@@ -83,6 +93,10 @@ past roughly six lines it belongs in the PR the entry links.
 ### Security
 
 ### Compatibility
+
+- Recovery output moves on ELF images carrying an `.eh_frame`: an address a declared FDE range covers is no
+  longer reported as a function start unless it is that range's own start. No true positive was lost on the
+  corpora this was measured over. (#327)
 
 ## Older releases
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ past roughly six lines it belongs in the PR the entry links.
   not only a gap-scan candidate -- the gap pointer reaches only what the gap scan walks to, so the prologue,
   reference and symbol scans seeded the rest unchecked. A procedure linkage table is exempt, the declared
   range's own start has to be a recovered function, and that owner's recovered extent has to surround the
-  address. *Measured on `857081f` against compiler symbol tables:* 72 AArch64 ELF cells 95.994 -> 97.063 PPV
+  address. *Measured on `0a0a4c6` against compiler symbol tables:* 72 AArch64 ELF cells 95.994 -> 97.063 PPV
   (-683 false positives) and 140 built C/C++ ELF cells 98.903 -> 98.969 (-77), both at identical true positives
   and false negatives; the PE, Go, ARM64 Mach-O and 57 malpedia cells are bit-identical, which is the control
   that it reaches only images carrying an `.eh_frame`. *Not reproducible from the bundled fixtures, which
--- a/src/smda/SmdaConfig.py
+++ b/src/smda/SmdaConfig.py
@@ -211,6 +211,21 @@ class SmdaConfig:
     # table names `dispatch`, and aarch64_static drops 0x400350 and 0x40DF30, both mid-function
     # instructions inside a declared FDE. None of the six carries a symbol or is a declared
     # start, so both baselines moved toward the truth.
+    # The same evidence also refuses a candidate from any other source, at the point analysis
+    # would begin on it, since the gap pointer reaches only what the gap scan walks to and the
+    # seeding scans reach the rest. That arm carries a third condition the gap scan does not
+    # need: the owner's own recovered extent has to surround the address. A declared range can
+    # reach past everything its function's control flow arrives at, and refusing an address out
+    # there discards bytes nothing else claims along with any reference only those bytes carry
+    # - without it the AArch64 corpus loses three functions reached through exactly that shape.
+    # Measured against compiler symbol tables, no corpus losing a true positive:
+    #   72 AArch64 ELF cells   PPV 95.994 -> 97.063  -683 FP at identical TP and FN
+    #   140 built C/C++ ELF    PPV 98.903 -> 98.969   -77 FP at identical TP and FN
+    # The 120 MinGW PE cells, 23 Go cells, 11 ARM64 Mach-O cells and all 57 malpedia dumps are
+    # bit-identical, the control that it reaches only images carrying an .eh_frame. Rust is too,
+    # and not because the rule is inert there: its images decode their ranges and 25 of 26 false
+    # positives on the first cell are interior to one, but the two conditions above decline all
+    # of them - 78 of 94 on the owner's extent, the rest on the owner not being recovered.
     USE_ELF_FDE_INTERIOR_GAPS = True
     RESOLVE_REGISTER_CALLS = True
     # resolve "call/jmp dword ptr [<reg> + <disp>]" against a runtime-built import table and
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -725,7 +725,15 @@ class FunctionCandidateManager:
         surround the address. An FDE can reach past everything its function's control flow
         arrives at, and refusing an address out there discards bytes nothing else claims --
         along with any reference only those bytes carry, which costs real functions elsewhere.
-        Inside the extent the owner already accounts for the address, so nothing is lost.
+        Inside the extent, the address is never one the owner decoded: a candidate landing on
+        recovered code is refused a step earlier against a byte-level code map, and
+        `function_borders` records only the extremes of that code, not its coverage. What
+        reaches here sits in the holes -- alignment between blocks, an unreached tail, a data
+        island -- so the exposure is a real entry with no FDE of its own in one of them, which
+        loses whatever reference only its bytes carry. An entry that carries its own FDE is
+        never interior to it, so the shape needs a routine covered by a neighbour's range, and
+        nothing in the format forbids that; what bounds it is measurement rather than
+        structure, no true positive lost across the ELF corpora this was measured over.
         """
         if not self.config.USE_ELF_FDE_INTERIOR_GAPS:
             return None
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -710,6 +710,35 @@ class FunctionCandidateManager:
             index -= 1
         return None
 
+    def declaredInteriorOwner(self, addr):
+        """The recovered function whose declared `.eh_frame` range contains `addr`, or None.
+
+        The same evidence the gap scan already refuses on, asked at the point a candidate from
+        any source is about to be analysed rather than only where the gap pointer reaches. Both
+        of that rule's guards apply unchanged. A PLT is exempt because the whole table sits
+        under one FDE, so the range test reads every stub after the first as interior to the
+        first. And the range's own start has to be a recovered function, because an FDE can
+        begin in the alignment padding ahead of its function, which leaves the real entry a few
+        bytes in interior to nothing.
+
+        A third guard the gap scan does not need: the owner's own recovered extent has to
+        surround the address. An FDE can reach past everything its function's control flow
+        arrives at, and refusing an address out there discards bytes nothing else claims --
+        along with any reference only those bytes carry, which costs real functions elsewhere.
+        Inside the extent the owner already accounts for the address, so nothing is lost.
+        """
+        if not self.config.USE_ELF_FDE_INTERIOR_GAPS:
+            return None
+        if self.isInDeclaredPltSection(addr):
+            return None
+        containing = self.declaredFdeRangeContaining(addr)
+        if containing is None or containing[0] not in self.disassembly.functions:
+            return None
+        borders = self.disassembly.function_borders.get(containing[0])
+        if borders is None or not borders[0] <= addr < borders[1]:
+            return None
+        return containing[0]
+
     def opensInsideDeclaredFdeRange(self, addr):
         """Whether `addr` falls inside a declared FDE range without being that range's start."""
         return self.declaredFdeRangeContaining(addr) is not None
--- a/src/smda/common/RecursiveDisassembler.py
+++ b/src/smda/common/RecursiveDisassembler.py
@@ -261,6 +261,13 @@ class RecursiveDisassembler:
             # state.getBlocks(); this collision path is unreachable from the gap pass today, so
             # the change is behavior-neutral (output stays byte-for-byte identical).
             return state
+        declared_owner = self.fc_manager.declaredInteriorOwner(start_addr)
+        if declared_owner is not None:
+            self.fc_manager.updateAnalysisAborted(
+                start_addr,
+                f"inside the declared range of function 0x{declared_owner:08x}",
+            )
+            return state
         blocks_processed = 0
         while state.hasUnprocessedBlocks():
             blocks_processed += 1
--- a/tests/testFdeInteriorGaps.py
+++ b/tests/testFdeInteriorGaps.py
@@ -15,6 +15,7 @@ import unittest
 import lief
 
 from smda.common.EhFrameDecoder import decodeEhFrameFdeRanges
+from smda.common.FunctionCandidateManager import FunctionCandidateManager
 from smda.Disassembler import Disassembler
 from smda.SmdaConfig import SmdaConfig
 
@@ -86,10 +87,6 @@ class FdeInteriorGapRuleTest(unittest.TestCase):
         self.assertEqual((off & named) - on, set())
 
 
-if __name__ == "__main__":
-    unittest.main()
-
-
 class FdeInteriorGapRuleAArch64Test(unittest.TestCase):
     """The rule lives in the shared candidate manager and both gap scans consult it.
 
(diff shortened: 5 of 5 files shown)


danielplohmann:
Reviewed against the source rather than the figures, since the figures are not reproducible here. The guards each have their own case and each case fails without its guard — the PLT exemption, the recovered-owner test, the recovered-extent test, the range-start exclusion and the flag control are eight unit cases, and `testAnalysisItselfDeclinesWhatTheRuleClaims` is what shows analysis consults the rule at all rather than the gap scan doing the work. Claiming one address rather than asserting over what the fixture happens to drop is the right way to write that, given what I measured below.

## One thing in the docstring does not hold, and it is the sentence the next person will lean on

> Inside the extent the owner already accounts for the address, so nothing is lost.

The hook sits immediately after `state.isProcessedFunction()`, which is `start_addr in self.disassembly.code_map` — every byte of every recovered instruction. So any candidate that lands on code the owner actually decoded is already gone by the collision path, and **the only addresses that can reach `declaredInteriorOwner` are ones the owner's analysis never covered**: alignment holes between its blocks, an unreached tail inside its envelope, a data island. `function_borders` is a `(min, max)` envelope built from the instruction extremes in `FunctionAnalysisState`, not a coverage map.

Confirmed on the two refusals I could provoke: both report `in_code_map=False`.

That does not make the rule wrong — it is the same evidence, and your 0 true positives lost across 212 ELF cells is what says it costs nothing. But the *reason* stated is not the reason it works, and the residual exposure is the one the sentence denies: a real entry with **no FDE of its own**, sitting in a hole inside another routine's declared range and recovered envelope, loses whatever reference only its bytes carry — the same shape as the three functions the extent guard rescued, one scale down. Anything carrying its own FDE is safe, since `declaredFdeRangeContaining` excludes a range's own start, which `testARangeStartIsNotInteriorToItself` pins.

Worth rewriting to say what the population is and that the measurement is what bounds the risk. A sentence, not a change of code.

## What the bundled fixtures say about this rule: nothing

Instrumented `declaredInteriorOwner` over every bundled ELF that reaches it:

| fixture | refusals |
|---|---|
| `elf_cxx_landing_pads_x64_xored` | 0 |
| `elf_cxx_landing_pads_arm64_xored` | 0 |
| `elf_cet_landing_pads_x64_xored` | 0 |
| `elf_ehframe_hdr_x64_xored` | 0 |
| `aarch64_static_xored` | 0 |

Exactly as your test's docstring says, and I am not asking for a fixture that changes it — by the standard we just settled on #322 item 2, the mechanism is pinned in-repo and the magnitude rests on your corpora. It does mean the v4.7.0 changelog entry has to say so: under the provenance rule from #323 the −683 and −77 go in as measured on your corpora and not reproduced here, the way v4.5.1 marked #310's and #312's figures.

## Sequencing

I am holding this rather than merging it now, because #329 is stacked on it and has a blocking finding of its own — it breaks `testArm64PeInteriorGapFixture` on current master, details over there. Since the fix lands in the shared helper this PR introduces, doing both in one rebase over `a043145` is less work than merging this and then rebasing that. Suite on your two plus #335 stacked over master: 2,127 passed, 2 skipped, 2,596 subtests, and the single failure is the #333 one described on #329.

r0ny123:
You are right, and it is worse than a loose sentence — it stated the opposite of what the code does.

Confirmed against the source here: `isProcessedFunction()` is `start_addr in self.disassembly.code_map`, `code_map` is written per byte of every recovered instruction in `FunctionAnalysisState`, and `function_borders[start] = (fn_min, fn_max)` is built from the instruction extremes in the same loop. So the collision path has already taken every address the owner decoded, and the only ones that can reach the rule are precisely the ones it did not.

The paragraph now says what the population is and what bounds the risk:

> Inside the extent, the address is never one the owner decoded: a candidate landing on recovered code is refused a step earlier against a byte-level code map, and `function_borders` records only the extremes of that code, not its coverage. What reaches here sits in the holes — alignment between blocks, an unreached tail, a data island — so the exposure is a real entry with no FDE of its own in one of them, which loses whatever reference only its bytes carry. An entry that carries its own FDE is never interior to it, so the shape needs a routine covered by a neighbour's range, and nothing in the format forbids that; what bounds it is measurement rather than structure, no true positive lost across the ELF corpora this was measured over.

The magnitude stays out of the docstring deliberately — an absolute false-positive count is a property of every other rule in the engine on the day it was taken, and belongs in the entry that dates it. The invariant is what keeps its meaning in a comment.

## One more, from re-reading the diff after yours

`analysedFixture()` patches `declaredInteriorOwner` onto the class and restores it in a `finally`, but the sample write and the disassembler construction sat outside the `try`. Either raising would leave the claim installed for every later test in the process — one failure becoming a cascade that does not name its cause. Everything after the patch is inside the `try` now.

## On the changelog

Agreed, and it is yours to write at release: under the provenance rule the −683 and −77 go in as measured on corpora not bundled here, the way v4.5.1 marked the earlier two. Nothing in this PR's tests rests on them — the bundled ELF fixtures produce 0 refusals, exactly as the test docstring says.

## Sequencing

#329's blocking finding is fixed and pushed; the shared-helper change lands there, on top of this. Master is merged into both rather than rebased — I would rather not force-push over a branch you have already reviewed, so taking the rebase at merge is yours.

Full suite green on this tip: **2,119 passed, 1 skipped, 2,596 subtests**. `ruff check .` and `ruff format --check .` clean, `make typecheck` exit 0.

r0ny123:
Rebased over `a043145` as you asked, replacing the merge commit — three linear commits now, and the tree is unchanged by it (`git diff` against the pre-rebase tip is empty), so the suite, lint and typecheck results reported above still stand as measured.

```
a4c1513 test(tests): keep a failed claim from leaking onto the class
12ec24a docs(common): state what the interior refusal can actually reach
c4e3ca1 feat(common): refuse a candidate the image declares interior, not only a gap one
```

#329 is rebased onto this tip, so it still stacks cleanly and has to merge after it.

danielplohmann:
The docstring now says what the rule can actually reach, which is the correction I was after. The old sentence — "inside the extent the owner already accounts for the address, so nothing is lost" — was the one claim in the PR that did not hold: the hook sits immediately after `state.isProcessedFunction()`, which is membership in a byte-level code map, so every address that reaches the rule is one the owner's analysis never covered, and `function_borders` is a `(min, max)` envelope rather than a coverage map.

Naming the residual exposure rather than denying it is the better docstring. The shape is exactly as you state it: a real entry with no FDE of its own, sitting in a hole inside a neighbour's declared range and recovered envelope, losing whatever reference only its bytes carry. And the observation that an entry carrying its own FDE is never interior to it — a range's own start being excluded — is what narrows the exposure to that one shape. What bounds it is the measurement over your ELF corpora, not the format, and saying so is the honest version.

`a4c1513` is a real catch independent of any of this. `analysedFixture()` patched `declaredInteriorOwner` on the class with the sample write and the disassembler construction outside the `try`, so either raising would leave the claim installed for every later test in the process — one failure becoming a cascade that does not name its cause. Worth having regardless of this rule.

Standing note, unchanged: #327 fires zero times on all five bundled ELF fixtures, exactly as its own test docstring says, so the in-repo evidence is the unit cases and the -683/-77 rests on your corpora. That is fine, and it is what the entry has to say.

Verified stacked over `a043145` with #336, #337, #329 and #335: 2,142 passed, 2 skipped, 2,596 subtests; `ruff` clean; `ty` exit 0.

Held with #329 as agreed, so one rebase does both. Same request as there: an `## [Unreleased]` entry once #336 and #337 are in, marking the -683/-77 as measured on your corpora and not reproduced here.

r0ny123:
Nothing open here, and the `## [Unreleased]` entry you ask for is already on the branch — pushed before this comment landed, so the timing is just unlucky.

One thing your standing note made me re-check about my own entry, and it was wrong: it credited the −683/−77 to `a043145`, which is where the branch sat rather than where the numbers came from. They were carried forward from the original measurement. Re-run against compiler symbol tables on `857081f`, the tree this actually lands on:

| corpus | n | before → after | delta | true positives lost |
|---|---|---|---|---|
| AArch64 ELF | 72 | 95.994 → 97.063 PPV | **−683 FP** | **0** |
| built C/C++ ELF | 140 | 98.903 → 98.969 PPV | **−77 FP** | **0** |
| MinGW PE | 120 | — | — | 120/120 bit-identical |
| Go | 23 | — | — | 23/23 bit-identical |
| ARM64 Mach-O | 11 | — | — | 11/11 bit-identical |
| malpedia dumps | 57 | — | — | 57/57 bit-identical |

Every figure reproduces exactly, so the entry now names `857081f` and the whole sentence is verified rather than inherited. They did not decay because nothing landing since moves recovery — #326, #330, #331, #332 and #333 are tests, logging and documentation — which is a reason to name the tree, not a reason to skip the check.

The entry says the rule is not reproducible from the bundled fixtures, since all five produce 0 refusals, exactly as you note.

#329 is restacked on this tip and pushed; both are linear over `857081f` with no merge commits.

r0ny123:
`master` merged in — #339's `### Fixed` and `### Compatibility` bullets collide with this one's in `## [Unreleased]` by construction, and both are kept.

**Worth more than the conflict: #339 touches the same gap path this does, so I re-ran the figures against it rather than assuming they carried.** A candidate this rule refuses is not in `code_map`, which is exactly the condition #339's resume keys on, so an interaction was plausible.

There is none. On `0a0a4c6`:

| corpus | n | before → after | delta | true positives lost |
|---|---|---|---|---|
| AArch64 ELF | 72 | 95.994 → 97.063 PPV | **−683 FP** | **0** |
| built C/C++ ELF | 140 | 98.903 → 98.969 PPV | **−77 FP** | **0** |
| MinGW PE | 120 | 93.191 → 93.191 | — | 120/120 bit-identical |

Identical to the run on `857081f`, so the entry now names `0a0a4c6`.

One detail that argues for the rule we just wrote into the changelog: the PE control's *own baseline* moved 93.204 → 93.191 between those two trees, purely because #339 landed. The figures being compared did not change, but the tree they sit against did — which is precisely what naming the tree is for.

danielplohmann:
Only thing left here is a merge of master, and the conflict is mine rather than yours.

You based this on `0a0a4c6`; since then `8a4d994` (#340) and `130fa02` (#335) have landed, and both wrote into the same `Unreleased` subsections this PR does. So `CHANGELOG.md` is the only conflicting file, and the resolution is purely additive — keep both sides in each block. Nothing in `src/` collides.

I resolved and verified it locally to be sure I was not asking you to chase something real:

- source diff against master is **byte-identical** to the tree I reviewed, so the review stands;
- full suite **2138 passed, 2 skipped, 2596 subtests**; `ruff check` and `ruff format --check` clean; `ty` at 274, which is master's 270 plus the four `None | Unknown` warnings on this PR's own new lines — the same family and the same count I measured when I reviewed it.

I would have pushed the merge to your branch myself, but pushing to a contributor's fork is not something I will do without asking first, so it is yours. `git merge upstream/master`, keep both sides in the two conflict blocks, done.

Your two follow-up commits are the right call, and the second one especially — re-measuring against the base you actually sit on rather than carrying the old figure forward is exactly the provenance rule from #323 doing its job the first time it could.

Once this and #329 are in, v4.7.0 is the next thing.

r0ny123:
Merged, both sides kept, pushed — and thanks for resolving it locally first rather than sending me after something that might not have been additive.

One block rather than two here: `### Compatibility` merged cleanly because #339's bullet and this one land in order, so only `### Changed` collided, with #340. `## [Unreleased]` now carries six bullets across five PRs with no duplicates, and nothing in `src/` moved — the source diff against master is this PR's own three files and nothing else.

Verified on the merged tree: **2,139 passed, 1 skipped, 2,596 subtests**; `ruff check .` and `ruff format --check .` clean; `ty` at 267, which is this box's master baseline of 263 plus the same four `None | Unknown` warnings on the new lines — the same +4 you measured, on a different absolute. Your 2,138/2 against my 2,139/1 is one platform-conditional skip either way, so the totals agree.

#329 is merged over this and pushed too; its `### Fixed` bullet was the only extra collision and it resolved the same way. Suite there **2,156 passed, 1 skipped**.

Noted on `make test` — the fast tier is the default now, so `make test-all` is what I will run before pushing rather than the old target.

danielplohmann:
Re-verified at `b2671e3`: the source diff against master is still byte-identical to the tree I reviewed, so the merge changed nothing but the changelog. Both entries are in place — the `Changed` bullet with the -683/-77 marked as measured on your corpora and the `Compatibility` line naming what moves for ELF consumers.

Your read on the conflict shape is right, and it is the useful generalisation: `Compatibility` merged cleanly because bullets there land in PR order, and only `Changed` collided. That is the part worth remembering next time — the collisions are per-subsection, not per-file, so two PRs writing to different subsections never conflict at all.

Merging. This closes #324.

PR #329, opened 2026-09-10
feat(common): let the exception directory refuse an interior candidate too

### LINKED ISSUE #328: The .pdata arm of the gap-only rules: measured at 47 false positives, and inert on dumps
Follow-up to #324, and the third of the three rules named in the sweep there. Filed with its measurement rather than as a to-do, because the number is small enough that it should decide whether this is worth doing at all.

**Implemented and measured in #329**, stacked on #327.

## The shape

Three rules refuse a candidate on evidence the image declares, and all three are reached only from the gap scan — they test `self.gap_pointer`. #327 widens the `.eh_frame` one to a candidate from any source. The other two are:

- **`USE_LSDA_LANDING_PADS`** — nothing left to reach. Of 302 false positives still standing on four AArch64 static cells after #327, none is a declared landing pad. A pad is interior by construction and the gap scan is what finds it, so the gap pointer is where its evidence is exhausted. Nothing to do here.
- **`USE_PE_X64_PDATA_INTERIOR_GAPS`** — this one, below.

## What widening the `.pdata` rule is worth

Measured over the full corpus, `feat/fde-interior-analysis-gate` (#327) → #329, so the figures isolate this arm:

| corpus | n | TP | ΔFP |
|---|---|---|---|
| Built C/C++ MinGW PE x64 | 120 | unchanged | **−47** |
| Built Rust, `windows-gnu-x64` | 2 | unchanged | **−2** |

Everything else is bit-identical — 140 x86-64 ELF cells, 72 AArch64 ELF, 23 Go, 11 ARM64 Mach-O, the ELF and 32-bit Rust cells, and all 57 malpedia dumps. Only PE x64 cells move in any corpus, and no corpus loses a true positive.

**Correction to the first version of this issue.** I opened it quoting *14 of 1,038* from a 20-cell sample and scaled that to roughly 80 across the 120 cells. Both numbers were wrong as a statement about the corpus: the sample was a real measurement of 20 cells, the extrapolation was not a measurement of anything. The corpus figure is 47. The table below is what the sample actually showed, kept because the per-class split is only measured there.

| 20 MinGW x64 PE cells, with #327 applied | count |
|---|---|
| remaining false positives | 1,038 |
| interior to a declared `RUNTIME_FUNCTION` extent | 59 |
| — primary record, owner **not** recovered | 40 |
| — primary record, owner recovered **and** its extent surrounds the address | 14 |
| — primary record, owner recovered but its extent does not surround | 5 |
| — fragment record | 0 |

The 45 that are interior but not refused are declined by the two conditions that keep the rule from costing recall: an owner the analysis never recovered says nothing about what covers the address, and an extent that does not surround it is the shape that cost three true positives on AArch64 in #327.

## The part worth knowing before deciding

**On memory dumps the rule is inert, and not because of any guard.** None of the 57 malpedia dumps declares an exception directory at all — `_pdata_ranges` is empty, so the rule never gets as far as its first condition. That is the corpus this project cares most about, and this evidence source does not reach it.

Between those two facts the case is thin: 47 false positives on one PE corpus, nothing on dumps. It is real, it is free of recall risk under the existing guards, and it is a small change on top of #327 — but it is not the 683 false positives the `.eh_frame` arm was worth, and it moves PE, which is exactly the control that makes #327 easy to read.

## Where it stands

#329 is open, green, and stacked on #327. Closing this and dropping #329 is a reasonable call on these numbers; the measurement is the argument either way.

r0ny123:
Closes #328. **Stacked on #327** — it extends the helper that PR introduces, so it must merge after it. The diff shown against `master` will include #327's commit until then; this PR's own commit is `2efa621`.

Third of the three rules named in the #324 sweep, and the last one with anything left to reach. #327 widened the `.eh_frame` arm to a candidate from any source; this does the same for the PE exception directory, under the guards that arm established.

## The rule

The same two conditions, unchanged: the `RUNTIME_FUNCTION` record's own function has to be recovered, and its *recovered* extent has to surround the address. The two evidence sources stay format-disjoint exactly as the gap scan's comment already notes — `_pdata_ranges` is only ever filled from a PE exception directory and the `.eh_frame` ranges decode nothing unless lief reports an ELF — so no address is ever arbitrated between them.

One deliberate difference from the gap scan. There, a **fragment record is evidence on its own**: it says the extent is part of some other function, and the scan needs nothing more. Here it is not taken as that. A fragment's own start is not the function covering the address, so it fails the recovered-owner test and declines rather than refusing. At a point where the better evidence — an actual recovered owner — is available, using the weaker form would be the wrong reading. No fragment record appeared among the interior false positives in the sample either way.

## Measured

`feat/fde-interior-analysis-gate` (#327) → this branch, so the figures isolate this arm.

| corpus | n | TP | ΔFP | bit-identical cells |
|---|---|---|---|---|
| Built C/C++ MinGW PE x64 | 120 | unchanged | **−47** | 109 / 120 |
| Built Rust, `windows-gnu-x64` | 2 | unchanged | **−2** | — |
| Built C/C++ ELF (gcc, clang) | 140 | unchanged | 0 | **140 / 140** |
| Built C/C++ AArch64 ELF | 72 | unchanged | 0 | **72 / 72** |
| Built Go | 23 | unchanged | 0 | **23 / 23** |
| ARM64 Mach-O | 11 | unchanged | 0 | **11 / 11** |
| Rust, ELF and 32-bit PE cells | 22 | unchanged | 0 | **22 / 22** |
| Malpedia dumps | 57 | unchanged | 0 | **57 / 57** |

**No corpus loses a true positive, and only PE x64 cells differ anywhere** — which is the control this arm has to pass, the mirror of #327's ELF-only one.

## The part worth reading before merging

**On memory dumps this evidence does not exist.** None of the 57 malpedia dumps declares an exception directory at all — `_pdata_ranges` is empty, so the rule never reaches its first condition. That is the corpus SMDA cares most about, and it is the honest limit of this change.

Between that and 47 false positives on one PE corpus, this is a small win, and #328 says so rather than selling it. It is here because it is free of recall risk under guards that are already measured, and because leaving one of three siblings unwidened without a reason is worse than either doing it or declining it on the record. If you would rather close #328 and keep the surface smaller, that is a reasonable call and the numbers are the argument either way.

## One thing I had wrong

I filed #328 saying 14 of 1,038 on a 20-cell sample, and scaled that to roughly 80 across 120 cells. The real figure is **47**. The sample over-predicted; the table above is the measurement.

I also assumed the Rust corpus was ELF, which is why the two moving cells looked like an anomaly at first. They are `windows-gnu-x64` — PE x64, exactly where this applies. The corpus is mixed, and every one of its ELF and 32-bit cells is bit-identical.

## Tests

Seven cases in `tests/testFdeInteriorGaps.py::DeclaredInteriorOwnerPeTest`, one condition each, exercising the arm through a stubbed exception directory so the ELF and PE arms are tested apart. Each was checked to fail when the guard it describes is removed: dropping the arm fails two, dropping its flag check fails one.

- `python -m pytest tests/` → **2,076 passed, 1 skipped, 2,593 subtests**
- `make typecheck` → exit 0, 0 error-level diagnostics, no new diagnostics
- Diff coverage against this PR's base → **100%** (12 lines, 0 missing)

No bundled fixture baseline moves.

## Fixed after review

**The shared rule answered to the wrong architecture's flag.** Both native backends fill `_pdata_ranges` — `intel` from an x64 exception directory, `aarch64` from an ARM64 PE's — and each gap scan gates on its own flag, but `declaredInteriorOwner` named only `USE_PE_X64_PDATA_INTERIOR_GAPS`. On an ARM64 PE that left `USE_PE_ARM64_PDATA_INTERIOR_GAPS` unable to turn the ARM64 rule off, and the x64 flag able to. Default output does not move — with both flags on the gap scan reaches those addresses first — but `testArm64PeInteriorGapFixture` fails, five of its eight pinned addresses stopping short of `off - on`.

Each backend now answers for itself through a `_pdataInteriorRefusalEnabled()` hook, mirroring the split the two gap scans already have. The base returns the disabled answer, which is what a manager carving no exception directory should give.

**The PE unit cases had to move with it.** `declaredOwnerOf` built a bare common manager, which now inherits that answer: two of the seven cases would fail outright and the other five would pass vacuously, taking their `None` from the gate rather than from the condition each is about. The helper takes a `manager_class` and the PE class builds the backend that fills the ranges. `DeclaredInteriorOwnerArm64PeTest` asks all seven of the same conditions of `aarch64`'s manager, each class asserting that the *other* architecture's flag does not gate it, and `DeclaredInteriorOwnerUnbackedPeTest` pins the inherited answer so a manager added without an override cannot start refusing on extents it never filled.

Four mutations verified, each caught by the intended case and nothing else: either backend's override answering to the other's flag, the shared gate naming a flag directly, and the base returning `True`.

**Also documented:** disjointness between the two structures is load-bearing here in a way it is not in the gap scan — a declared FDE range whose start was never recovered sets `owner` and then fails the recovered-owner test, so on an image carrying both the exception directory would never be asked about that address.

Full suite on the current tip: **2,136 passed, 1 skipped, 2,596 subtests**, `ruff` clean, `make typecheck` exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,18 @@ past roughly six lines it belongs in the PR the entry links.
 
 ### Changed
 
+- **(common)** Let a PE exception directory refuse an interior candidate too, not only an ELF's
+  `.eh_frame`. A `RUNTIME_FUNCTION` extent names the addresses inside a routine the way an FDE range
+  does, and it now answers where analysis would begin on a candidate from any source, under the same
+  guards: the record's own function has to be recovered, and its recovered extent has to surround the
+  address. A fragment record is not the shortcut it is in the gap scan -- its own start is not the
+  function covering the address, so it declines rather than refusing. *Measured on `fb510c4` against
+  compiler symbol tables:* 120 MinGW PE x64 cells 93.191 -> 93.235 PPV (-47 false positives, 109/120
+  bit-identical) and 2 Rust windows-gnu-x64 cells (-2), both at identical true positives and false
+  negatives; the 140 C/C++ ELF cells and all 57 malpedia dumps are bit-identical. Three of the dumps
+  do declare an exception directory, and the rule is consulted on all three and refuses nothing, which
+  is what makes that a control rather than an absence. (#329)
+
 - **(tests)** `make test` now runs the fast tier and `make test-all` runs the whole suite. The `slow`
   marker already existed, was already applied to the eleven fixture-corpus files, and was documented in
   `pyproject.toml` with its own deselect recipe — but no entry point used it, so every local run paid for
@@ -129,6 +141,10 @@ past roughly six lines it belongs in the PR the entry links.
 ### Security
 
 ### Compatibility
+
+- Recovery output moves on PE images that declare an exception directory: an address a
+  `RUNTIME_FUNCTION` extent covers is no longer reported as a function start unless it is that
+  record's own start. No true positive was lost on the corpora this was measured over. (#329)
 
 - Recovery output moves on Intel PE and ELF images whose gap scan meets a failed candidate: the
   bytes after it are now scanned instead of skipped. No true positive was lost on the corpora
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,13 @@ past roughly six lines it belongs in the PR the entry links.
   before the memo was written, so a section naming one dead pointer from every record read it once per record;
   remembered by address, that is now one read. *All 447 cells across the six corpora are bit-identical.* (#335)
 
+- **(common)** Let each backend decide whether its own exception-directory interior rule is enabled. Both native
+  backends fill the same `_pdata_ranges` list, but the shared rule named `USE_PE_X64_PDATA_INTERIOR_GAPS` alone,
+  so on an ARM64 PE the refusal answered to a switch for a format it never sees: the ARM64 flag no longer turned
+  it off, and the x64 flag would have. Default output does not move -- with both flags on the gap scan reaches
+  those addresses first -- so this is flag semantics rather than a recovery change, and the bundled ARM64 PE
+  fixture is what makes it visible. (#329)
+
 ### Security
 
 ### Compatibility
--- a/src/smda/SmdaConfig.py
+++ b/src/smda/SmdaConfig.py
@@ -147,8 +147,11 @@ class SmdaConfig:
     #   2 Rust windows-gnu-x64    -2 FP at identical TP and FN
     # Nothing else moves at all. The 140 x86-64 ELF cells, 72 AArch64 ELF cells, 23 Go cells,
     # 11 ARM64 Mach-O cells, the 32-bit and ELF Rust cells and all 57 malpedia dumps are
-    # bit-identical. The dumps are the reading worth keeping: none of the 57 declares an
-    # exception directory at all, so this evidence does not reach a mapped image.
+    # bit-identical. The dumps are the reading worth keeping, and not because the evidence is
+    # absent: three of the 57 do declare an exception directory, 282 to 430 records each, and on
+    # all three the rule is consulted and refuses nothing - every address its guards see is
+    # declined. Reachable and declining is the stronger control; an earlier note here claimed
+    # none of the dumps carried a directory, which is not so.
     USE_PE_X64_PDATA_INTERIOR_GAPS = True
     # Refuse a gap candidate that an ARM64 PE image's own exception directory places inside a
     # routine. The same evidence and the same rule as the x64 flag above, reached differently:
--- a/src/smda/SmdaConfig.py
+++ b/src/smda/SmdaConfig.py
@@ -137,6 +137,18 @@ class SmdaConfig:
     # an analysis that ran past the end of the routine the address sat inside and absorbed
     # the small aligned functions after it. On the cell examined, all eight functions
     # recovered sit 11 to 79 bytes past a declared extent's end.
+    # The same extents also answer for a candidate from any other source, at the point analysis
+    # would begin on it, under the guards the .eh_frame arm carries: the record's own function
+    # has to be recovered and its recovered extent has to surround the address. A fragment is
+    # not the shortcut it is in the gap scan - its start is not the function covering the
+    # address, so it declines here rather than refusing.
+    # Worth much less than the .eh_frame arm, and the numbers are why it is a change of its own:
+    #   120 MinGW PE x64 cells   -47 FP at identical TP and FN, 109 of 120 bit-identical
+    #   2 Rust windows-gnu-x64    -2 FP at identical TP and FN
+    # Nothing else moves at all. The 140 x86-64 ELF cells, 72 AArch64 ELF cells, 23 Go cells,
+    # 11 ARM64 Mach-O cells, the 32-bit and ELF Rust cells and all 57 malpedia dumps are
+    # bit-identical. The dumps are the reading worth keeping: none of the 57 declares an
+    # exception directory at all, so this evidence does not reach a mapped image.
     USE_PE_X64_PDATA_INTERIOR_GAPS = True
     # Refuse a gap candidate that an ARM64 PE image's own exception directory places inside a
     # routine. The same evidence and the same rule as the x64 flag above, reached differently:
--- a/src/smda/aarch64/FunctionCandidateManager.py
+++ b/src/smda/aarch64/FunctionCandidateManager.py
@@ -1032,6 +1032,10 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
             or auth_or_exception_return
         )
 
+    def _pdataInteriorRefusalEnabled(self):
+        """The flag this backend's gap scan already answers to, for the analysis-time rule."""
+        return self.config.USE_PE_ARM64_PDATA_INTERIOR_GAPS
+
     def nextGapCandidate(self, start_gap_pointer=None):
         # AArch64 gap scan: a fixed-stride linear sweep of unanalyzed executable bytes
         # for functions that no prologue, call reference or stored pointer reached
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -733,8 +733,23 @@ class FunctionCandidateManager:
             index -= 1
         return None
 
+    def _pdataInteriorRefusalEnabled(self):
+        """Whether this backend's exception-directory interior rule is enabled.
+
+        `_pdata_ranges` is filled by whichever backend read the directory, and the two that do
+        answer to separate flags: the x64 and ARM64 PE rules are separate features with separate
+        defaults. Naming either flag here would leave one architecture's rule answering to the
+        other architecture's switch, so each backend says instead. Those two are also the only
+        managers this engine builds today, so what is inherited here is the answer for one that
+        carves no directory at all: off, there being no extents for the rule to read.
+        """
+        return False
+
     def declaredInteriorOwner(self, addr):
-        """The recovered function whose declared `.eh_frame` range contains `addr`, or None.
+        """The recovered function whose declared range contains `addr`, or None.
+
+        Either structure that declares one answers: an ELF's `.eh_frame` FDE ranges, or a PE
+        exception directory's `RUNTIME_FUNCTION` extents.
 
         The same evidence the gap scan already refuses on, asked at the point a candidate from
         any source is about to be analysed rather than only where the gap pointer reaches. Both
@@ -747,9 +762,13 @@ class FunctionCandidateManager:
         Both structures answer here, as they do in the gap scan, and they are format-disjoint:
         `_pdata_ranges` is only ever filled from a PE exception directory and the `.eh_frame`
         ranges decode nothing unless lief reports an ELF, so no address is arbitrated between
-        them. A fragment record is not the shortcut it is in the gap scan: its own start is not
-        the function that covers the address, so it fails the recovered-owner test below and
-        declines, which is the conservative reading at a point where a better one is available.
+        them. Disjointness is load-bearing rather than merely tidy, unlike in the gap scan: a
+        declared FDE range whose start was never recovered sets `owner` and then fails the
+        recovered-owner test, so on an image carrying both structures the exception directory
+        would never be asked about that address. A fragment record is not the shortcut it is in
+        the gap scan: its own start is not the function that covers the address, so it fails the
+        recovered-owner test below and declines, which is the conservative reading at a point
+        where a better one is available.
 
         A third guard the gap scan does not need: the owner's own recovered extent has to
         surround the address. An FDE can reach past everything its function's control flow
@@ -770,7 +789,7 @@ class FunctionCandidateManager:
             declared = self.declaredFdeRangeContaining(addr)
             if declared is not None:
                 owner = declared[0]
-        if owner is None and self._pdata_ranges and self.config.USE_PE_X64_PDATA_INTERIOR_GAPS:
+        if owner is None and self._pdata_ranges and self._pdataInteriorRefusalEnabled():
             declared = self.declaredExceptionRangeContaining(addr)
             if declared is not None:
                 owner = declared[0]
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -744,6 +744,13 @@ class FunctionCandidateManager:
         begin in the alignment padding ahead of its function, which leaves the real entry a few
         bytes in interior to nothing.
 
+        Both structures answer here, as they do in the gap scan, and they are format-disjoint:
+        `_pdata_ranges` is only ever filled from a PE exception directory and the `.eh_frame`
+        ranges decode nothing unless lief reports an ELF, so no address is arbitrated between
+        them. A fragment record is not the shortcut it is in the gap scan: its own start is not
+        the function that covers the address, so it fails the recovered-owner test below and
+        declines, which is the conservative reading at a point where a better one is available.
+
         A third guard the gap scan does not need: the owner's own recovered extent has to
         surround the address. An FDE can reach past everything its function's control flow
         arrives at, and refusing an address out there discards bytes nothing else claims --
@@ -758,17 +765,21 @@ class FunctionCandidateManager:
         nothing in the format forbids that; what bounds it is measurement rather than
         structure, no true positive lost across the ELF corpora this was measured over.
         """
-        if not self.config.USE_ELF_FDE_INTERIOR_GAPS:
+        owner = None
+        if self.config.USE_ELF_FDE_INTERIOR_GAPS and not self.isInDeclaredPltSection(addr):
+            declared = self.declaredFdeRangeContaining(addr)
+            if declared is not None:
+                owner = declared[0]
+        if owner is None and self._pdata_ranges and self.config.USE_PE_X64_PDATA_INTERIOR_GAPS:
+            declared = self.declaredExceptionRangeContaining(addr)
+            if declared is not None:
+                owner = declared[0]
+        if owner is None or owner not in self.disassembly.functions:
             return None
-        if self.isInDeclaredPltSection(addr):
-            return None
-        containing = self.declaredFdeRangeContaining(addr)
-        if containing is None or containing[0] not in self.disassembly.functions:
-            return None
-        borders = self.disassembly.function_borders.get(containing[0])
+        borders = self.disassembly.function_borders.get(owner)
         if borders is None or not borders[0] <= addr < borders[1]:
             return None
-        return containing[0]
+        return owner
 
     def opensInsideDeclaredFdeRange(self, addr):
         """Whether `addr` falls inside a declared FDE range without being that range's start."""
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -239,6 +239,10 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
             return None
         return target
 
+    def _pdataInteriorRefusalEnabled(self):
+        """The flag this backend's gap scan already answers to, for the analysis-time rule."""
+        return self.config.USE_PE_X64_PDATA_INTERIOR_GAPS
+
     def nextGapCandidate(self, start_gap_pointer=None):
         if self.language_candidates_only:
             return None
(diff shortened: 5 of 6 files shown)


danielplohmann:
**Blocking, and it is an interaction with #333 rather than anything wrong in the measurement.** #333 merged an hour ago as `50a818a`, so master now carries a real ARM64 PE fixture — and on this branch its test fails:

```
tests/testArm64PeInteriorGapFixture.py::testTheDeclaredExtentsRefuseTheirInteriorAddresses
AssertionError: Items in the second set but not the first:
  0x1400014E8  0x1400015F8  0x140001614  0x140001628  0x140001674
```

Five of the eight pinned addresses stop appearing in `off - on`, because with `USE_PE_ARM64_PDATA_INTERIOR_GAPS` off they are now refused anyway — by this PR's arm, which is gated on the **x64** flag.

## The cause

`declaredInteriorOwner` reads `self._pdata_ranges` and gates it on `USE_PE_X64_PDATA_INTERIOR_GAPS`. But that list is filled by whichever backend carved it, and each backend's gap scan gates on its own flag:

| | fills `_pdata_ranges` | gap scan gates on |
|---|---|---|
| `intel/FunctionCandidateManager.py` | `:878`, x64 exception directory | `:332` → `USE_PE_X64_PDATA_INTERIOR_GAPS` |
| `aarch64/FunctionCandidateManager.py` | `:287`, ARM64 PE `.pdata` | `:1102` → `USE_PE_ARM64_PDATA_INTERIOR_GAPS` |

The shared helper has one flag name for both, so on an ARM64 PE the analysis-time refusal answers to the x64 flag. Measured on `arm64_pe_probe_xored`, instrumenting `declaredInteriorOwner`:

| config | analysis-time refusals | functions recovered |
|---|---|---|
| both flags on (default) | 0 | 105 |
| `USE_PE_ARM64_PDATA_INTERIOR_GAPS` off | **2** (`0x1400014E8`, `0x140001614`) | 108, where master gives 113 |
| `USE_PE_X64_PDATA_INTERIOR_GAPS` off | 0 | 105 |

So the ARM64 flag no longer turns the ARM64 rule off, and the x64 flag would turn it off on an image it has nothing to do with.

**Default output on that fixture does not move** — with both flags on the gap scan reaches all eight first, which is why 0 refusals — so this is flag semantics plus a broken test, not a behaviour regression. It is also why your corpus table could not see it: there is no ARM64 PE row in it, and at defaults there is nothing to see.

## The fix I would take

A backend hook rather than a flag name in shared code, mirroring the split the two gap scans already have:

```python
# common/FunctionCandidateManager.py
def _pdataInteriorRefusalEnabled(self):
    """Whether this backend's exception-directory interior rule is on. See the two overrides."""
    return False
```

with `intel` returning `USE_PE_X64_PDATA_INTERIOR_GAPS` and `aarch64` returning `USE_PE_ARM64_PDATA_INTERIOR_GAPS`; `cil` and `dalvik` inherit the `False` and lose nothing, since neither fills `_pdata_ranges`. Then `declaredInteriorOwner` asks the hook.

I applied that here to check it: `testArm64PeInteriorGapFixture` goes green and nothing else moves.

**It does cost your PE unit cases, and in a way worth knowing before you write it.** `declaredOwnerOf` builds a bare common `FunctionCandidateManager`, which inherits the `False`, so two of the seven `DeclaredInteriorOwnerPeTest` cases fail outright — `testAnAddressInsideADeclaredExtentIsRefused` and `testTheElfFlagDoesNotGateTheExceptionDirectory` — and the other five keep passing **vacuously**: they assert `None` and now get it from the gate rather than from the condition each one is about. So the helper has to build an `intel` manager for the PE arm, or the PE half of that class stops testing anything. The ELF cases are unaffected.

Worth a case pinning the architecture split as well: the ARM64 flag off, over the fixture #333 bundles, should produce #333's eight and nothing else — which is exactly the assertion that just failed.

## The rest of the review, which stands

The fragment decision is right and it is the interesting difference from the gap scan: a fragment's own start is not the function covering the address, so taking the gap scan's shortcut here would be using the weaker evidence at the point the stronger one is available. Declining rather than refusing is the conservative direction.

The format-disjointness argument holds as written — but note the code makes it load-bearing in a way the gap scan does not: an `.eh_frame` range whose start was not recovered sets `owner` and then fails the recovered-owner test, so the `.pdata` arm is never consulted for that address. Fine while no image carries both, which is what the comment says; worth knowing it is now a structural assumption rather than an ordering one.

Rebase over `a043145` and re-run — #326, #330, #331, #332 and #333 are all in master now.

r0ny123:
Reproduced and fixed, and your diagnosis is exactly right — `testArm64PeInteriorGapFixture` fails here on this branch with the same five addresses, and the cause is the shared helper naming one architecture's flag for a list both backends fill.

Took your design. `_pdataInteriorRefusalEnabled()` on the common manager returning the disabled answer, `intel` returning `USE_PE_X64_PDATA_INTERIOR_GAPS` and `aarch64` returning `USE_PE_ARM64_PDATA_INTERIOR_GAPS`, and `declaredInteriorOwner` asking the hook. The fixture test goes green and nothing else moves.

## One correction to the note, and it was in my commit message too

`cil` and `dalvik` do not inherit the `False` — they never reach it. Neither implements `createCandidateManager`; both have standalone disassemblers that never construct a candidate manager or touch `RecursiveDisassembler` at all. The two native backends are the only managers this engine builds.

That does not change the design — the base still has to answer, because the bare common class is what a unit helper builds and what any manager added later inherits — but the reason is "this is the answer for a manager that carves no directory", not "these two backends rely on it". Stated that way in the docstring, the test docstring and the commit body now.

## The unit cases

Exactly as you measured: left alone, `testAnAddressInsideADeclaredExtentIsRefused` and `testTheElfFlagDoesNotGateTheExceptionDirectory` fail outright and the other five pass vacuously. `declaredOwnerOf` takes a `manager_class` and the PE class builds `intel`'s.

`DeclaredInteriorOwnerArm64PeTest` subclasses it with `aarch64`'s, so all seven conditions are asked of both backends, and each class asserts that **the other** architecture's flag does not gate it. `DeclaredInteriorOwnerUnbackedPeTest` pins the inherited answer directly, so a manager added without an override cannot start refusing on extents it never filled.

Verified by mutation — each of these fails the intended case and nothing else:

| mutation | caught by |
|---|---|
| `aarch64` override returns the x64 flag (the shipped bug) | `DeclaredInteriorOwnerArm64PeTest::testTheOtherArchitecturesFlagDoesNotGateThisOne`, `testArm64PeInteriorGapFixture` |
| shared gate names the x64 flag directly | those two plus `DeclaredInteriorOwnerArm64PeTest::testTheFlagTurnsTheRefusalOff` and the unbacked case |
| `intel` override returns the ARM64 flag | `DeclaredInteriorOwnerPeTest::testTheFlagTurnsTheRefusalOff`, `testTheOtherArchitecturesFlagDoesNotGateThisOne` |
| base returns `True` | `DeclaredInteriorOwnerUnbackedPeTest` |

## The format-disjointness note

Written into the docstring, since you are right that it is now structural rather than incidental:

> Disjointness is load-bearing rather than merely tidy, unlike in the gap scan: a declared FDE range whose start was never recovered sets `owner` and then fails the recovered-owner test, so on an image carrying both structures the exception directory would never be asked about that address.

Left as an assumption rather than restructured — an arm that re-tries on a *qualifying* owner rather than any owner would be code for a case no image produces.

## Sweep

For the class behind this — shared code gating on one backend's flag where a sibling reaches the same path — the whole tree has one other hit, `_splitMergedFunctionsAtPdataEnds`, which is guarded by `backend.name != "intel"` and has no ARM64 counterpart flag. Not a sibling. `USE_ELF_FDE_INTERIOR_GAPS`, `USE_ELF_EH_FRAME_CANDIDATES` and `USE_MACHO_FUNCTION_STARTS` are format-scoped rather than architecture-scoped and are correctly named in shared code.

## Verification

Full suite on this tip: **2,136 passed, 1 skipped, 2,596 subtests**. `ruff` clean, `make typecheck` exit 0 at the same diagnostic count this stack already had.

Master is merged in rather than rebased on both this and #327 — I would rather not force-push over branches you have already reviewed, so taking the rebase at merge is yours. #327 carries the docstring correction from your review there, and this still stacks on it.

r0ny123:
Rebased over `a043145` as you asked, replacing the two merge commits — five linear commits now, the first three being #327's, and the tree is unchanged by it (`git diff` against the pre-rebase tip is empty), so the suite, lint and typecheck results reported above still stand as measured.

```
7cad9ff fix(common): let each backend say whether its own pdata rule is on
615bed6 feat(common): let the exception directory refuse an interior candidate too
a4c1513 test(tests): keep a failed claim from leaking onto the class      <- #327
12ec24a docs(common): state what the interior refusal can actually reach  <- #327
c4e3ca1 feat(common): refuse a candidate the image declares interior      <- #327
```

Once #327 merges this reduces to its own two commits.

danielplohmann:
The backend hook is the fix, and `7cad9ff` is the shape I checked locally before recommending it — `_pdataInteriorRefusalEnabled()` asked of the manager that filled `_pdata_ranges`, mirroring the split the two gap scans already carry, with the common class answering no because a manager that carves no directory has no extents to refuse on.

You went further than the fix in the one place it mattered. Making the PE cases build the real backend rather than a bare common manager is what stops five of the seven passing vacuously — they were getting their `None` from the gate rather than from the condition each case is about — and `DeclaredInteriorOwnerArm64PeTest` subclassing the Intel cases pins the split from both sides in the cheapest way available. `DeclaredInteriorOwnerUnbackedPeTest` is the one I would not have asked for and is the one that will still be right in a year: it pins what a backend added later inherits, so a new manager cannot quietly start refusing on extents it never filled.

`testArm64PeInteriorGapFixture` goes green, which was the blocker — five of #333's eight pinned addresses were dropping out of `off - on` because the ARM64 rule answered to the x64 switch.

The widened `declaredInteriorOwner` summary is worth having; it had still named only `.eh_frame` after the rule learned to read an exception directory. The added paragraph on disjointness being load-bearing here rather than merely tidy — a declared FDE range whose start was never recovered sets `owner` and then fails the recovered-owner test, so the directory never gets asked — is the kind of thing that is invisible until someone builds an image carrying both.

Verified stacked over `a043145` with #336, #337, #327 and #335: 2,142 passed, 2 skipped, 2,596 subtests; `ruff` clean; `ty` exit 0.

No further review notes. One thing before it merges: with #336 and #337 in, this needs its own `## [Unreleased]` entry — `Changed` or `Fixed`, with the -47 marked as measured on your corpora and not reproduced here, per the provenance rule from #323.

danielplohmann:
Two things: a merge of master is needed, and there is an entry missing.

## The merge is mine, not yours

You based this on `0a0a4c6`; `8a4d994` (#340) and `130fa02` (#335) have landed since, into the same `Unreleased` subsections. `CHANGELOG.md` is the only conflicting file and the resolution is purely additive — keep both sides. Nothing in `src/` collides any more, since your merge of `0a0a4c6` already took #339's `_failedGapResumeTarget` alongside your `_pdataInteriorRefusalEnabled` in the intel manager.

Resolved and verified locally: **2155 passed, 2 skipped, 2596 subtests**, `ruff` clean, `ty` at 274 against master's 270 — the same four `None | Unknown` warnings on this stack's own new lines. I would have pushed it to your branch but I will not push to a contributor's fork without asking, so it is yours to land.

## The entry for this PR's actual change is missing

The `Unreleased` section carries exactly one `(#329)` entry, and it is the backend-hook fix — the flag-semantics correction from `7cad9ff`, which as you correctly say does not move default output. What has no entry is `615bed6`, the change this PR is named after: the exception directory refusing an interior candidate from any source. That is the one #328 measured at **-47 false positives across 120 MinGW PE cells and -2 on the 2 windows-gnu-x64 Rust cells**, and it does move recovery output on PE x64.

So it needs its own bullet under `### Changed`, and a `### Compatibility` line in the same shape as the one #327 carries — "recovery output moves on PE x64 images carrying an exception directory". Both #338 and #327 have theirs; this one is the gap.

Worth saying plainly because it is the exact failure the format was adopted to prevent, and it has appeared on the first engine PR after it landed: at release time a missing entry is indistinguishable from a change that needed none, and only you can tell the difference while the PR is open. The `Changelog entry` check passed here because the check is file-level — this PR does touch `CHANGELOG.md` — which is the acceptable false negative you and I both signed off on in #336. It is doing what it was designed to do; it just cannot catch this shape.

The figures are already written in #328, so this is transcription rather than measurement. Once it is in and master is merged, this is ready and v4.7.0 is next.

danielplohmann:
#327 is merged as `fb510c4`, so this needs a rebase — and it is the squash-merge kind rather than another changelog collision.

Your branch still carries #327's three commits as their original objects (`8933750`, `51c92cd`, `b2af2df`) plus its two docs commits. Master now has the same content under one squashed commit, so git sees both sides adding the same lines: `src/smda/common/FunctionCandidateManager.py` and `tests/testFdeInteriorGaps.py` both conflict. Rebasing onto `fb510c4` and dropping the five commits that are now upstream leaves this PR as what it actually is — `80fc3bf` (the exception-directory arm), `f049a63` (the backend hook) and `48ae03a` (its entry).

Two things then, and the second is the one that matters:

1. **The rebase.** Mechanical, and the conflicts disappear once the duplicated commits are gone rather than needing resolution.

2. **The missing entry, still open from my note above.** `## [Unreleased]` carries exactly one `(#329)` bullet and it is `48ae03a`, for the backend hook — the flag-semantics fix that, as you correctly wrote, does not move default output. `80fc3bf` has nothing: no `Changed` bullet and no `Compatibility` line, and that is the change this PR is named after and the one #328 measured at **-47 false positives across 120 MinGW PE cells and -2 on the 2 windows-gnu-x64 Rust cells**.

The shape to match is what #327 just landed with: a `Changed` bullet carrying the mechanism and the figures marked as measured on your corpora and not reproduced here, and a `Compatibility` line saying recovery output moves on PE x64 images carrying an exception directory. The numbers are already written in #328, so this is transcription.

Everything else is verified and stands — I ran the merged tree at `1fb296b5` before #327 landed: 2155 passed, 2 skipped, 2596 subtests, `ruff` clean, `ty` +4 on this stack's own lines. Nothing in the review is outstanding.

This is the last thing before v4.7.0. It closes #328 when it lands.

r0ny123:
Both done, and the missing entry turned up something in the config comment.

## The merge

Merged rather than rebased — the duplicate-add conflicts resolve without rewriting history, since my side of each block is `#327`'s content *as this PR modifies it* and master's is `#327` alone, so taking mine is the union rather than a choice. Verified structurally rather than by eye: the working tree against master is exactly this PR's own six files, `RecursiveDisassembler` correctly absent because that arm is yours now, and no method or test class appears twice.

That check earned its keep. My earlier "keep both sides" resolution had silently **duplicated `#327`'s `Compatibility` bullet** — master's copy and mine, both kept. Removed; every bullet in `## [Unreleased]` is now unique, and the section's diff against master is exactly this PR's three.

## The missing entry

You are right that `80fc3bf` had nothing. Written now as a `Changed` bullet and a `Compatibility` line, and **re-measured on `fb510c4` rather than carried forward** — both #327 and #339 have landed under this since those figures were taken, and #339 moves MinGW PE recovery, so the −47 could easily have shifted.

It did not. Every figure reproduces exactly:

| corpus | n | before → after | delta | TP lost |
|---|---|---|---|---|
| MinGW PE x64 | 120 | 93.191 → 93.235 PPV | **−47 FP** | **0** |
| Rust windows-gnu-x64 | 2 | — | **−2 FP** | **0** |
| built C/C++ ELF | 140 | — | — | 140/140 bit-identical |
| malpedia dumps | 57 | — | — | 57/57 bit-identical |

## One correction, and it is to a line you quoted approvingly

The config comment said *"none of the 57 declares an exception directory at all, so this evidence does not reach a mapped image."* **Three of them do** — `x64-prikorma`, `x64-trickbot` and `x64-vawtrak`, carrying 282, 430 and 370 records.

The conclusion survives; the reason was wrong, and the truth is the better control. Instrumented on all three, the rule **is** consulted and refuses **nothing** — every address its guards see is declined. Reachable-and-declining says the guards hold on mapped dumps; an absence says only that the corpus never asked. The comment now says that instead.

Suite on the merged tree **2,156 passed, 1 skipped, 2,596 subtests**; `ruff` clean; `ty` at 267, this box's master baseline of 263 plus the same four.

r0ny123:
Rebased onto `fb510c4` as you asked. The five upstream commits are gone and this PR is now the four it actually is:

```
024f4bc docs(docs): add the missing entry for the exception-directory arm
2dd4ab5 docs(docs): add the Unreleased entry for the backend pdata hook
9f2115c fix(common): let each backend say whether its own pdata rule is on
fa32f7f feat(common): let the exception directory refuse an interior candidate too
```

Zero merge commits, and the tree is unchanged by the rebase — `git diff` against the pre-rebase tip is empty — so the suite, lint and typecheck figures from the merged tree stand as measured rather than needing a re-run. Re-ran them anyway since the history moved: **2,156 passed, 1 skipped, 2,596 subtests**, `ruff` clean, `ty` at 267 against this box's master baseline of 263.

Two conflicts came up in the replay rather than disappearing, both additive and both because the commits predate what landed under them: `f049a63` puts `_pdataInteriorRefusalEnabled` where master now has #339's `_failedGapResumeTarget`, and `48ae03a` adds its bullet to a `### Fixed` that has since gained #338's and #335's. Both methods kept, bullet appended after the existing ones.

The diff against master is the six files this PR owns and nothing else.

danielplohmann:
Merged as `eae880a`.

The backend hook is the right shape and it clears what I blocked on: `testArm64PeInteriorGapFixture` goes green again, and each backend now answers for its own flag rather than both answering to `USE_PE_X64_PDATA_INTERIOR_GAPS`. The `DeclaredInteriorOwnerUnbackedPeTest` case is the part I would not have asked for and is worth having - it pins the inherited answer, so a manager added later without an override cannot start refusing on extents it never filled. Same for taking the five vacuous passes seriously rather than leaving them green.

Verified on saturn against master at `2c74bd7`: 2031 passed, 2 skipped, 1663 subtests on the fast tier, and the full matrix green on your side.

Two notes, neither blocking:

- The dump result is the honest part of this PR and the reason #328 was worth filing either way. None of the 57 malpedia dumps declares an exception directory, so on the corpus SMDA cares most about this arm is inert by construction. The changelog entry says so, which is what the provenance rule was for.
- Declining on a fragment record rather than refusing is the right call. A fragment's start is not the covering function, and taking the weaker evidence where the stronger one is available would have been a different rule than the one measured.

This closes #328 and empties the v4.7.0 milestone. Cutting the release next, by hand - #343 is not in this milestone's scope, so it is not gating it. I will come to #343 after the tag is out.

PR #339, opened 2026-09-11
fix(intel): resume the gap scan inside the gap a failed candidate was in

### LINKED ISSUE #338: A failed gap candidate abandons the rest of the gap it was found in (52 of 58 missed function starts on the #247 binary)
Found while verifying #247 on `857081f`. That issue's original defect is fixed; this is a different one that its reproducer happens to expose cleanly.

## The defect

A gap candidate that fails to become a function causes the scan to abandon the rest of the gap it was found in.

`RecursiveDisassembler.py:490` resumes the scan with `getNextGap(dont_skip=True)`. That method's refinement — the one whose comment says *"we potentially just disassembled a function and want to continue directly behind it in case we would otherwise miss more"* — is gated on `self.gap_pointer in self.disassembly.code_map`:

```python
if (dont_skip and self.gap_pointer is not None
        and self.disassembly is not None
        and self.gap_pointer in self.disassembly.code_map):
    function = self.disassembly.ins2fn[self.gap_pointer]
    next_gap = min(next_gap, self.disassembly.function_borders[function][1])
```

A candidate that **became** a function is in `code_map`, so the scan correctly continues just past it. A candidate that **failed** is not, so the refinement never fires and `next_gap` keeps the value computed above it — the start of the next entry in `self.function_gaps`.

`function_gaps` is the problem. It is built by `updateFunctionGaps()`, which is called from exactly one place, `initGapSearch()`, and only when `gap_pointer is None`. It is a snapshot of the gaps as they stood when the gap phase opened, and is never refreshed. So "the next gap" means the next *original* gap, and everything between the failed candidate and the end of the gap it was sitting in is skipped.

## Measured on the #247 bundle

`hyperfine.exe`, i686 PE, against master `857081f`:

- **4328 gap candidates, of which 224 fail to become functions.**
- The PDB describes 2296 procedure entries; 58 are not recovered.
- **52 of those 58 are never yielded as a candidate at all** — the scan jumps over them.
- Of those 52, **52 are preceded by a failed candidate inside the same precomputed gap.** Unanimous, not a tendency.

A worked example:

```
precomputed gap          [0x40eb45, 0x40ee40)
candidate 0x40ed18       analysed, did NOT become a function
next candidate yielded   0x410543   <- start of the next precomputed gap
skipped                  0x40ed40, a 213-byte function the PDB names
                         hyperfine::error::impl$4::fmt
```

The 58 missed entries are ordinary code, not edge cases: all in `.text`, all nonzero size, none overlapping a recovered function, none in the code map. 29 open directly after an `int3` run and the rest after a `ret` plus padding. All 58 are 16-byte aligned. They are dominated by Rust `core::fmt` vtable shims — reachable only through a function pointer, so the gap scan is the only thing that would ever find them.

## Two prototypes, and why neither is a PR yet

Both add a `_failedGapResumeTarget()` hook on the common manager returning `None` (historical behaviour), overridden in the Intel backend. Patch attached below.

**(a) Resume one byte past the failed candidate.** Correct in principle, unshippable in practice:

| | master | resume+1 |
|---|---|---|
| runtime | 2.6s | **31.2s** |
| unmatched procedure records | 58 | 10 |
| functions not in PDB truth | 4212 | 5923 |

An order of magnitude in runtime, because the scan then walks the interior of everything that failed, byte by byte.

**(b) Resume past the next `int3` run inside the gap.** The shape the missed entries actually have, and it steps over the failed body in one move:

| | master | resume-at-padding |
|---|---|---|
| runtime | 2.6s | **2.8s** |
| unmatched procedure records | 58 | **10** |
| functions gained | — | 270 |
| of those, in PDB truth | — | **48** |
| of those, not in truth | — | 222 |
| lost | — | **0** |

48 of the 58 recovered, nothing regressed, 8% runtime. That is the promising one.

## Why I am filing this rather than opening the PR

**The 222 are unverified, and at least some are wrong.** On `dotnet_readytorun_pe_xored` variant (b) adds 117 functions. The first few are textbook — `push rbp; sub rsp, 0x20` after `int3` padding. But 75 of the 117 start with `and` or `cmp` immediately after a *lone* `0xCC` byte sitting inside what reads as a data structure, not alignment padding. That is the resume landing mid-structure and booking the remainder as a function.

**I predicted 16-byte alignment would separate the two groups, and it did not.** All 58 real missed entries are 16-aligned, as are 2278 of 2296 PDB procedures, so rounding the resume target up to the next 16-byte boundary looked like the obvious filter. Measured, it is a wash:

| fixture | master | (b) padding | (b) + 16-byte alignment |
|---|---|---|---|
| `dotnet_readytorun_pe` | 615 | 732 | **734** |
| `cutwail` | 504 | 511 | 505 |
| `asprox_0x008D0000` | 107 | 108 | 107 |
| `mirai_i386` | 166 | 169 | 167 |
| `rust_pe_gnu` | 3317 | 3343 | 3341 |
| `hyperfine.exe` (procs missed) | 58 | 10 | 10 |

It moves *which* candidates appear rather than removing bad ones — on the fixture with the worst-looking additions it adds two more. Recording it so nobody else spends the hour.

## Blast radius

Bundled fixtures, (b) against master: **25 of 34 bit-identical, 9 move, and all 9 are Intel.** Every AArch64, Mach-O, Dalvik and CIL fixture is untouched, which is the control an Intel-only hook should produce. Of the 9, four move only in `num_failed_functions` with an unchanged function set.

Full suite on the patch: **2110 passed, 2 skipped, 2596 subtests** — identical to master's own count on this box. `ruff` clean.

*(A methodology note, since it bit me: comparing report identity hashes across two source trees shows spurious differences, because an unsupported-architecture report embeds a traceback containing the source path. Exclude `message` as well as `timestamp` and `execution_time`, or the unsupported-arch fixtures all read as changed.)*

## What it needs

A precision measurement against symbol-based truth on the built corpora, which I do not have here — one binary and a fixture set cannot tell 48 true positives plus 222 unknowns from 48 plus 222 false positives. @r0ny123, this is squarely in the territory your #324 and #328 measurements cover; if the corpora say the added population is mostly real, (b) is a small change with a real recall win, and if they say it is mostly noise then the honest fix is narrower — resume only when the gap's remainder has a padding-delimited, aligned entry that decodes cleanly.

Either way the abandonment itself is a bug independent of what the fix should be: 224 failed candidates on one binary each silently drop the rest of their gap.

<details>
<summary>Prototype patch — this is variant (b) <em>with</em> the alignment rounding; drop the final <code>_ENTRY_ALIGNMENT</code> round-up to get plain (b)</summary>

```diff
diff --git a/src/smda/common/FunctionCandidateManager.py b/src/smda/common/FunctionCandidateManager.py
index cd16e96..65991a2 100644
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -288,9 +288,33 @@ class FunctionCandidateManager:
                 dont_skip,
                 next_gap,
             )
+        elif dont_skip and self.gap_pointer is not None and self.disassembly is not None:
+            # The branch above only fires for a candidate that became a function. One that did
+            # not leaves gap_pointer outside code_map, so the scan resumes at the next entry of
+            # the gap map -- which was snapshotted once in initGapSearch() and never refreshed --
+            # abandoning whatever is left of the gap this candidate was found in. Ask the backend
+            # where the next plausible entry inside this gap is instead.
+            resume = self._failedGapResumeTarget()
+            if resume is not None:
+                next_gap = min(next_gap, resume)
+                LOGGER.debug(
+                    "getNextGap(%s) => resuming inside the gap a failed candidate was in: 0x%08x",
+                    dont_skip,
+                    next_gap,
+                )
         LOGGER.debug("getNextGap(%s) final gap_ptr: 0x%08x", dont_skip, next_gap)
         return next_gap
 
+    def _failedGapResumeTarget(self):
+        """Where to resume after a gap candidate failed to become a function, or None.
+
+        None keeps the historical behaviour: abandon the rest of the gap. A backend overrides
+        this when it can name the next plausible entry cheaply -- scanning byte by byte from the
+        failed address instead costs an order of magnitude in runtime for candidates that are
+        mostly the interior of whatever the failed one was.
+        """
+        return None
+
     def nextGapCandidate(self, start_gap_pointer=None):
         """Architecture-specific gap scan; implemented per backend."""
         raise NotImplementedError
diff --git a/src/smda/intel/FunctionCandidateManager.py b/src/smda/intel/FunctionCandidateManager.py
index 5ec97e6..cd36bb5 100644
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -105,6 +105,14 @@ _PDATA_MAX_FUNCTION_SIZE = 0x100000
 _UNWIND_INFO_FIRST_BYTES = frozenset(0x01 | (flags << 3) for flags in range(8))
 
 
+_INT3 = b"\xcc"
+#: How far past a failed gap candidate to look for the padding run that ends it. A gap
+#: larger than this keeps the historical behaviour rather than guessing.
+_FAILED_GAP_RESUME_WINDOW = 4096
+#: Entry alignment a resumed candidate has to sit on.
+_ENTRY_ALIGNMENT = 16
+
+
 class FunctionCandidateManager(_CommonFunctionCandidateManager):
     CANDIDATE_CLASS = FunctionCandidate
 
@@ -193,6 +201,35 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
             is_alignment_sequence = False
         return is_alignment_sequence
 
+    def _failedGapResumeTarget(self):
+        """The first address past the next int3 run inside the gap the failed candidate was in.
+
+        Every entry this recovers on the measured binary opens after alignment padding -- an
+        int3 run, or a `ret` followed by one -- which is what a compiler leaves between two
+        functions it emitted adjacently. Resuming there steps over the interior of whatever the
+        failed candidate was in one move, where a byte-by-byte resume walks all of it.
+        """
+        if self.gap_pointer is None or self.disassembly is None:
+            return None
+        start = self.gap_pointer + 1 - self.disassembly.binary_info.base_addr
+        window = self.disassembly.getRawBytes(start, _FAILED_GAP_RESUME_WINDOW) or b""
+        run_start = window.find(_INT3)
+        if run_start < 0:
+            return None
+        cursor = run_start
+        while cursor < len(window) and window[cursor] == _INT3[0]:
+            cursor += 1
+        if cursor >= len(window):
+            # the run reaches the end of the window; nothing here says where it stops
+            return None
+        target = self.gap_pointer + 1 + cursor
+        # Round up to the entry alignment. A lone 0xCC inside a data structure is not padding,
+        # and resuming on the byte after it books the rest of that structure as a function --
+        # 75 of the 117 the unaligned form adds on the ReadyToRun fixture are that shape.
+        # Rounding up cannot step over an aligned entry, since the next aligned address at or
+        # after the target is exactly what it returns.
+        return (target + _ENTRY_ALIGNMENT - 1) & ~(_ENTRY_ALIGNMENT - 1)
+
     def nextGapCandidate(self, start_gap_pointer=None):
         if self.language_candidates_only:
             return None
```
</details>

r0ny123:
Closes #338.

## The answer to the question you asked

**The added population is mostly noise — 36% real — so the fix is the narrower one.** But the narrowing that works is not the one you tried: alignment as a **filter** rather than as a round-up.

Variant (b) on the 260-cell built C/C++ matrix, against compiler symbol tables:

| variant | true positives | false positives | ratio | micro PPV | bit-identical |
|---|---|---|---|---|---|
| master | — | — | — | 96.266 | — |
| (b) resume at padding | **+52** | **+93** | 0.56 : 1 | 96.225 | 248/260 |
| (b) + 16-byte round-up | +61 | +108 | 0.56 : 1 | 96.218 | 242/260 |
| **(b) + 16-byte filter** | **+47** | **+18** | **2.6 : 1** | **96.258** | **259/260** |

Declining to resume when the byte after the padding is not aligned keeps 90% of the real gains and removes 81% of the false ones. Rounding up keeps all of them and finds more, because it resumes *somewhere else* instead of not resuming — which is exactly why your measurement of it read as a wash.

## Why alignment separates, and why the round-up hid it

Taking the 145 addresses (b) adds across the corpus and classifying each against symbol truth:

| | real | false | precision |
|---|---|---|---|
| 16-byte aligned | **52** | 8 | **86.7%** |
| not aligned | **0** | 85 | 0.0% |

Every real gain is aligned and 91% of the false ones are not. That is your own hypothesis — it just cannot be applied by moving the target, only by refusing it. A lone `0xCC` inside a data structure leaves the remainder of that structure unaligned; an entry after real padding is aligned by construction.

**On the fixture you identified as the problem case**, loaded mapped so the baseline matches yours within 2%:

| | master | (b) | (b) + filter |
|---|---|---|---|
| `dotnet_readytorun_pe` | 626 | **+108** | **+0** |
| `cutwail` | 503 | +2 | +0 |

The 75-of-117 shape you found is gone entirely, not reduced.

## Full measurement

Paired before/after, symbol-table truth, no corpus losing a true positive:

| corpus | n | TP | FP | lost | bit-identical |
|---|---|---|---|---|---|
| MinGW PE | 120 | **+47** | +18 | **0** | 119/120 |
| malpedia dumps | 57 | +7 | +12 | **0** | 53/57 |
| Rust (mixed ELF/PE) | 24 | +0 | +6 | **0** | 18/24 |
| C/C++ ELF | 140 | +0 | +0 | **0** | **140/140** |
| AArch64 ELF | 72 | +0 | +0 | **0** | **72/72** |
| ARM64 Mach-O | 11 | +0 | +0 | **0** | **11/11** |

The last three are the control an Intel-only rule owes. Positive control for the zeroes: instrumented on the heaviest moving cell, 367 failed candidates consult the hook and it names a resume point for 328 — reachable and firing, not inert.

**Concentration, stated plainly.** On the built matrix the entire effect is one cell, `googletest_mingw-x86_O2-static`. Nine other `mingw-x86_O2-static` cells are bit-identical, so it is not "statically linked x86" as a population — it is the images that carry the shape, which is also what `hyperfine.exe` is. This is a rule that is inert until a binary has vtable shims behind padding, and then worth 47 functions.

**Runtime.** 2.51s on master, 2.24s with this, on that same cell — no cost, nothing like variant (a)'s order of magnitude. Resuming inside the gap also lets the scan finish regions it would otherwise re-enter.

## Scope

Two guards on top of your prototype, both for shapes it could reach:

- **A negative offset is refused.** `getRawBytes` slices the image directly, so a `gap_pointer` below the image base would read its *tail* rather than fail. `testAnAddressBelowTheImageBaseNamesNothing` uses a buffer shorter than the search window, because at a normal image size the bad slice comes back empty and the test passes without exercising the guard.
- **A missing `binary_info` is refused** rather than raising.

**AArch64 keeps the defect, deliberately.** The shared `getNextGap` is what both backends call, so its gap scan abandons gaps the same way; `int3` runs are an Intel shape and a rule for AArch64 alignment padding needs its own measurement, on a corpus where it can be checked. The base hook returns `None`, so nothing changes for it.

Worth knowing: `nextGapCandidate` already carries a narrow version of this for one shape — a retained hot-patch pad that produced no function resumes past itself rather than abandoning the gap. This generalises that reasoning rather than introducing it.

## Verification

13 new cases, each verified to fail under a mutation of the line it covers: dropping the resume branch, rounding instead of declining, the base returning a target, taking the resume without `min()`, and each of the two guards. Two of them were vacuous when first written — the overshoot case put its padding outside the search window, and the negative-offset case produced an empty slice — and both are noted in the test where the geometry matters.

Full suite **2,124 passed, 1 skipped, 2,596 subtests**; `ruff check .` and `ruff format --check .` clean; `make typecheck` exit 0 at master's own 263 diagnostics. The tree measured and the tree here produce byte-identical output on spot-checked cells, re-verified after the two guards were added.

No bundled fixture moves, so the mechanism is pinned by unit cases and the magnitude rests on corpora that are not bundled — the standard settled on #322 item 2. The `Unreleased` entry says so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,9 +80,26 @@ past roughly six lines it belongs in the PR the entry links.
 
 ### Fixed
 
+- **(intel)** Resume the gap scan inside the gap a failed candidate was found in, rather than
+  abandoning the rest of it. `getNextGap`'s resume refinement is gated on the candidate being in
+  `code_map`, so one that failed to become a function fell through to the next entry of a gap map
+  that is snapshotted once when the gap phase opens and never refreshed -- everything between the
+  failed candidate and the end of its gap was never offered as a candidate at all. The Intel
+  backend now names the first entry past the padding run that ends the failed candidate, and only
+  when that entry is already 16-byte aligned; a backend with no reading of its own padding keeps
+  the previous behaviour. *Measured on `857081f` against compiler symbol tables:* 120 MinGW PE
+  cells +47 true positives against +18 false, 24 Rust cells +0 against +6, 57 malpedia dumps +7
+  against +12, and **no true positive lost on any corpus**. The 140 C/C++ ELF cells, 72 AArch64
+  ELF cells and 11 ARM64 Mach-O cells are bit-identical. *Not reproducible from the bundled
+  fixtures, which do not move.* (#338)
+
 ### Security
 
 ### Compatibility
+
+- Recovery output moves on Intel PE and ELF images whose gap scan meets a failed candidate: the
+  bytes after it are now scanned instead of skipped. No true positive was lost on the corpora
+  this was measured over. (#338)
 
 ## Older releases
 
--- a/src/smda/common/FunctionCandidateManager.py
+++ b/src/smda/common/FunctionCandidateManager.py
@@ -288,8 +288,31 @@ class FunctionCandidateManager:
                 dont_skip,
                 next_gap,
             )
+        elif dont_skip and self.gap_pointer is not None and self.disassembly is not None:
+            # The branch above only fires for a candidate that became a function. One that did
+            # not leaves gap_pointer outside code_map, so the scan resumes at the next entry of
+            # the gap map -- snapshotted once in initGapSearch() and never refreshed -- which
+            # abandons whatever is left of the gap this candidate was found in.
+            resume = self._failedGapResumeTarget()
+            if resume is not None:
+                next_gap = min(next_gap, resume)
+                LOGGER.debug(
+                    "getNextGap(%s) => resuming inside the gap a failed candidate was in: 0x%08x",
+                    dont_skip,
+                    next_gap,
+                )
         LOGGER.debug("getNextGap(%s) final gap_ptr: 0x%08x", dont_skip, next_gap)
         return next_gap
+
+    def _failedGapResumeTarget(self):
+        """Where to resume after a gap candidate failed to become a function, or None.
+
+        None keeps the historical behaviour of abandoning the rest of the gap. A backend
+        overrides this when it can name the next plausible entry cheaply; scanning byte by byte
+        from the failed address instead costs an order of magnitude in runtime, because the
+        candidates it then walks are mostly the interior of whatever the failed one was.
+        """
+        return None
 
     def nextGapCandidate(self, start_gap_pointer=None):
         """Architecture-specific gap scan; implemented per backend."""
--- a/src/smda/intel/FunctionCandidateManager.py
+++ b/src/smda/intel/FunctionCandidateManager.py
@@ -104,6 +104,14 @@ _PDATA_MAX_FUNCTION_SIZE = 0x100000
 # EHANDLER/UHANDLER/CHAININFO are defined -- so just these eight byte values are legal.
 _UNWIND_INFO_FIRST_BYTES = frozenset(0x01 | (flags << 3) for flags in range(8))
 
+_INT3 = b"\xcc"
+#: How far past a failed gap candidate to look for the padding run that ends it. A failed
+#: candidate whose gap holds no padding within this costs nothing and keeps the old behaviour.
+_FAILED_GAP_RESUME_WINDOW = 4096
+#: Alignment a resumed entry has to already sit on. Compilers align function entries; a byte
+#: that follows padding without being aligned is padding inside something else.
+_ENTRY_ALIGNMENT = 16
+
 
 class FunctionCandidateManager(_CommonFunctionCandidateManager):
     CANDIDATE_CLASS = FunctionCandidate
@@ -192,6 +200,44 @@ class FunctionCandidateManager(_CommonFunctionCandidateManager):
         ]:
             is_alignment_sequence = False
         return is_alignment_sequence
+
+    def _failedGapResumeTarget(self):
+        """The first address past the next int3 run after the failed candidate, when aligned.
+
+        A candidate that failed leaves the rest of its gap unscanned. Resuming one byte along
+        instead walks the interior of whatever the failed candidate was, byte by byte, which
+        costs an order of magnitude in runtime; the padding run a compiler leaves between two
+        adjacent functions steps over it in one move. The answer can fall past the end of the
+        gap, which is why the caller takes the nearer of it and the next gap.
+        """
+        if self.gap_pointer is None or self.disassembly is None:
+            return None
+        binary_info = self.disassembly.binary_info
+        if binary_info is None:
+            return None
+        start = self.gap_pointer + 1 - binary_info.base_addr
+        if start < 0:
+            # a slice from a negative offset reads the tail of the image instead of failing
+            return None
+        window = self.disassembly.getRawBytes(start, _FAILED_GAP_RESUME_WINDOW) or b""
+        run_start = window.find(_INT3)
+        if run_start < 0:
+            return None
+        cursor = run_start
+        while cursor < len(window) and window[cursor] == _INT3[0]:
+            cursor += 1
+        if cursor >= len(window):
+            return None
+        target = self.gap_pointer + 1 + cursor
+        # Decline rather than round up. Rounding moves the resume point to the next aligned
+        # address, which resumes somewhere else instead of not resuming -- measured, that keeps
+        # every unaligned candidate and finds more of them. Requiring the byte after the padding
+        # to be aligned already is what separates the two populations: a lone 0xCC inside a data
+        # structure leaves the remainder of that structure unaligned, while a compiler-emitted
+        # entry after real padding is aligned by construction.
+        if target % _ENTRY_ALIGNMENT:
+            return None
+        return target
 
     def nextGapCandidate(self, start_gap_pointer=None):
         if self.language_candidates_only:
--- a/tests/testFailedGapResume.py
+++ b/tests/testFailedGapResume.py
@@ -1,0 +1,162 @@
+#!/usr/bin/python
+"""What the gap scan does after a candidate fails to become a function.
+
+`function_gaps` is a snapshot taken once, when the gap phase opens, and never refreshed. The
+resume path in `getNextGap` only looks past the candidate when that candidate *became* a
+function -- it tests membership in `code_map` -- so one that failed resumed at the next entry of
+that stale snapshot, abandoning whatever was left of the gap it was found in. Everything between
+the failed candidate and the end of its gap was never offered as a candidate at all.
+"""
+
+import logging
+import unittest
+
+from smda.common.FunctionCandidateManager import FunctionCandidateManager as CommonFunctionCandidateManager
+from smda.DisassemblyResult import DisassemblyResult
+from smda.intel.FunctionCandidateManager import _FAILED_GAP_RESUME_WINDOW, FunctionCandidateManager
+from smda.SmdaConfig import SmdaConfig
+
+logging.disable(logging.CRITICAL)
+
+BASE_ADDR = 0x400000
+GAP_START = 0x401000
+GAP_END = 0x402000
+#: the entry of the next gap in the snapshot, which is what an abandoned gap resumes at
+NEXT_GAP_START = 0x403000
+
+
+class _BufferBinaryInfo:
+    """The parts of BinaryInfo the gap resume reads, over a raw buffer."""
+
+    def __init__(self, binary):
+        self.bitness = 64
+        self.base_addr = BASE_ADDR
+        self.binary = binary
+        self.binary_size = len(binary)
+        self.code_areas = []
+
+
+def bufferWithPadding(pad_start, pad_length):
+    """A buffer whose only 0xCC run covers `pad_length` bytes from `pad_start` (an address)."""
+    buffer = bytearray(b"\x90" * (GAP_END + 0x2000 - BASE_ADDR))
+    offset = pad_start - BASE_ADDR
+    buffer[offset : offset + pad_length] = b"\xcc" * pad_length
+    return bytes(buffer)
+
+
+def managerAt(gap_pointer, buffer, manager_class=FunctionCandidateManager, recovered=None, gaps=None):
+    manager = manager_class(SmdaConfig())
+    disassembly = DisassemblyResult()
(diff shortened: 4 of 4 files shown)


danielplohmann:
This is the right fix, and the part I got wrong is the part that matters.

**Alignment as a filter rather than a round-up.** I had the discriminator and then applied it in the one way that cannot work: rounding moves the resume point to a different address, so it resumes *somewhere else* instead of declining to resume. That is why my table read as a wash and why the round-up column finds *more* unaligned candidates rather than fewer. `if target % _ENTRY_ALIGNMENT: return None` is four characters of difference in intent and the whole result. Your 52-real/8-false against 0-real/85-false split is the evidence I should have gone looking for and did not.

## What I verified here

**No bundled fixture moves — confirmed, and my own earlier claim to the contrary was wrong.** I had reported variant (b) moving 9 of 34 fixtures. That measurement loaded each fixture with `disassembleBuffer(raw, 0)`, i.e. a PE's raw file bytes as a flat buffer at address 0, which is a path no test uses and which gives the padding and alignment geometry of the file rather than of the mapped image. Re-measured through `disassembleUnmappedBuffer`, the way the suite actually loads these, master and this branch are bit-identical on all of `dotnet_readytorun_pe`, `rust_pe_gnu`, `cutwail`, `mirai_i386`, `asprox` and `cxx_pe_gnu`. Your claim holds; mine was an artefact.

**On `hyperfine.exe`, the binary #338 was written from:**

| | master | this branch |
|---|---|---|
| unmatched procedure records | 58 | **15** |
| functions | 6470 | 6717 |
| lost | — | **0** |
| runtime | 2.6s | 2.7s |

43 of the 58 recovered, nothing lost. That is 89.6% of what the unfiltered variant reached, which lands exactly on the 90% you quote.

**Gates**, stacked on `857081f`: 2123 passed, 2 skipped, 2596 subtests; `ruff check` and `ruff format --check` clean; `ty` at 270 diagnostics, master's own baseline to the number. Your run reports 2124 passed and 1 skipped against my 2123 and 2 — the same one-test environment difference I see between my master and your reported master, so the totals agree.

## One number from a different angle, not an objection

By PDB-entry truth the 247 addresses this adds on `hyperfine.exe` are 43 real and 204 not — 17% — against the 72% your MinGW cells show. I do not think that contradicts anything: the PDB names procedure and public records, and a Rust binary's outlined funclets and cold chunks are real code with no entry record, so "not in truth" is an upper bound on false positives rather than a count of them. Master already carries 4212 such addresses before this change. Worth having on the record next to the corpus figure, since it is the one cell where a reader can check anything themselves.

## On the parts I cannot check

The +47/+18 and the per-corpus controls rest on corpora that are not bundled, and the `Unreleased` entry says so, which is the convention working as intended. The two ELF controls and the AArch64 and Mach-O rows being bit-identical is what an Intel-only rule owes, and the instrumented positive control — 367 failed candidates consulting the hook, 328 answered — is the right answer to "how do I know the zeroes are not inertness". I would not have thought to ask for that.

Holding AArch64 at the base `None` is correct. The defect is in the shared `getNextGap` so it applies there too, but `int3` runs are an Intel shape and an AArch64 rule needs its own corpus to be checked against. Better an honest gap than a guess.

The two guards are both real: `getRawBytes` slicing from a negative offset reading the image tail is exactly the kind of thing that would have sat there for a year. Noting in the test that a normal image size makes that case pass vacuously, and sizing the buffer so it does not, is the detail I would have missed.

Merging.

Issue #342, opened 2026-09-11
Generate GitHub releases from the changelog section, starting with smda


danielplohmann:
Decision record from the discussion on 2026-09-11, so the reasoning survives the session that produced it. Nothing is implemented yet; smda goes first deliberately.

## The question

The repository publishes to PyPI and keeps history in `CHANGELOG.md`, so distribution and history are already covered. What is left for a GitHub release to do?

## Four things a release does that a tag cannot

1. **A semantic "latest".** Tag ordering is lexicographic, so `v4.10.0` sorts *below* `v4.5.0` — one minor version from now, anything reading the tag list for "newest smda" gets the wrong answer. `/releases/latest` is defined as the most recent non-prerelease release and cannot. This is the same class of error as the changelog ordering bug #337 fixed, which survived 186 entries because nobody can see it by eye.
2. **"Watch → Releases only".** That subscription does not exist for this repository today, because there is nothing to subscribe to.
3. **Landing-page presence.** 262 stars and 40 forks say the GitHub page is a real front door, plausibly ahead of PyPI for this kind of tool. A visitor currently sees no version information without digging into `/tags`.
4. **Zenodo DOI minting**, which hooks on releases rather than tags. Optional, noted rather than argued for.

## Decisions

**Release bodies are extracted from `CHANGELOG.md`, not generated.** #337 made the changelog machine-parseable — `## [vX.Y.Z] - YYYY-MM-DD` headings over a fixed subsection set — so the body can be lifted from the section for that tag. One source of truth and no second copy to drift.

The alternative, `gh release create --generate-notes`, is what [purepdb](https://github.com/danielplohmann/purepdb) does today. It emits a bulleted list of merged PRs and a compare link. For an entry like #327's — *72 AArch64 ELF cells 95.994 → 97.063 PPV (−683 false positives)... not reproducible from the bundled fixtures* — that collapses to "refuse a candidate the image declares interior by @r0ny123 in #327". The measurement, the corpus, the provenance caveat and the cost all vanish, and those are precisely what #323 spent ten points agreeing the entries must carry.

**smda goes first, and the rest scale out from here.** It is the most active repository of the family and therefore produces the most evidence about whether the workflow actually holds up. Porting outward before it has run for a release or two would be committing three repositories to an untested shape.

[...]

PR #343, opened 2026-09-11
ci(ci): release on a tag, with the checks that were memory before

r0ny123:
Releasing today is `make package` then `make publish` with an API token, and every check around it lives in the head of whoever is doing it. #341 and #337 wrote those checks down; this makes them run.

Pushing a `vX.Y.Z` tag builds, publishes, creates the release and closes the milestone. Nothing else is triggered by a branch push, so a release stays a deliberate act.

## What it refuses to publish

Five gates, each of which is a mistake that is only visible afterwards — and a version yanked from PyPI stays unusable under that number forever:

| gate | the mistake it catches |
|---|---|
| tag matches **both** version strings | the three-place bump with one place missed; the tag is the only thing a user sees |
| the version has a `CHANGELOG.md` section | `## [Unreleased]` renamed without being written, or not renamed at all |
| the commit is an ancestor of `master` | a release cut from a branch |
| CI concluded `success` for that commit | "tag after `master` is green" as a gate rather than a habit |
| the milestone named for the tag has no open items | #341's scope rule, enforced — a release cut with work still assigned to it |

There is already a `v4.7.0` milestone with two open items, so this gate is live rather than theoretical.

## What it does differently from the Makefile

**Trusted publishing, not a token.** `id-token: write` and no username or password, per the action's own README — so there is no secret to leak, rotate or forget to scope, and each file carries signed PEP 740 provenance.

**The build is isolated**, unlike `make package`'s `--no-isolation`, so the wheel is a product of the declared build requirements rather than of whatever the machine held.

**The artifact is tested, not the tree it came from.** Before anything is published the wheel goes into a clean environment carrying *only* the declared runtime dependencies, and disassembles a bundled sample with it. A missing module, absent package data or a dependency that only the development environment supplies fails there rather than on a user's machine. I found the shape of that check the useful way: running the full suite against the installed wheel fails, because a test imports a CI helper that needs `tqdm` — which is exactly the kind of dev-environment dependency this is meant to catch.

## Release notes

The body is this version's changelog section; `--generate-notes` appends the merged pull requests, contributors and compare link underneath, since [the API pre-pends a supplied body rather than replacing it](https://docs.github.com/en/rest/releases/releases). The entry says what changed and what it cost; the generated half says who did it and where to diff it, and neither substitutes for the other.

`.github/release.yml` puts the dependency bumps in their own group — 12 of the last 100 merged PRs are Dependabot's, and the repo already labels them, so otherwise they sit between the real changes.

## Verification

The parts that parse anything are in `.github/workflows/scripts/release_guard.py` with **12 tests**, because the changelog shape they read is a convention and a convention with no test drifts until the release that needs it finds out. Each test verified to fail under a mutation of the line it covers: the section not stopping at the next heading, the heading date not required, an empty section accepted, a version disagreement ignored, the `v` prefix not required.

Everything I could exercise without a release, I did:

- the smoke step extracted from the YAML and run verbatim — the heredoc survives the block scalar and prints `smda 4.6.0: 33 functions, status ok`
- the milestone gate run verbatim against this repo, in all three branches: no milestone, open items (fails), 100% (passes)
- `gh release create` accepts `--notes-file` and `--generate-notes` together — it reaches the API rather than rejecting the combination
- `repository-url: ''` resolves to PyPI: the canonical input carries no default, so the action falls through to the deprecated alias that defaults to the real index
- `zizmor --persona=regular` clean. It flagged `softprops/action-gh-release` as superfluous and **exited 11**, which would have failed the Workflow audit job — replaced with the runner's own `gh`, which also means one fewer third party in the path that signs and announces a release

Full suite **2,151 passed, 1 skipped, 2,596 subtests**; `ruff` clean; `make typecheck` exit 0 at **267**, master's own count, so the new script adds none.

## The one thing that cannot be automated

Trusted publishing needs the publisher configured on PyPI before the first run, and the workflow names two environments:

| environment | publisher on | workflow |
|---|---|---|
| `pypi` | pypi.org → smda → Publishing | `release.yml` |
| `testpypi` | test.pypi.org | `release.yml` |

Worth adding a required reviewer on `pypi` — that turns the publish into an approval step, which is the last chance to stop a release after everything else has passed.

**Rehearse before trusting it:** run the workflow manually from a tag. It takes the same five gates and publishes to TestPyPI, then stops without creating a release. `skip-existing` is on for that path only, so a second rehearsal of the same tag is not blocked by the first; on PyPI a collision means the version is already public and stopping is correct.

Closes nothing on its own — it is #323's rollout step 4, and it makes #341's "close it when the tag is pushed" automatic.

## Update: the ecosystem-wide shape

The second commit brings this workflow to the shape now shared by purepdb, mcrit, mcritweb, mcrit-plugin and docker-mcrit — the same guard over the same changelog convention, the same gates, the same changelog check, and a `RELEASING.md` in each repository stating only the values that differ. What changed here as a result:

- **Pre-releases.** The guard accepts a PEP 440 pre-release tag (`v4.7.0rc1`) and reports it, so the GitHub release is marked as a pre-release and never becomes latest; PyPI does not install one without `--pre`. It accepts a changelog heading with or without the `v` prefix, since the other repositories write `[1.9.0]` and this one wrote `[v4.6.0]`.
- **The wheel smoke test runs with `python -P`**, so the checkout in the working directory cannot stand in for the wheel. This repository's `src` layout hid that; running the same step in a flat-layout repository imported the source tree and passed on a wheel that was never checked.
- **The changelog check** exempts Dependabot and also covers `pyproject.toml`.
- **`SmdaConfig.VERSION` is read from `smda.__version__`** rather than kept in sync by hand, so the version is declared once, the guard reads one place, and a release bumps one line. `report.smda_version` is unchanged.
- **`pyproject.toml`** declares the license as the SPDX expression `BSD-2-Clause` with `license-files` (PEP 639) in place of the free-text form and the deprecated classifier — the wheel now carries `License-Expression: BSD-2-Clause` — and gains `Repository`, `Issues` and `Changelog` URLs. Building needs `setuptools>=77`; the `wheel` requirement was already unnecessary.
- **Python floor: stays at 3.11**, with the policy for moving it written down instead — see the update below.
- `RELEASING.md` documents cutting, rehearsing, pre-releases, recovery and the one-time configuration; `AGENTS.md` points at it and no longer describes a three-place bump.

Re-verified on the second commit: full suite **2,151 passed, 1 skipped** (`make test-all`), `ruff` clean, `make typecheck` at master's own 267; `python -m build` + `twine check --strict` pass and the wheel, installed into a clean Python 3.13 venv with `-P`, disassembles `cutwail_xored` to 33 functions with `SmdaConfig.VERSION == smda.__version__ == 4.6.0`; `zizmor --persona=regular` and `actionlint` clean.

## Update: the 3.11 floor is kept

An earlier revision raised the floor to 3.12 for ecosystem alignment. It is back at 3.11, because the premise did not hold yet: no dependency needs 3.12, MCRIT and purepdb still declare 3.11, and raising SMDA's floor alone would have left it the only component at 3.12 — with pip on 3.11 silently resolving MCRIT's unbounded `smda>=4.2.13` back to 4.7.0 while MCRIT kept advertising 3.11.

Reverted: `requires-python`, the 3.11 classifier, `ruff`'s `target-version`, the `ci.yml` matrix with its Windows and macOS legs and the coverage conditions keyed to them, the MCRIT install legs, and the interpreter the profile, security, benchmark and release workflows pin. The release workflow builds and smoke-tests on 3.11, since an installed wheel is only proven on the oldest interpreter it claims.

Kept: the PEP 639 licence expression and `license-files`, `setuptools>=77` without `wheel`, and the project URLs.

In place of the bump, a written policy in `RELEASING.md` and a `Deprecated` notice under `Unreleased`: SMDA's floor follows MCRIT's and moves in the same wave, at whichever comes first of MCRIT and purepdb moving to 3.12, a runtime dependency dropping 3.11, or 3.11's end of life in October 2027.

Verified on the current head: full suite **2,168 passed, 1 skipped, 2,601 subtests**; `ruff` clean at `target-version = "py311"`; `make typecheck` exit 0; `zizmor` clean. A real `python -m build` produces a wheel whose metadata reads `Requires-Python: >=3.11`, `License-Expression: BSD-2-Clause` and `License-File: LICENSE` with 3.11–3.14 classifiers, and `twine check --strict` passes both files. The release script and its tests were written under the 3.12 floor, so I checked they hold on 3.11: `ruff` at py311 reports syntax a version cannot parse and finds none, and all three changed Python files compile under 3.9.

## Related pull requests across the ecosystem

The same process, one PR per repository. None depends on another to merge; release order once merged is smda → purepdb → picblocks → MCRIT → MCRITweb → docker-mcrit.

- smda: danielplohmann/smda#343
- purepdb: danielplohmann/purepdb#59
- picblocks: danielplohmann/picblocks#5
- MCRIT: danielplohmann/mcrit#193
- MCRITweb: fkie-cad/mcritweb#176
- mcrit-plugin: danielplohmann/mcrit-plugin#15
- docker-mcrit: danielplohmann/docker-mcrit#12

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,6 +1,13 @@
 # Not a workflow: GitHub reads the generated-release-notes configuration from this path and
-# nowhere else. The workflow that cuts a release is `.github/workflows/publish-release.yml`.
+# nowhere else. The release itself is cut by the workflow under `.github/workflows/`.
+#
+# The generated list is appended under the hand-written CHANGELOG.md section, so it exists to
+# credit contributors and link every pull request, not to describe the release. Dependency and
+# tooling bumps are grouped so they do not sit between the real changes.
 changelog:
+  exclude:
+    authors:
+      - github-actions[bot]
   categories:
     - title: Changes
       labels:
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,17 @@
+# Not a workflow: GitHub reads the generated-release-notes configuration from this path and
+# nowhere else. The workflow that cuts a release is `.github/workflows/publish-release.yml`.
+changelog:
+  categories:
+    - title: Changes
+      labels:
+        - "*"
+      exclude:
+        labels:
+          - dependencies
+          - github_actions
+          - pre_commit
+    - title: Dependency and tooling updates
+      labels:
+        - dependencies
+        - github_actions
+        - pre_commit
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -14,7 +14,6 @@ on:
       - unlabeled
     branches:
       - master
-      - main
 
 permissions:
   contents: read
@@ -28,7 +27,7 @@ jobs:
       # The job runs unconditionally and decides inside the step, rather than skipping on an
       # `if`: a skipped job reports as neither passing nor failing, and the reason a PR is
       # exempt belongs in the log where the next reader will look for it.
-      - name: Require a changelog entry beside an engine change
+      - name: Require a changelog entry beside a user-visible change
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           REPOSITORY: ${{ github.repository }}
@@ -36,6 +35,7 @@ jobs:
           # contains() matches an array element exactly, so a label carrying a comma or a
           # name this one is a prefix of cannot be mistaken for it
           SKIP_LABELLED: ${{ contains(github.event.pull_request.labels.*.name, 'no-changelog') }}
+          IS_DEPENDABOT: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}
         run: |
           set -euo pipefail
 
@@ -43,18 +43,22 @@ jobs:
             echo "Labelled no-changelog; entry not required."
             exit 0
           fi
+          if [ "${IS_DEPENDABOT}" = "true" ]; then
+            echo "Dependency bump; the generated release notes list these under their own heading."
+            exit 0
+          fi
 
           changed="$(gh api --paginate "repos/${REPOSITORY}/pulls/${PR_NUMBER}/files" --jq '.[].filename')"
 
-          if ! printf '%s\n' "${changed}" | grep -q '^src/smda/'; then
-            echo "No files under src/smda/ changed; entry not required."
+          if ! printf '%s\n' "${changed}" | grep -qE '^(src/smda/|pyproject.toml$)'; then
+            echo "No shipped files changed; entry not required."
             exit 0
           fi
 
           if printf '%s\n' "${changed}" | grep -qx 'CHANGELOG.md'; then
-            echo "CHANGELOG.md touched beside the engine change."
+            echo "CHANGELOG.md touched beside the change."
             exit 0
           fi
 
-          echo "::error file=CHANGELOG.md::This PR changes src/smda/ but does not touch CHANGELOG.md. Add an entry under ## Unreleased describing what moved and what it cost, or apply the no-changelog label if the change genuinely needs no entry."
+          echo "::error file=CHANGELOG.md::This PR changes shipped files but does not touch CHANGELOG.md. Add an entry under ## [Unreleased], or apply the no-changelog label if the change genuinely needs no entry."
           exit 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
           cache-dependency-path: pyproject.toml
 
@@ -69,12 +69,12 @@ jobs:
         # Linux carries the full interpreter range; Windows and macOS run the floor
         # version only, to catch path/encoding portability without tripling the matrix.
         os: [ubuntu-latest]
-        python-version: ["3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.12", "3.13", "3.14"]
         include:
           - os: windows-latest
-            python-version: "3.11"
+            python-version: "3.12"
           - os: macos-latest
-            python-version: "3.11"
+            python-version: "3.12"
 
     steps:
       - name: Checkout code
@@ -98,11 +98,11 @@ jobs:
           python -m pip install -e ".[dev]"
 
       - name: Run tests
-        if: matrix.os != 'ubuntu-latest' || matrix.python-version != '3.11'
+        if: matrix.os != 'ubuntu-latest' || matrix.python-version != '3.12'
         run: python -m pytest tests/test*
 
       - name: Run tests with coverage
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
         run: |
           set +e
           python -m pytest --cov=smda --cov-branch --cov-report=xml --cov-report=term-missing tests/ 2>&1 | tee coverage-summary.txt
@@ -123,7 +123,7 @@ jobs:
       # Non-blocking for now: full coverage of changed source lines is a policy change,
       # and the metric gets observed over the next few PRs before it holds merges up.
       - name: Check diff coverage
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
         continue-on-error: true
         env:
           COMPARE_BRANCH: origin/${{ github.event.repository.default_branch }}
@@ -142,7 +142,7 @@ jobs:
           exit "$status"
 
       - name: Upload coverage report
-        if: always() && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
+        if: always() && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: coverage-report
@@ -159,16 +159,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.12", "3.13", "3.14"]
         exclude:
-          - os: windows-latest
-            python-version: "3.11"
           - os: windows-latest
             python-version: "3.12"
           - os: windows-latest
             python-version: "3.13"
-          - os: macos-latest
-            python-version: "3.11"
           - os: macos-latest
             python-version: "3.12"
           - os: macos-latest
--- a/.github/workflows/perf_benchmark.yml
+++ b/.github/workflows/perf_benchmark.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
           cache-dependency-path: pyproject.toml
 
@@ -119,7 +119,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
           cache-dependency-path: pyproject.toml
 
@@ -183,7 +183,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Download benchmark runs
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
           cache: pip
           cache-dependency-path: pyproject.toml
 
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -20,6 +20,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  PYTHON_VERSION: "3.12"
+
 jobs:
   verify:
     name: Verify the tag
@@ -31,6 +34,7 @@ jobs:
       issues: read
     outputs:
       version: ${{ steps.guard.outputs.version }}
+      prerelease: ${{ steps.guard.outputs.prerelease }}
     steps:
       - name: Refuse anything that is not a tag
         env:
@@ -51,15 +55,15 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Check the tag against the packaged version and the changelog
         id: guard
         env:
           TAG: ${{ github.ref_name }}
         run: |
-          python .github/workflows/scripts/release_guard.py --tag "${TAG}" --notes release-notes.md
-          echo "version=${TAG#v}" >> "${GITHUB_OUTPUT}"
+          python .github/workflows/scripts/release_guard.py \
+            --tag "${TAG}" --notes release-notes.md --github-output "${GITHUB_OUTPUT}"
 
       - name: Check the tag is on master
         env:
@@ -128,7 +132,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.11"
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Build the sdist and the wheel
         run: |
@@ -136,14 +140,15 @@ jobs:
           python -m build
           python -m twine check --strict dist/*
 
-      - name: Disassemble with the built wheel, runtime dependencies only
+      # -P keeps the checkout off sys.path, so the import cannot fall back to the source tree
+      - name: Smoke-test the built wheel
         env:
           VERSION: ${{ needs.verify.outputs.version }}
         run: |
           python -m venv /tmp/smoke
           /tmp/smoke/bin/python -m pip install --upgrade pip
-          /tmp/smoke/bin/python -m pip install "dist/smda-${VERSION}-py3-none-any.whl"
-          /tmp/smoke/bin/python - <<'PY'
+          /tmp/smoke/bin/python -m pip install dist/smda-"${VERSION}"-py3-none-any.whl
+          /tmp/smoke/bin/python -P - <<'PY'
           import os
           import pathlib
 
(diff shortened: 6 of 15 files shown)


r0ny123:
Renamed the workflow — two files called `release.yml` read as a duplicate in a diff even though they are unrelated.

Only one of the names was ever mine to choose. `.github/release.yml` is a fixed path: GitHub reads the generated-notes configuration from there and [nowhere else](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes). The workflow's filename is free, so that is the one that moved:

| | |
|---|---|
| `.github/release.yml` | configures the notes GitHub generates. Not a workflow. |
| `.github/workflows/publish-release.yml` | the workflow. `name: Publish release`. |

Each now names the other in its header, so the next reader does not have to work out why there were two.

No behaviour change: nothing referenced the workflow by filename except the `AGENTS.md` line added here, `zizmor` is still clean, and the suite is **2,151 passed, 1 skipped, 2,596 subtests** with `ruff` clean and `make typecheck` at master's own 267.

r0ny123:
Cut the comments back to what the code cannot say, and squashed — the intermediate states were "comments added, then removed" and a rename, neither of which is worth a reviewer's time.

`AGENTS.md` says no comments in code unless asked, and I had written 40 lines of them into the workflow. Most were either restating the line below or carrying rationale that belongs in the commit body, which is where it now is. Nine survive, and each is behaviour a reader cannot see from the code:

- **`repository-url: ''` resolves to PyPI.** The canonical input carries no default, so the action falls through to its deprecated alias, which defaults to the real index. Without this noted, it reads like a bug and gets "fixed".
- **`--generate-notes` appends under `--notes-file` rather than replacing it** — the ordering the whole notes design rests on.
- **`skip-existing` only for a rehearsal**, because a collision on PyPI means the version is already public and stopping is correct.
- **`cancel-in-progress: false`** — a cancelled publish can leave PyPI holding some files of a version and not others.
- **`id-token: write` is mandatory** for trusted publishing and must be the only credential in the job.
- **`.github/release.yml` is not a workflow** — GitHub reads that configuration from that path and nowhere else, which is also why the workflow is named apart from it.

The rest went to the commit body.

Verified it is a comment-only change rather than trusting the diff: both YAML files **parse identically** to the reviewed version, and every remaining changed line is inside a comment or a docstring. The smoke step was re-extracted from the trimmed YAML and run again — the heredoc still survives the block scalar and still prints `smda 4.6.0: 33 functions, status ok`.

Suite **2,151 passed, 1 skipped, 2,596 subtests**; `ruff` clean; `make typecheck` at master's 267; `zizmor` clean across all workflows.

r0ny123:
Merged v4.7.0 in. One thing about how, since the obvious resolution would have written a false entry into the release you just cut.

The bump moved every `Unreleased` bullet under `## [v4.7.0] - 2026-09-13` and opened a fresh empty `Unreleased`. Git showed one conflict block — this PR's `Removed` bullet — but **silently auto-merged the other three of this PR's four bullets into the dated v4.7.0 section**, by context, with no marker. Resolving only the visible block would have left the 4.7.0 entry claiming this release workflow, the 3.12 floor and the licence change had shipped in 4.7.0.

So the changelog is master's verbatim with this PR's four bullets placed explicitly in the new `Unreleased`, and checked rather than eyeballed: all four `(#343)` bullets sit under `## [Unreleased]`, and the diff against master removes nothing, so your v4.7.0 section is byte-identical to what you released.

`SmdaConfig.py` resolved to this PR's side: the bump re-set the literal to `"4.7.0"`, this PR replaces the literal with `__version__`, and `__init__.py` already reads 4.7.0 — so `SmdaConfig.VERSION` resolves to 4.7.0 with no second copy to keep in sync.

A useful side effect: run against the merged tree, `release_guard.py --tag v4.7.0` **accepts your release** — the version matches and it finds the real dated section. The first release the workflow will actually see is the one after this, but it already agrees with the one you cut by hand.

Suite **2,168 passed, 1 skipped, 2,601 subtests**; `ruff` clean; `make typecheck` exit 0; `zizmor` clean.

danielplohmann:
Reviewed against the source rather than the body, on a worktree at `0108f9a` (which already carries this morning's `6517d88`). The guard is the right shape and the tests are where they need to be — that script runs once per release, minutes before the artifact becomes permanent, and it is the only part of this that can be exercised beforehand.

## Verified here

- **The guard against the real tree, for the tag that just shipped.** `release_guard.py --tag v4.7.0` passes, writes `version=4.7.0 prerelease=false`, and extracts exactly the v4.7.0 section — 74 lines, opening at `### Changed` and stopping before `## Older releases`, with nothing from the format documentation above it or the older entries below. That is the check I most wanted to see on a file this one is, since the convention it reads was only written down two days ago.
- **The 12 tests**: `12 passed, 5 subtests` in 0.05s.
- **`SmdaConfig.VERSION is smda.__version__`** is literally true, and there is no import cycle — `smda/__init__.py` imports `logging` and nothing else.
- **The single `VERSION_SOURCES` entry is complete**, which I checked rather than assumed: `pyproject.toml` already carried `dynamic = ["version"]` with `version = {attr = "smda.__version__"}`, so after this PR there is exactly one literal and the guard reads it.
- `LICENSE` exists, so `license-files = ["LICENSE"]` resolves; the matrices are consistent at 3.12/3.13/3.14 with Windows and macOS on the floor, and the five MCRIT install legs match.
- Independently, the v4.7.0 wheel now on PyPI installed into a clean venv and run with `python -P` gives `33 functions, status ok` — the same figure your smoke step asserts, on the artifact rather than the tree.

## Five things, none of them blocking

1. **`RELEASING.md` claims a step the workflow does not run.** Its build description ends "...to disassemble a bundled sample with it and run the `smda` entry point". The smoke step disassembles the sample and stops there — and this package declares no `[project.scripts]` at all, so there is no `smda` entry point to run. It reads like a line that is true in a repository that has console scripts. Drop the clause here, or add the entry point to the smoke step in the repository that has one.

2. **The CI gate is narrower than "master is green".** It requires a successful `ci.yml` run for the commit, but master also runs Fuzzing, Security and the Performance & Correctness Benchmark, and the benchmark is the one that would catch a recovery regression. I am not asking for it in the gate — it reports `skipping` on some events, and a gate that can never complete is worse than one that is narrow — but `RELEASING.md` should say that "CI passed" means `ci.yml` specifically, so nobody reads the gate as covering the benchmark.

3. **The first rehearsal needs a tag that does not exist yet.** `workflow_dispatch` runs the workflow file from the ref it is given, so a rehearsal against `v4.7.0` or anything older finds no workflow there. In practice the first one has to be a throwaway pre-release tag cut after this merges. One sentence in the *Rehearsing* section saves somebody ten confused minutes.

4. **Two heading forms are now in play.** `AGENTS.md` and `RELEASING.md` both say to write `## [X.Y.Z] - YYYY-MM-DD`, while the file itself carries `## [v4.7.0] - 2026-09-13` and the anchor `[v4.7.0]:` at the foot. The guard accepts both deliberately, which is right for an ecosystem where the other repositories write `[1.9.0]` — but within this file it would be better to pick one and have the docs say the one that is there. Either is fine; the docs currently describe the form the file does not use.

5. **The 3.11 floor, which is the only part I would call a product decision rather than a review note.** Nothing in the engine needs 3.12, so this is ecosystem alignment, and the cost falls outside this repository: `smda.ida` is installed into whatever interpreter IDA bundles, and a user whose IDA runs 3.11 stops at 4.7.0 for good. If the supported IDA line ships 3.12 or newer this is free and worth saying so in the entry; if it does not, the `Removed` entry should say who it costs. Not mine to settle — flagging it for @danielplohmann, since it is a support-policy call rather than a packaging one.

One observation rather than a request: none of `publish-release.yml` itself runs until a tag is pushed, so the YAML around the tested script is exercised for the first time by a real release. That is what the TestPyPI rehearsal is for, and it is why I would rehearse before the next tag rather than after any change to the workflow.

r0ny123:
All five taken in `6715097`, and thanks for checking the published v4.7.0 wheel directly — an independent `33 functions, status ok` on the artifact is a better confirmation than anything the tree can give.

**1. The entry point.** Dropped. Right that it was a line true in a repository with console scripts; this one declares none.

**2. Which CI.** `RELEASING.md` now says the gate reads `ci.yml` only, and names fuzzing, the security audit and the benchmark as not covered — with your reason for leaving them out, and the benchmark called out as the one to confirm by eye before tagging, since it is the one that catches a recovery regression.

**3. The first rehearsal.** Added to *Rehearsing*: a manual run uses the workflow file as it exists at the chosen tag, so the first rehearsal needs a tag cut after this merges — a throwaway `vX.Y.Zrc0` is enough.

**4. Heading forms.** Aligned the docs to the file: `RELEASING.md` and `AGENTS.md` now say `## [vX.Y.Z]`, including the pre-release form. The sweep turned up one more of the same kind that you did not list — the guard's own error text told a user to write `## [4.8.0]`, the form this file does not use. Since the script is shared and other repositories write the bare form, it now names both rather than switching, which keeps it right everywhere it runs.

**5. The 3.11 floor.** I checked rather than guessed, against IDA itself. Both 9.3 and 9.4 ship `idapyswitch`, whose help says *"the minimum supported version is Python 3.8"*, so IDAPython is linked to whichever Python 3 the user selects rather than to a bundled one.

So the floor is not free, but the cost is narrower than stopping at 4.7.0 for good: an IDA pointed at 3.8–3.11 has to be switched to a 3.12-or-newer Python before `smda.ida` can be installed into it. That is a one-command change where a newer Python is available, and a real barrier on a locked-down image pinned to an older one. The `Removed` entry now says exactly that.

The floor itself is unchanged — whether that cost is worth ecosystem alignment is the support-policy call you flagged, and it is yours.

Suite **2,168 passed, 1 skipped, 2,601 subtests**; `ruff` clean; `make typecheck` exit 0; `zizmor` clean; and the guard still accepts `v4.7.0` against the real tree.

Agreed on the closing observation — the YAML around the script first runs on a real release, so the rehearsal belongs before the next tag.

danielplohmann:
Answering the 3.11 floor, since it is the one thing I left open — and thanks for going to `idapyswitch` rather than guessing at it, which changed the shape of the question.

**Keep 3.11 for now.** Not because the long-term argument is wrong, but because the premise the entry rests on is not true yet, and the cost of waiting is one CI leg.

## What I checked

- **No dependency needs 3.12.** `purepdb` and `pycxxfilt` are `>=3.11`; `lief` `>=3.9`, `capstone` `>=3.8`, `dnfile` `>=3.8`, `dncil` `>=3.7`.
- **3.11 is exercised, not merely claimed.** master's matrix runs 3.11 through 3.14 on Linux with Windows and macOS on the floor, and ubuntu + 3.11 is the leg that carries coverage and diff-coverage. So there is no 3.12-only syntax in the tree to be accommodated.
- **The ecosystem floor is 3.11, not 3.12.** `mcrit` is `requires-python = ">=3.11"` and classifies 3.11; `purepdb` the same. The `Removed` bullet's "the MCRIT ecosystem now shares a 3.12 floor so one interpreter serves every component" describes a state that does not exist yet — after this PR smda would be the only component at 3.12.
- **Moving alone breaks mcrit quietly.** mcrit depends on `smda>=4.2.13` with no upper bound while declaring 3.11. Ship a 3.12-only smda and pip on 3.11 silently resolves smda back to 4.7.0 — mcrit keeps advertising 3.11 while its engine is frozen, and nothing errors. That is the failure shape #342 was written to design against.

## On the long-term argument

Your case is that the switch is paid once and is better paid deliberately than under pressure. I agree with that, and it is why I want the policy written down rather than the floor left to drift. The timing does not follow from it, though:

1. **The cost of bumping does not grow while we wait.** It is one line plus matrix legs, and this repository shipped four releases in three weeks — every release between now and 3.11's EOL in October 2027 can carry it. There is no window to catch.
2. **The population that pays shrinks while we wait.** The EOL is itself what moves distributions and IDA images along. Someone pinned to 3.11 today has thirteen months of their own vendor pushing them forward; bumping now hits that person, bumping on a trigger hits fewer of them for the same benefit.
3. **For the person it actually costs, it is not "once".** Your own finding is that `idapyswitch` makes it one command *where a newer Python is available*. On a locked-down image it is a stop until somebody re-images, and "ecosystem alignment, a year early" is a hard reason to hand them. "lief dropped 3.11" is an easy one.

One mechanical note on reversibility: `UP` is in the ruff `select` list and those rules are gated on `target-version`, so `py312` brings UP040 and UP046/047 live, at which point the floor stops being metadata and becomes syntax. Nothing would be rewritten today — `src/` has no `TypeAlias`, `TypeVar` or `Generic[` — but that is how "we can always revert" stops being true without anyone deciding it.

## What I would like in this PR

Drop the floor, keep everything else. To be specific about the hunks:

**Out:** `requires-python = ">=3.12"` and the dropped 3.11 classifier; `ruff target-version = "py312"`; the `ci.yml` python-version changes (the matrix, the Windows and macOS legs, and the four `if:` conditions keyed to the coverage leg); the MCRIT install matrix legs; and the fourth `Removed` bullet with its IDA paragraph.

**In, unchanged:** PEP 639 (`license = "BSD-2-Clause"` with `license-files`, `setuptools>=77`, dropping `wheel`), the `Repository`/`Issues`/`Changelog` URLs, and the build bullet that describes them. None of it depends on the floor.

## And in place of the bump, a policy

Since what you are after is predictability, let us have it in writing rather than as a floor set early:

- smda's floor follows mcrit's, and the two move in the same release wave.
- The trigger is whichever comes first: mcrit and purepdb move to 3.12, a runtime dependency drops 3.11, or 3.11 reaches EOL in October 2027.
- Say it publicly now — a supported-Pythons line in `RELEASING.md` and a note in the next release — so `smda.ida` users get thirteen months of warning rather than discovering it as a silent pip downgrade.

Either in this PR or a follow-up, whichever keeps this one moving. Everything else here I am ready to merge.

r0ny123:
Done in `466cf71` — floor dropped, everything else kept, and the policy in writing. Your mcrit argument settles it: a 3.12-only smda under mcrit's unbounded `smda>=4.2.13` would resolve silently to 4.7.0 on 3.11, which is exactly the failure this PR exists to design out.

**Out**, per your list: `requires-python` and the 3.11 classifier, `ruff`'s `target-version`, the `ci.yml` matrix with its Windows and macOS legs and the four coverage-leg `if:` conditions, and the MCRIT install legs.

Three hunks you did not list were the same floor in other files, so they went too: the pinned interpreter in `profile.yml`, `security.yml` and `perf_benchmark.yml`, the `README.md` line, and three `AGENTS.md` lines that said 3.12+. And the release workflow's own interpreter, which matters more than it looks: the smoke step installs the wheel and runs it, so on 3.12 it would not catch a 3.11 incompatibility in the artifact. It builds and smoke-tests on 3.11 now.

**Kept**, verified in a real build rather than read off the diff: the wheel's metadata carries `Requires-Python: >=3.11`, `License-Expression: BSD-2-Clause` and `License-File: LICENSE` with 3.11–3.14 classifiers, and `twine check --strict` passes both files. `changelog.yml` stays as it was too — its Dependabot exemption and `pyproject.toml` coverage are not about the floor.

**The policy** is in `RELEASING.md` under *Supported Python versions*, with a `Deprecated` notice under `Unreleased`: the floor follows MCRIT's and moves in the same wave, at whichever comes first of MCRIT and purepdb moving to 3.12, a runtime dependency dropping 3.11, or end of life. I checked that date against the devguide rather than carrying it — PEP 664, **2027-10**, security-only now. Your point about `UP` rules turning the floor from metadata into syntax is recorded there as well, since it is the reason a later bump is not a one-line revert.

**One risk worth closing, since the release script was written while the floor was 3.12**: there is no 3.11 on this machine and I did not want to install an interpreter unasked, so I checked two ways instead. `ruff` at `target-version = "py311"` reports syntax a version cannot parse and finds none, and all three changed Python files compile under 3.9, a stricter lower bound. CI's restored `ubuntu + 3.11` leg is the authoritative run.

Suite **2,168 passed, 1 skipped, 2,601 subtests**; `make typecheck` exit 0; `zizmor` clean; the guard still accepts `v4.7.0`. The PR body now says the floor is kept rather than raised.

danielplohmann:
Verified `466cf71` on saturn. Everything checks out, including the parts I could only take on trust last time.

## The floor revert is complete, and byte-identical where it should be

`ci.yml`, `security.yml`, `perf_benchmark.yml`, `profile.yml` and `README.md` no longer appear in the diff against master at all — not reverted approximately, reverted exactly. `pyproject.toml` is now PEP 639, the four project URLs and `setuptools>=77` without `wheel`, and nothing else: `requires-python = ">=3.11"` and the 3.11 classifier are back where they were.

**The release workflow building on 3.11 is the part I would not have asked for and is the right call.** An artifact proven only on 3.12 while its metadata claims 3.11 is a gap the smoke step exists to close, and it would have survived the revert unnoticed.

## Wheel metadata, built here rather than read off the diff

`python -m build` then `twine check --strict` — both files PASSED — and the wheel carries exactly what you said:

```
Metadata-Version: 2.4
License-Expression: BSD-2-Clause
License-File: LICENSE          (shipped at smda-4.7.0.dist-info/licenses/LICENSE)
Requires-Python: >=3.11
Classifier: Programming Language :: Python :: 3.11 / 3.12 / 3.13 / 3.14
Project-URL: Homepage, Repository, Issues, Changelog
```

## The four documentation points

All four fixed in `6715097`, each in the way I would have written it:

- the `smda` entry point clause is gone from the build description;
- the CI gate now says it reads `ci.yml` only, and names the benchmark as the one to confirm by eye — which is better than what I suggested, since it says *why* the narrow gate is deliberate rather than just what it covers;
- the rehearsal section says the first one needs a tag cut after this workflow exists;
- the heading form is aligned to `## [vX.Y.Z]` throughout, and the guard's error message now names both forms instead of steering toward the one this file does not use. That is the right resolution for a script four other repositories share.

Re-ran the guard against the real tree at this head: `v4.7.0` still passes, `version=4.7.0 prerelease=false`, 12 tests in 0.04s. All 24 checks pass — 24 rather than 22, which is the restored 3.11 legs showing up.

## The policy

The *Supported Python versions* section is the part that will still be doing work in a year. Stating the resolver mechanism — that a floor raised alone becomes a silent downgrade rather than an error — is what makes the rule hold up when someone later asks why SMDA is waiting on MCRIT. Same for recording that `UP` rules make the bump non-reversible; that is the detail a future reader would otherwise have to rediscover by reverting and finding it did not work.

Merging. Thank you for the two follow-ups inside a day, and for checking 3.11 compatibility two ways rather than installing an interpreter unasked.

PR #344, opened 2026-09-13
chore: cut v4.7.0

danielplohmann:
Cuts the release the v4.7.0 milestone reads 100% for. No code changes: both version strings move to 4.7.0 and `## [Unreleased]` becomes `## [v4.7.0] - 2026-09-13`.

## What is in it

Seven merged PRs, each carrying the entry its author wrote at merge time — the first release cut under the keep-a-changelog format rather than reconstructed from merge commits afterwards.

**Changed** — the interior-refusal rules widened beyond the gap scan (#327, #329) and the test tiering entry point (#340).
**Fixed** — the gap-scan resume defect (#338/#339), the LSDA read bound (#335), and the per-backend flag for the exception-directory rule (#329).
**Compatibility** — three entries, all "recovery output moves and no true positive was lost on the corpora this was measured over": PE images declaring an exception directory, Intel images whose gap scan meets a failed candidate, and ELF images carrying an `.eh_frame`.

`Added`, `Deprecated`, `Removed` and `Security` were empty and are dropped, per the format's own rule. A fresh empty `## [Unreleased]` sits above with the full subsection scaffold.

**None of the three compatibility constants moved.** `ESCAPER_DOWNWARD_COMPATIBILITY` stays at 4.4.5 and both `PIC_HASH_ESCAPE_VERSION` constants are unchanged since v4.6.0, so this release's `Compatibility` section names recovery movement only and no constant transition. Verified against the `v4.6.0` tag rather than from the entries.

## Provenance of the figures

Every accuracy figure in this release was measured by @r0ny123 on corpora that are not bundled — built C/C++ MinGW PE, Rust, ELF, AArch64 ELF, Go, ARM64 Mach-O cells and the malpedia dumps. The entries say so individually. Reproduced here: #327 fires zero times on the five bundled ELF fixtures and #338's fixture baselines do not move, both exactly as their entries state; #329's ARM64 PE flag behaviour is visible on the bundled fixture and was reproduced.

## Verification

On this branch, saturn:

- `python -m pytest tests/` → **2,156 passed, 1 skipped, 2,596 subtests** (the one skip is `/usr/bin/true is not a fat/universal Mach-O on this host`)
- fast tier alone → 2,032 passed, 124 deselected, 1m59s
- compare links: `[Unreleased]` now spans `v4.7.0...HEAD` and `[v4.7.0]` spans `v4.6.0...v4.7.0`

**Tag after this is merged and `master` is green, not at merge** — v4.6.0's tag was created before master's CI had confirmed the merged tree, and the milestone closes when the tag is pushed.

#343 is deliberately not in this release. It is not in the milestone, so by the scope rule #341 wrote down it does not gate the cut; it lands after the tag and makes the *next* release an automated one.


r0ny123:
@danielplohmann I wanted to merge https://github.com/danielplohmann/smda/pull/343 first then cut a release 4.7.0.

PR #349, opened 2026-09-15
feat(cli): auto-detect container formats and modernize the repository furniture

r0ny123:
## What this changes

Three commits, reviewable separately:

**`chore(build)`: package metadata and repository furniture.** Fixes the `disassmbler` typo that has shipped as the PyPI summary, adds keywords and audience/typing classifiers, and adds a `MANIFEST.in` that prunes `tests/` from the sdist (it carried 1.6 MB of a suite that could not run from the archive, plus the xored malware corpora). `make init` now pins the same setuptools range as `[build-system].requires`. CI, security and fuzzing workflows no longer run twice for in-repo PR branches; the changelog and PR-title checks get a per-PR concurrency group and a timeout; the profiling comment step fails soft on a fork's read-only token. Dependabot groups updates, applies a cooldown on pip, and ignores the ruff and capstone pins whose rationale is documented in pyproject. Adds `CONTRIBUTING.md`, a PR template and YAML issue forms. README rewritten against the tree: supported targets, `SmdaConfig` defaults, report format, CLI flags and escaper markers documented from source, stale example output replaced, duplicated caveats collapsed.

**`chore(ci)`: drops comments the first commit added that restated the line they sat on**, per the AGENTS.md rule.

**`feat(cli)`: `analyze.py` auto-detects containers without `-p`.** `shouldParseHeader()` asks every loader in `FileLoader.file_loaders` whether it claims the buffer and takes the header-parsing path when one does. An explicit `-a`/`-i` still selects raw buffer mode, and `-p` overrides both. No change under `src/`; `batch_analyze.py` already always parsed headers.

Deliberately not added: a CodeQL workflow (default setup is a repo setting and an in-repo workflow disables it), Scorecard, CODEOWNERS, a code of conduct. Settings only the maintainer can change are listed at the end.

## How it was verified

```
yaml.safe_load over all .github YAML   ok
actionlint                             clean
zizmor --offline --persona=regular     no findings
python -m build && twine check         both PASSED; wheel has py.typed, no tests/
pytest tests/testPackageMetadata.py tests/testReleaseGuard.py   23 passed
pytest tests/testAnalyzeCliRouting.py  4 passed
make test                              2051 passed, 2 skipped, 124 deselected
ruff check . / ruff format --check .   clean
```

CLI routing checked by hand: `python analyze.py /bin/cat` now matches `-p` at 140 functions; `-a 0x400000` still forces raw mode; the blockblast DEX fixture yields the same 2219 functions / 2527 blocks / 9824 instructions through either path.

- [x] `make lint` and `make format`
- [x] `make test-all` (fast tier plus the touched suites; full tier not re-run for the doc-only commits)
- [x] `CHANGELOG.md` entry under `## [Unreleased]`
- [x] Golden fixtures unchanged

## Settings to set in the GitHub UI

1. Code scanning: enable CodeQL default setup (Python).
2. Enable private vulnerability reporting; SECURITY.md already links the advisory form.
3. Enable Dependabot alerts and security updates, secret scanning, push protection.
4. Ruleset on `master` requiring the CI checks; ruleset on `v*` blocking tag moves and deletion.
5. Confirm the `pypi` environment is restricted to `v*` tags with a required reviewer.
6. Labels `bug`, `enhancement`, `no-changelog`, `profile` must exist for the templates and workflows.


danielplohmann:
Milestoned for `v4.9.0`. The routing change reads well, and asking `FileLoader.file_loaders` rather than keeping a second copy of the magic bytes is the right call — detection cannot drift from what the loader path actually handles. Two notes, neither blocking.

**`readFileContent()` is now unconditional, so the header path reads the file twice.** `shouldParseHeader()` needs the bytes, but `disassembleFile()` then constructs `FileLoader(file_path, map_file=True)`, which reads the same file from disk again. Harmless at CLI scale and nothing under `src/` moves, but it is free to avoid: `isCompatible()` only needs a prefix, or the buffer could be threaded into the loader path.

**CI no longer runs on tag pushes.** Restricting `push:` to `branches: [master]` in `ci.yml`, `fuzzing.yml` and `security.yml` means a `v*` tag push stops triggering them. I traced this against the release pipeline, because `publish-release.yml` gates on *Check CI passed on this commit* — that step resolves by `head_sha`, and the tag already has to be an ancestor of `master`, so the master-push run satisfies the gate. **This is safe**, and it removes a duplicate run rather than a real one. Recording it because it is not visible from the diff and would otherwise only surface at the next release.

On the settings list at the end of the description: `bug`, `enhancement` and `no-changelog` all exist already, so item 6 reduces to creating `profile` — and that gap predates this PR, since `profile.yml` on master already gates on that label. The rest is additive hardening that nothing merged here depends on, so none of it holds up the merge.

PR #350, opened 2026-09-15
fix(report): bound stored buffer inflation by MAX_IMAGE_SIZE

r0ny123:
## What this changes

`SmdaReport.fromDict()` restores a report's stored `buffer` field with `ZipFile.read()`, which has no output limit. A few kilobytes of base85 in a crafted serialized report can inflate to gigabytes on load.

`_unpackBuffer()` now:

- accepts only the layout `_packBuffer()` writes: a single deflated entry named `buffer`;
- rejects an entry whose declared uncompressed size exceeds `SmdaConfig.MAX_IMAGE_SIZE`;
- inflates the entry's compressed bytes with an output cap of `MAX_IMAGE_SIZE + 1`, so a stream that is larger than its metadata claims is cut off instead of trusted;
- requires the inflated data to end the deflate stream exactly and to match the declared size and CRC.

Every rejection is a `ValueError` (or the `BadZipFile`/`zlib.error` the parser already raised), which `fromDict()` already treats as operational: the buffer stays `None`, a warning is logged, and the rest of the report loads. Reports written by `_packBuffer()` round-trip unchanged. No other decompression site exists under `src/smda/`.

## How it was verified

```
pytest tests/testSmdaReportBuffer.py   17 passed; 6 of the new tests fail on master
make test                              2178 passed, 1 skipped (full tree, slow tier included)
ruff check . / ruff format --check .   clean
make typecheck                         exit 0, no diagnostics in the changed code
fuzzing/replay.py report <corpus floor>  all OK
```

A local mutation loop (120,000 byte flips, truncations and insertions over packed buffers of 0, 40 and 1,024 bytes) raised only `ValueError`, `BadZipFile`, `zlib.error`, `NotImplementedError` and `UnicodeDecodeError`, all of which `fromDict()` absorbs; none escaped as a non-operational exception.

- [x] `CHANGELOG.md` entry under `## [Unreleased]`
- [x] Golden fixtures unchanged


danielplohmann:
Milestoned for `v4.9.0`. The bound itself looks right, and routing every rejection through `ValueError`/`BadZipFile`/`zlib.error` so `fromDict()` keeps absorbing them is the correct call — a malformed report still loads with the buffer dropped.

One question on the implementation.

`_unpackBuffer()` reimplements the local-header walk (`PK\x03\x04`, the `<HH` at offset 26, the compressed-data bounds) to reach a bounded inflate. `ZipFile.open()` already does that walk, and `ZipExtFile.read(n)` is incrementally bounded, so the same DoS bound is available in about five lines:

```python
with zip_file.open(entry) as fh:
    buffer = fh.read(max_size + 1)
    if len(buffer) > max_size:
        raise ValueError(f"stored buffer exceeds MAX_IMAGE_SIZE ({max_size} bytes)")
    if fh.read(1):
        raise ValueError("stored buffer does not match its archive metadata")
```

I checked that with the limit dialled down to 1 KiB:

| input | result |
| --- | --- |
| honest round-trip | returns the original bytes |
| 50 MB of `A` packed normally | rejected on declared size, 0.003 s |
| central-directory `file_size` forged down to 10 | `BadZipFile: Bad CRC-32`, 0.003 s, never inflates the rest |

Where the stdlib path is genuinely weaker: forging `file_size` **and** `CRC`, in both the central directory and the local header, to match a 10-byte prefix gets that truncated buffer returned as if it were the whole thing. Your `decompressor.eof` / `unconsumed_tail` / `unused_data` checks reject it, because they require the deflate stream to end exactly where the metadata says it does.

So the hand-rolled walk does buy something. My question is whether it is meant to. That case is a silently truncated buffer rather than unbounded decompression, and the docstring and the changelog entry both describe the bound, not the strictness. If the strictness is intentional, a line saying so would stop the next reader simplifying it away — it is the only thing justifying a second copy of zipfile's on-disk layout in this file. If it is not, the stdlib path is ~25 fewer lines.

Happy either way; I would just like the reason recorded next to the code.

